PR #957, opened 2022-06-15
feat: check doc links in CI

dvilelaf:
## Proposed changes

Adds a test to check the documentation links. Should not be merged until the docs are public.

## Fixes

n/a

## Types of changes

## Checklist


8ball030 (dismissed):

Adamantios (dismissed):

docs/api/skills/abstract_round_abci/utils.md
```diff
@@ -16,7 +16,7 @@ Tool to verify Randomness retrieved from various external APIs.
 
 The ciphersuite used is BLS_SIG_BLS12381G2_XMD:SHA-256_SSWU_RO_NUL_
 
-https://drand.love/docs/specification/`cryptographic`-specification
```
dvilelaf, line 19: These two drand links contain a #, and pydocs-markdown removes it and reformats the link, effectively breaking it. I have not found a way to escape this behaviour or replace the hashtag with an equivalent code. We could use a url shortener but i felt like this way is more informative.

DavidMinarsch (dismissed):

scripts/check_doc_links.py
```diff
@@ -0,0 +1,96 @@
+# -*- coding: utf-8 -*-
```
angrybayblade, line 1: Missing shebang?

dvilelaf: It was the python commands in the tox file, although i dont know why the other do not require it.

DavidMinarsch (approved):

PR #964, opened 2022-06-15
Added node modules

8ball030:
## Proposed changes

Hi david, i think this fixes your problem,

### DIFF
--- a/tox.ini
+++ b/tox.ini
@@ -165,7 +165,8 @@ commands = {toxinidir}/scripts/generate_api_documentation.py --check-clean
 skipsdist = True
 skip_install = True
 deps =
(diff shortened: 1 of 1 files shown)


tox.ini
```diff
@@ -165,7 +165,8 @@ commands = {toxinidir}/scripts/generate_api_documentation.py --check-clean
 skipsdist = True
 skip_install = True
 deps =
-commands = {toxinidir}/scripts/check_doc_links.py
+    requests==2.27.1
+commands = python {toxinidir}/scripts/check_doc_links.py
```
8ball030, line 169: You need to install requests and then call python on the script;
image

DavidMinarsch (approved):

Issue #672, opened 2022-04-05
We should never raise `ABCIAppInternalError` outside `abstract_round_abci`


Adamantios:
**Is your feature request related to a problem? Please describe.**
We should never raise `ABCIAppInternalError` outside `abstract_round_abci` because it kills the app.

**Describe the solution you'd like**
We should transition to a proper round instead of raising.

**Describe alternatives you've considered**
Transition to `FailedRound`. Example solution in a7b114f48ed6c6df2f8d86025b56f214e6a20a9a.

**Additional context**
`ABCIAppInternalError` is raised [here](https://github.com/valory-xyz/consensus-algorithms/blob/ba4848062851cea87062ec00d06294044e2587ae/packages/valory/skills/transaction_settlement_abci/rounds.py#L191-L193) and also [here](https://github.com/valory-xyz/consensus-algorithms/blob/ba4848062851cea87062ec00d06294044e2587ae/packages/valory/skills/transaction_settlement_abci/rounds.py#L136-L139).

Adamantios:
We do not raise internal errors anymore, however, there are still two exceptions being raised in behaviours or rounds:

https://github.com/valory-xyz/open-autonomy/blob/a41a826e826ba50a44471322475d216b70efa00c/packages/valory/skills/apy_estimation_abci/behaviours.py#L1305-L1308

https://github.com/valory-xyz/open-autonomy/blob/a41a826e826ba50a44471322475d216b70efa00c/packages/valory/skills/transaction_settlement_abci/rounds.py#L162

These are there to avoid developer mistakes, so they could probably be removed.

Adamantios:
Closing the issue because they are both design errors, not runtime errors.

PR #983, opened 2022-06-20
[WIP] removing ABCIAppInternalError from TransactionSettlement skill

Karrenbelt:
## Proposed changes

Addresses the following issue: https://github.com/valory-xyz/open-autonomy/issues/672 

EDIT: now also https://github.com/valory-xyz/open-autonomy/issues/993 - during payload redesign this can be revisited

## Fixes

## Types of changes

## Checklist

## Further comments


packages/valory/skills/transaction_settlement_abci/rounds.py
```diff
@@ -197,20 +191,9 @@ def should_check_late_messages(self) -> bool:
...
-        )
-        late_arriving_tx_hashes_parsed = []
-        for unparsed_hash in late_arriving_tx_hashes_unparsed:
-            if len(unparsed_hash) % TX_HASH_LENGTH != 0:
-                # if we cannot parse the hashes, then the developer has serialized them incorrectly.
-                raise ABCIAppInternalError(
-                    f"Cannot parse late arriving hashes: {unparsed_hash}!"
-                )
-            parsed_hashes = textwrap.wrap(unparsed_hash, TX_HASH_LENGTH)
-            late_arriving_tx_hashes_parsed.extend(parsed_hashes)
```
Karrenbelt, line 211: simply not checking length here anymore. If it is wrong the developer made a mistake, one that should be caught elsewhere and ideally also earlier. This business logic would then have to become part of the abstract_round_abci. It might be worth considering using a specific type, something wrapping serialized data alongside it's expected length of individual "blocks".

packages/valory/skills/transaction_settlement_abci/rounds.py
```diff
@@ -379,17 +362,24 @@ class ValidateTransactionRound(VotingRound):
...
         # if reached participant threshold, set the result
+
         if self.positive_vote_threshold_reached:
+            # We obtain the latest tx hash from the `tx_hashes_history`.
+            # We keep the hashes sorted by their finalization time.
+            # If this property is accessed before the finalization succeeds,
+            # then it is incorrectly used.
+            synchronized_data = cast(SynchronizedData, self.synchronized_data)
+            if not synchronized_data.to_be_validated_tx_hash:
+                return synchronized_data, Event.INCORRECT_SERIALIZATION
```
Karrenbelt, line 373: we now return a incorrect serialization event instead of raising an Error

packages/valory/skills/transaction_settlement_abci/behaviours.py
```diff
@@ -370,20 +370,23 @@ def async_act(self) -> Generator:
 
     def has_transaction_been_sent(self) -> Generator[None, None, Optional[bool]]:
         """Transaction verification."""
+
+        if not self.synchronized_data.tx_hashes_history:
+            return False
+
+        to_be_validated_tx_hash = self.synchronized_data.tx_hashes_history[-1]
```
Adamantios, line 377: Why not use the `to_be_validated_tx_hash` property instead and check for `None` to return early?

packages/valory/skills/transaction_settlement_abci/rounds.py
Adamantios: Please open an issue for what you described.

packages/valory/skills/transaction_settlement_abci/rounds.py
Karrenbelt: If we don't have a tx hash than something went wrong during serialization. The best alternative seems `FINALIZATION_FAILED` which I think is less appropriate.

packages/valory/skills/transaction_settlement_abci/behaviours.py
```diff
@@ -370,27 +370,31 @@ def async_act(self) -> Generator:
 
     def has_transaction_been_sent(self) -> Generator[None, None, Optional[bool]]:
         """Transaction verification."""
+
+        to_be_validated_tx_hash = self.synchronized_data.to_be_validated_tx_hash
+        if not to_be_validated_tx_hash:
+            return False
```
dvilelaf, line 376: So we enter validation without a to_be_validated_tx_hash? Doesnt that mean that we have a problem in our logic then if we reach this situation?

Adamantios: Exactly, this is why we were raising before, but we need a better solution because we should only raise in the abstract abci level.

packages/valory/skills/transaction_settlement_abci/rounds.py
```diff
@@ -379,17 +362,24 @@ class ValidateTransactionRound(VotingRound):
     def end_block(self) -> Optional[Tuple[BaseSynchronizedData, Enum]]:
         """Process the end of the block."""
         # if reached participant threshold, set the result
+
         if self.positive_vote_threshold_reached:
+            # We obtain the latest tx hash from the `tx_hashes_history`.
+            # We keep the hashes sorted by their finalization time.
+            # If this property is accessed before the finalization succeeds,
+            # then it is incorrectly used.
```
DavidMinarsch, line 370: Indeed, so the event INCORRECT_SERIALIZATION is very misleading. The problem is not necessarily a serialisation issue. Taking into account discussion here https://github.com/valory-xyz/open-autonomy/pull/983/files#r901875239 I suggest calling it `NOTHING_TO_VALIDATE`.

DavidMinarsch (changes requested):
Please also resolve the new merge conflicts which have arisen

packages/valory/skills/transaction_settlement_abci/rounds.py
Karrenbelt: I implemented this in a very simple manner for now. When the payload redesign occurs this may be revisited.

packages/valory/skills/abstract_round_abci/base.py
```diff
@@ -1091,16 +1103,22 @@ def process_payload(self, payload: BaseTxPayload) -> None:
...
             raise ABCIAppInternalError(
                 f"sender {sender} has already sent value for round: {self.round_id}"
             )
 
+        if self._hash_length:
+            content = payload.data.get(self.payload_attribute)
+            if not content or len(content) % self._hash_length:
+                msg = "FSM design error: expecting serialized data of chunk size"
+                raise ABCIAppInternalError(f"{msg} {self._hash_length}, got: {content}")
+
```
Karrenbelt, line 1121: this compensates the loss of assertion on the serialized data format in by raising a `ABCIAppInternalError` on processing the payload.

On a separate note: I suggest to move these checks in process_payload method to the check_payload method and call latter from within the former method. I noticed they raise slightly different errors, `ABCIAppInternalError` vs `TransactionNotValidError`; is there a reason?

packages/valory/skills/transaction_settlement_abci/behaviours.py
Karrenbelt: now that it is strict, there is no longer the need to transition with an event `NOTHING_TO_VALIDATE` as an error will be raised. Hence I've removed this transition again in a7a767b

packages/valory/skills/abstract_round_abci/base.py
```diff
@@ -1091,16 +1103,22 @@ def process_payload(self, payload: BaseTxPayload) -> None:
...
-                f"{sender} not in list of participants: {sorted(self.synchronized_data.participants)}"
+                f"{sender} not in list of participants: {sorted(self.accepting_payloads_from)}"
             )
 
         if sender in self.collection:
             raise ABCIAppInternalError(
                 f"sender {sender} has already sent value for round: {self.round_id}"
             )
 
+        if self._hash_length:
```
Adamantios, line 1116: Why name this attribute `_hash_length`? I mean it could be made up of any type of fixed-size items, not just hashes as it happened in the current case with Tx settlement skill.

packages/valory/skills/transaction_settlement_abci/rounds.py
```diff
@@ -472,6 +464,7 @@ class SynchronizeLateMessagesRound(CollectNonEmptyUntilThresholdRound):
     none_event = Event.NONE
     selection_key = "participant"
     collection_key = "late_arriving_tx_hashes"
+    _hash_length = TX_HASH_LENGTH
```
Adamantios, line 467: `CollectNonEmptyUntilThresholdRound` returns a `list` so this is not going to work as expected here. In essence, we retrieve a list from each agent that has late-arriving hashes. This list contains strs with concatenated hashes.

packages/valory/skills/transaction_settlement_abci/rounds.py
```diff
@@ -155,16 +154,13 @@ def to_be_validated_tx_hash(self) -> str:
...
         """
-        if len(self.tx_hashes_history) > 0:
-            return self.tx_hashes_history[-1]
-        raise ABCIAppInternalError(  # pragma: no cover
-            "An Error occurred while trying to get the tx hash for validation: "
-            "There are no transaction hashes recorded!"
-        )
+        if not self.tx_hashes_history:
+            raise ValueError("FSM design error: tx hash should exist")
+        return self.tx_hashes_history[-1]
```
Adamantios, line 163: Why do we still raise here? Is it possible to avoid this by updating `tx_hashes_history` to use `get_strict`?

Adamantios (changes requested):
I like the simplicity of the solution, but I think that it needs some changes because of the `CollectNonEmptyUntilThresholdRound` returning a `list`.

packages/valory/skills/transaction_settlement_abci/rounds.py
Karrenbelt: we could do that. But what would be the benefit? We would also end up raising a `ValueError`

packages/valory/skills/abstract_round_abci/base.py
Karrenbelt: It is only used by the `SynchronizeLateMessagesRound`. I suggest to add support for checks on other payloads during payload redesign, for which there exists a separate ticket.

packages/valory/skills/transaction_settlement_abci/rounds.py
Karrenbelt: The incoming payload might be a concatenated set of tx_hashes, but it's not a list.
https://github.com/valory-xyz/open-autonomy/blob/c2bf8beda36c4ed5744a5cb2944aa357170919c3/packages/valory/skills/transaction_settlement_abci/payloads.py#L172-L175

Adamantios: Please take a look at the `CollectNonEmptyUntilThresholdRound`, I believe it would help you understand my point. Also, notice that the `collection_key` has nothing to do with the `tx_hashes` property that you referenced.

packages/valory/skills/abstract_round_abci/base.py
Adamantios: So just to clarify, would this be removed during the redesign?

Adamantios (approved):
LGTM. I'd suggest ensuring that the e2e tests are not broken before merging, because I noticed that they timed out.
Also, let's address https://github.com/valory-xyz/open-autonomy/pull/983#discussion_r905565813 first.

packages/valory/skills/abstract_round_abci/base.py
```diff
@@ -1091,38 +1103,54 @@ def process_payload(self, payload: BaseTxPayload) -> None:
...
+            content = payload.data.get(self.payload_attribute)
+            if not content or len(content) % self._hash_length:
+                msg = f"Expecting serialized data of chunk size {self._hash_length}"
+                raise TransactionNotValidError(
+                    f"{msg}, got: {content} in {self.round_id}"
+                )
+
```
Karrenbelt, line 1153: here we raise `TransactionNotValidError`, which is intercepted in `ABCIRoundHandler.deliver_tx` here: 

https://github.com/valory-xyz/open-autonomy/blob/2822968f858def392ebfd4d48906ce1a81548234/packages/valory/skills/abstract_round_abci/handlers.py#L189-L208

packages/valory/skills/abstract_round_abci/base.py
Karrenbelt: I would say most likely so. This information should be present on the payload class in my oppinion, not on the round.

Adamantios (dismissed):

packages/valory/skills/abstract_round_abci/base.py
Karrenbelt: it is now, logging error on processing, and raising the `TransactionNotValidError` captured for logging in the `ABCIRoundHandler.deliver_tx` method

Karrenbelt:
```
2022-06-24T15:42:45.3410749Z =========================== short test summary info ============================
2022-06-24T15:42:45.3410964Z FAILED tests/test_agents/test_register_reset_abci.py::TestTendermintReset::test_run[4]
2022-06-24T15:42:45.3411220Z FAILED tests/test_agents/test_registration.py::TestRegistrationStartUpFourAgentsCatchUp::test_run[4]
```

so last one is known to be flaky, other as well I think. PR should not accepted on that premise though. WIP

packages/valory/skills/abstract_round_abci/base.py
```diff
@@ -1093,38 +1105,54 @@ def process_payload(self, payload: BaseTxPayload) -> None:
...
             raise ABCIAppInternalError(
                 f"sender {sender} has already sent value for round: {self.round_id}"
             )
 
+        if self._hash_length:
+            content = payload.data.get(self.payload_attribute)
+            if not content or len(content) % self._hash_length:
+                msg = f"Expecting serialized data of chunk size {self._hash_length}"
+                _logger.error(f"{msg}, got: {content} in {self.round_id}")
+                return
```
DavidMinarsch, line 1123: you should `raise ABCIAppInternalError` here.

DavidMinarsch (changes requested):

packages/valory/skills/abstract_round_abci/base.py
DavidMinarsch: as mentioned here: https://github.com/valory-xyz/open-autonomy/pull/983#discussion_r905565821

DavidMinarsch: And consistent with the other logic in this method

Karrenbelt: I thought the idea was we should NOT do that because the content might come from a malicious agent, which could thus terminate the application https://github.com/valory-xyz/open-autonomy/pull/983#discussion_r905565813

DavidMinarsch: That's not what the comment says. The comment just explains the type or error.

DavidMinarsch (approved):

PR #985, opened 2022-06-20
Refactor reset and pause

Karrenbelt:
## Proposed changes

There were also an `ABCIAppInternalError`s being raised here. I noticed that these two methods are not needed, since they are present on the base class [`CollectionRound`](https://github.com/valory-xyz/open-autonomy/blob/ff987b6b673fa70d3dd708d629ef5f64891df6c3/packages/valory/skills/abstract_round_abci/base.py#L1086-L1124), where they are more comprehensive (here we don't check `round_count`(

## Fixes

A flag was added on `CollectionRound` to be able to specify whether a concrete round inheriting from it should consider payloads only from the active set of agents, or from all registered agents, thus also those currently not participating - which is used to allow agents to rejoin.

## Types of changes

## Checklist

## Further comments


Adamantios (approved):

packages/valory/skills/reset_pause_abci/rounds.py
```diff
@@ -53,38 +50,6 @@ class ResetAndPauseRound(CollectSameUntilThresholdRound):
     allowed_tx_type = ResetPausePayload.transaction_type
     payload_attribute = "period_count"
 
-    def process_payload(self, payload: BaseTxPayload) -> None:  # pragma: nocover
```
DavidMinarsch, line 56: It is not identical to the base class!

The base class uses `self.synchronized_data.participants`. Here we use `self.synchronized_data.all_participants`. The difference is because we want ResetAndPauseRound to allow agents which are not currently in the set `participants` to rejoin this set.

So this "refactoring" changes the implementation.

DavidMinarsch: Solution suggestion: on the `CollectionRound` class add a flag which by default is `False` which determines whether we use `self.synchronized_data.participants` or `self.synchronized_data.all_participants`.

packages/valory/skills/abstract_round_abci/base.py
```diff
@@ -1068,11 +1068,21 @@ class CollectionRound(AbstractRound):
...
+    _allow_rejoin_payloads: bool = False
+
     def __init__(self, *args: Any, **kwargs: Any):
         """Initialize the collection round."""
         super().__init__(*args, **kwargs)
         self.collection: Dict[str, BaseTxPayload] = {}
 
+    @property
+    def accepting_payloads_from(self) -> FrozenSet[str]:
+        """Accepting from the active set, or also from (re)joiners"""
```
Adamantios, line 1081: Very insignificant suggestion. How about:

`accepting_payloads_from` -> `whitelisted_participants`? :sweat_smile:

DavidMinarsch (approved):

PR #988, opened 2022-06-21
[WIP] Fix/repricing

### LINKED ISSUE #986: Transaction Settlement ABCI assumes pricing mechanism
### Subject of the issue
When using the gas_station pricing strategy, the transaction settlement skill fails.

### Your environment
- OS: ubuntu
- Python version: [e.g. 3.10,5

- docker-compose deployment however the issue is localised to the transaction settlement abci.

### Steps to reproduce

when running an oracle-agent,
```
public_id: valory/ledger:0.1.0
type: connection
config:
  ledger_apis:
    ethereum:
      address: ${CONNECTION_LEDGER_CONFIG_LEDGER_APIS_ETHEREUM_ADDRESS:str:http://localhost:8545}
      chain_id: ${CONNECTION_LEDGER_CONFIG_LEDGER_APIS_ETHEREUM_CHAIN_ID:int:31337}
      default_gas_price_strategy: gas_station
      gas_price_strategies:
        gas_station:
          gas_price_api_key: null
          gas_price_strategy: null
```
if you use a gas_pricing strategy of gas_station, you fail with the following error;

image

Note. this is caused by the assumption within the transaction_settlement_abci that there  post London fork pricing paramters within the transaction.

image

### Expected behaviour
The transaction settlement abci should choose the appropriate gas strategy from the ledger connection.

### Actual behaviour
We fail with a key error

dvilelaf:
## Proposed changes

- Guess gas strategy from transaction to avoid assuming eip1559
- Trigger reprice when UNDERPRICED status detected

## Fixes

#986 
 [underpriced ticket](https://trello.com/c/e50xtivz/778-missing-logic-for-underpriced-transactions)

## Types of changes

## Checklist


packages/valory/skills/transaction_settlement_abci/behaviours.py
```diff
@@ -165,27 +203,18 @@ def _get_tx_data(
                 "send_raw_transaction unsuccessful! Insufficient funds."
             )
 
+        # Trigger repricing mechanism if we have succesfully detected the strategy price
+        # and we receive an underpriced status.
+        if rpc_status == RPCResponseStatus.UNDERPRICED and gas_price:
+            self.params.gas_price = gas_price
+
```
dvilelaf, line 210: I had to move this from the bottom to here, or then we would not trigger the repricing mechanism as the next IF returns even for UNDERPRICED status.

packages/valory/skills/transaction_settlement_abci/behaviours.py
```diff
@@ -165,27 +203,18 @@ def _get_tx_data(
                 "send_raw_transaction unsuccessful! Insufficient funds."
             )
 
+        # Trigger repricing mechanism if we have succesfully detected the strategy price
+        # and we receive an underpriced status.
```
Adamantios, line 207:
```suggestion
        # Store gas price if we have successfully detected the strategy and we receive an underpriced status, 
        # so that the repricing mechanism is triggered next time.
```

packages/valory/skills/transaction_settlement_abci/behaviours.py
```diff
@@ -139,16 +140,53 @@ def _get_tx_data(
...
+        gas_strategy = None
+        if (
+            "maxFeePerGas" in message.raw_transaction.body
+            and "maxPriorityFeePerGas" in message.raw_transaction.body
+        ):
+            gas_strategy = "eip"
+        if "gasPrice" in message.raw_transaction.body:
+            gas_strategy = "gas_station"
+
+        self.context.logger.info(f"Detected gas strategy: {gas_strategy}")
```
Adamantios, line 165: Let's also open an issue for this part, so that we remove it when we can access the connection's configuration.

packages/valory/skills/transaction_settlement_abci/behaviours.py
Adamantios: We need to think if this is safe. Do we always have the information that you moved if `rpc_status != RPCResponseStatus.SUCCESS`?

packages/valory/skills/transaction_settlement_abci/behaviours.py
```diff
@@ -139,16 +140,53 @@ def _get_tx_data(
...
+
+        # Get the gas params
+        gas_price_params = []
+        if gas_strategy == "eip":
+            gas_price_params = ["maxPriorityFeePerGas", "maxFeePerGas"]
+        if gas_strategy == "gas_station":
+            gas_price_params = ["gasPrice"]
+
+        tx_data["tx_digest"] = cast(str, tx_digest)
```
dvilelaf, line 174: I need to move this line down again, this was a mistake.

packages/valory/skills/transaction_settlement_abci/behaviours.py
Adamantios: I meant nonce and digest. The latter has now been moved again.

packages/valory/skills/transaction_settlement_abci/behaviours.py
dvilelaf: #195

dvilelaf: Please note: even if we address that issue, we could only access to the default price strategy config, but an agent could be not using the default strategy (try_get_gas_pricing can be called with a different gas_price_strategy) so I think we still need to keep using this method.

Adamantios: Yes, but neither our contract nor tx settlement skill will allow that behaviour. Also, if a user implements their own skill and does that, then it is a mistake. The strategy should always be set depending on the global configuration IMO.

packages/valory/skills/transaction_settlement_abci/behaviours.py
```diff
@@ -165,27 +200,21 @@ def _get_tx_data(
                 "send_raw_transaction unsuccessful! Insufficient funds."
             )
 
+        # Store gas price if we have successfully detected the strategy and we receive an underpriced status,
+        # so that the repricing mechanism is triggered next time.
+        if rpc_status == RPCResponseStatus.UNDERPRICED and gas_price:
+            self.params.gas_price = gas_price
```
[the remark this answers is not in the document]
DavidMinarsch: Previously we stored it all the time, why the change now?

DavidMinarsch (changes requested):

packages/valory/skills/transaction_settlement_abci/behaviours.py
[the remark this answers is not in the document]
dvilelaf: I think that is not 100% correct. The [current code](https://github.com/valory-xyz/open-autonomy/blob/main/packages/valory/skills/transaction_settlement_abci/behaviours.py#L200) that sets gas_price ATM won't be executed due to what I think it is a [bug](https://github.com/valory-xyz/open-autonomy/blob/main/packages/valory/skills/transaction_settlement_abci/behaviours.py#L168-L172): we cannot make because we immediately return if RPCResponseStatus is not SUCCESS

```
        if rpc_status != RPCResponseStatus.SUCCESS:
            self.context.logger.warning(
                f"send_raw_transaction unsuccessful! Received: {rpc_status}"
            )
            return tx_data
```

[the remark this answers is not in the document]
Adamantios: I am not sure about this PR. The decision to return when we implemented the tx settlement skill was because a "replacement transaction underpriced" error means that we have not repriced correctly, which cannot happen since the ledger will take care of the correct increase. If we received a non-successful response, then we thought it would be better to use another keeper (therefore return immediately).

To clarify, we only replace txs that have been sent. If a tx does not have a nonce etc., then we cannot replace a non-existing tx. The error that has been mentioned, "Transaction underpriced" is different and needs further investigation on what this error means, as I believe it is not the same.

PR #1009, opened 2022-06-24
Enable e2e tests to check for successful events other than `DONE`

Adamantios:
## Proposed changes

We now define a happy path for our e2e tests, using a `RoundChecks` dataclass. This enables us to specify the name of the successful event which previously could only be `DONE` for the test to succeed. This will also allow us to re-enable the APY e2e tests.

## Fixes

n/a

## Types of changes

## Checklist

## Further comments

n/a

### DIFF
--- a/tests/test_agents/base.py
+++ b/tests/test_agents/base.py
@@ -130,10 +130,6 @@ class BaseTestEnd2End(AEATestCaseMany, UseFlaskTendermintNode):
             ANY_ADDRESS,
         )
         self.set_config(
-            "vendor.valory.connections.abci.config.host",
-            ANY_ADDRESS,
-        )
-        self.set_config(
             "vendor.valory.connections.abci.config.port",
             self.get_abci_port(i),
         )
--- a/tests/test_agents/base.py
+++ b/tests/test_agents/base.py
@@ -109,9 +109,10 @@ class BaseTestEnd2End(AEATestCaseMany, UseFlaskTendermintNode):
         dotted_path: str,
         value: Any,
         type_: Optional[str] = None,
+        aev: bool = True,
     ) -> Result:
         """Set config value."""
-        return super().set_config(dotted_path, value, type_, aev=True)
+        return super().set_config(dotted_path, value, type_, aev)
 
     def __set_extra_configs(self) -> None:
         """Set the current agent's extra config overrides that are skill specific."""
--- a/tests/test_agents/base.py
+++ b/tests/test_agents/base.py
@@ -23,6 +23,7 @@ import logging
 import os
 import time
 import warnings
+from dataclasses import dataclass
 from pathlib import Path
 from typing import Any, Dict, List, Optional, Tuple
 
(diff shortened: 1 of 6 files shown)


tests/test_agents/base.py
```diff
@@ -129,10 +145,6 @@ def __set_configs(self, i: int, nb_agents: int) -> None:
             "vendor.valory.connections.abci.config.host",
             ANY_ADDRESS,
         )
-        self.set_config(
-            "vendor.valory.connections.abci.config.host",
-            ANY_ADDRESS,
-        )
```
Adamantios, line 135: Not related to this PR, I removed it because it was a duplicate.

tests/test_agents/base.py
```diff
@@ -109,9 +124,10 @@ def set_config(
         dotted_path: str,
         value: Any,
         type_: Optional[str] = None,
+        aev: bool = True,
     ) -> Result:
         """Set config value."""
-        return super().set_config(dotted_path, value, type_, aev=True)
+        return super().set_config(dotted_path, value, type_, aev)
```
Adamantios, line 130: Not related to this PR, added `aev` as an argument.

tests/test_agents/base.py
```diff
@@ -37,6 +38,21 @@
...
+    """
+    Class for the necessary checks of a round during the tests.
+
+    name: is the name of the round for which the checks should be performed.
+    event: is the name of the event that is considered as successful.
+    n_periods: is the number of periods this event should appear for the check to be considered successful.
+    """
+
+    name: str
+    success_event: str = "DONE"
```
Adamantios, line 52: The default success event is still `DONE`; we do not need to type it every time, except if it is different.

tests/test_agents/test_apy_estimation_abci.py
```diff
@@ -23,27 +23,26 @@
...
+HAPPY_PATH = (
+    RoundChecks("collect_history"),
+    RoundChecks("transform"),
+    RoundChecks("preprocess"),
+    RoundChecks("randomness"),
+    RoundChecks("optimize"),
+    RoundChecks("train"),
+    RoundChecks("train", success_event="FULLY_TRAINED"),
+    RoundChecks("test"),
+    RoundChecks("estimate", n_periods=2, success_event="ESTIMATION_CYCLE"),
```
Adamantios, line 40: Here is an example of how we can use other successful events for the happy path's definition.

Karrenbelt: unrelated to this PR an sich, this sentence confuses me. What do you mean by "train on test data" after optimization?

Adamantios: Yes, I did not phrase that well. I meant simply on all the available data.

In the beginning, after optimizing, we split the data, train on the train split, and test on the test split. Then, if the performance is acceptable on the test split, we train on all the available data in order to create the final model. 

Hence, you can see why I incorrectly phrased it before.

DavidMinarsch (approved):

PR #1027, opened 2022-06-27
Fix/round id uniqueness

dvilelaf:
## Proposed changes

Enforce round id uniqueness

## Fixes

n/a

## Types of changes

## Checklist


packages/valory/skills/abstract_round_abci/abci_app_chain.py
```diff
@@ -39,6 +39,17 @@
 AbciAppTransitionMapping = Dict[AppState, AppState]
 
 
+def check_set_uniqueness(sets: Tuple) -> Optional[Any]:
+    """Checks that all elements in the set list are unique and not repeated among different sets"""
+    all_elements = set.union(*sets)
```
DavidMinarsch, line 44: why not just use `set.intersection`?

tests/test_skills/test_abstract_round_abci/test_abci_app_chain.py
```diff
@@ -255,3 +276,13 @@ def test_chain_two_negative_mapping_final_states(self) -> None:
...
             _ = chain((self.app1_class, self.app2_class), abci_app_transition_mapping)  # type: ignore
+
+    def test_chain_two_dupe(self) -> None:
+        """Test the AbciApp chain function."""
+
+        abci_app_transition_mapping: AbciAppTransitionMapping = {
+            self.round_1c: self.round_2a,
+            self.round_2c: self.round_1a,
+        }
+        with pytest.raises(AEAEnforceError):
```
DavidMinarsch, line 287: Also add error string, so test is more explicit

packages/valory/skills/abstract_round_abci/abci_app_chain.py
dvilelaf: I had left a comment to explain it but i don't see here anymore: set.intersection only works for chaining 2 apps. When chaining +2 apps, a round that is duplicated among 2 of those apps but not the other one won't appear on the intersection:
```
app1 = {a, b, c}
app2 = {c, d, e}
app3 = {f, g, h}
set.intersection(app1, app2, app3) = {}
```

DavidMinarsch (approved):

PR #1028, opened 2022-06-27
Fix/clean db history for current period

dvilelaf:
## Proposed changes

- Add a method that a dev can call in order to clean the current period history for every parameter in the db
- Add a new parameter that devs can optionally configure so current period history is cleaned every time we reset

## Fixes

If it fixes a bug or resolves a feature request, be sure to link to that issue.n/a
## Types of changes

## Checklist


packages/valory/skills/abstract_round_abci/base.py
```diff
@@ -629,13 +629,35 @@ def __repr__(self) -> str:
...
             for key in sorted(self._data.keys())[-cleanup_history_depth:]
         }
+        self.cleanup_current_histories(cleanup_history_depth_current)
+
+    def cleanup_current_histories(
+        self, cleanup_history_depth_current: Optional[int]
+    ) -> None:
+        """Reset the parameter histories for the current entry (period), keeping only the latest values for each parameter."""
+        if not cleanup_history_depth_current:
+            return
```
DavidMinarsch, line 653: would be cleaner to read if it was above on 646 and the method just does the cleaning

packages/valory/skills/abstract_round_abci/skill.yaml
```diff
@@ -9,8 +9,8 @@ fingerprint:
   README.md: QmYNXPppPSnW61ipqKNH8pyMh8qzTERwQaducYYq1dnBJb
```
DavidMinarsch, line 9: The new param needs to be set to null here otherwise it's not configurable (overrides only possible for existing keys)

DavidMinarsch (changes requested):

Karrenbelt (changes requested):
I'm not happy with this PR - neither with the state of the `AbciAppDB`.

The data structures don't make sense. Certain methods should not exist [[1](https://github.com/valory-xyz/open-autonomy/blob/1ff882422d800d142ea38c2380e33aa3458b1f10/packages/valory/skills/abstract_round_abci/base.py#L663), [2](https://github.com/valory-xyz/open-autonomy/blob/1ff882422d800d142ea38c2380e33aa3458b1f10/packages/valory/skills/abstract_round_abci/base.py#L624), [3](https://github.com/valory-xyz/open-autonomy/blob/1ff882422d800d142ea38c2380e33aa3458b1f10/packages/valory/skills/abstract_round_abci/base.py#L563)]. Now we have two cleanup methods `cleanup` and `cleanup_histories`. Why? The PR doesn't specify. Because in a single period it may otherwise grow unbounded? I am left to guess. But what's more, the purpose here should be defined and the structure designed accordingly. We use a dictionary, for something that should be a queue or a heap. Here's a red flag:

https://github.com/valory-xyz/open-autonomy/blob/1ff882422d800d142ea38c2380e33aa3458b1f10/packages/valory/skills/abstract_round_abci/base.py#L644

I suggest to define the structure of the data and its use cases, and then a data structure to accommodate. There are basically two axes. Name them and methods to operate on them accordingly. 

Also this. If I call create twice in a row I overwrite? Just no.
https://github.com/valory-xyz/open-autonomy/blob/1ff882422d800d142ea38c2380e33aa3458b1f10/packages/valory/skills/abstract_round_abci/base.py#L612

Why leave it up to the developer to call the method in case of a (repeated) event and not implement it as a configurable parameter at application level? Much clearer and easier no? 
`deque(maxlen=x)` or implement a fixed-size heap. How much here do we want to leave up to developer implementation and what should be a configurable systems parameter?

docs/api/skills/abstract_round_abci/base.md
```diff
@@ -2093,11 +2105,21 @@ Observe timestamp from last block.
 #### cleanup
 
 ```python
-def cleanup(cleanup_history_depth: int) -> None
+def cleanup(cleanup_history_depth: int, cleanup_history_depth_current: Optional[int] = None) -> None
```
Karrenbelt, line 2108: if you need more than two underscores for your variable naming, it's probably a bad idea :smiling_face_with_tear:

packages/valory/skills/abstract_round_abci/skill.yaml
dvilelaf: I have seen that this yaml specifies no parameters but all the other ones for other skills do, so for coherence I'll add the param in the same way.

Adamantios (commented):
The implementation looks functional, though I have two questions:
1. Why should we take care of emptying the current period's data if the developer decides not to use a period? Wouldn't it be simpler if the developer introduced a period for the skill? Why not add a section in the documentation instead warning about memory growth if the concept of a period is not introduced?
2. Since we decided to perform this change, wouldn't it be much simpler and less error-prone to choose a data structure that has a limited number of items it can keep, and discards the oldest entries?

dvilelaf:
> Why should we take care of emptying the current period's data if the developer decides not to use a period? Wouldn't it be simpler if the developer introduced a period for the skill? Why not add a section in the documentation instead warning about memory growth if the concept of a period is not introduced?

I wouldn't say we are taking care, at least not always: the developer would have three options now:
- Do nothing and let the current period's data grow unbounded as in the current implementation
- Call the new cleaning method himself, so he would be taking care of data
- Explicitly configure the parameter in skill.yaml so the current period's data is cleaned. This would be automatic but need configuration, so it does not happen without his knowledge


> Since we decided to perform this change, wouldn't it be much simpler and less error-prone to choose a data structure that has a limited number of items it can keep, and discards the oldest entries?

I considered that option (like using a deque), yes, but in the end I thought that it made more sense to either leave the dev decide when the cleaning happens or, if done automatically, synchronize the cleaning with the other cleanup to keep both coherent. I think that having one cleanup on reset and another one that is independent and not period related might introduce more confusion, but this of course is just an opinion. If in the end we decide to go the deque route, I think we should do it for both dimensions of the db: number of periods and current period history size.

Adamantios (approved):
LGTM & thanks for answering the questions.

DavidMinarsch:
> LGTM & thanks for answering the questions.

I think the documentation should be extended to explain the cleaning up

DavidMinarsch (changes requested):
Just some docs updates needed to explain cleanup and then this is ready

dvilelaf:
Opened an issue for the deque discussion: #1038

DavidMinarsch (approved):

packages/valory/skills/abstract_round_abci/base.py
```diff
@@ -522,8 +522,23 @@ class AbciAppDB:
         2: ...
     }
 
+    # Adding and removing data from the current period
+    --------------------------------------------------
     To update the current period entry, just call update() on the class. The new values will be appended to the current list for each updated parameter.
+
+    To clean up old data from the current period entry, call cleanup_current_histories(cleanup_history_depth_current), where cleanup_history_depth_current
+    is the amount of data that you want to keep after the cleanup. The newest cleanup_history_depth_current values will be kept for each parameter in the DB.
```
Adamantios, line 530: Should we also add a warning about the unbounded memory growth here?

DavidMinarsch (approved):

PR #1033, opened 2022-06-29
[WIP] Fix list representations in yaml files

angrybayblade:
## Proposed changes

This PR fixes list representations in config.yaml files.

## Fixes

[#182](https://github.com/valory-xyz/open-aea/issues/182)

## Types of changes

## Checklist

## Further comments


packages/valory/services/oracle_hardhat/service.yaml
```diff
@@ -31,7 +31,11 @@ models:
       args:
         url: https://api.coingecko.com/api/v3/simple/price
         api_id: coingecko
-        parameters: '[["ids", "bitcoin"], ["vs_currencies", "usd"]]'
+        parameters:
+        - - ids
+          - bitcoin
+        - - vs_currencies
+          - usd
```
angrybayblade, line 38: @DavidMinarsch , re https://github.com/valory-xyz/open-autonomy/pull/913#discussion_r893834818 I think you confusing actual python list with yaml string representation. `agent_vars` is a python object containing env var representations generated using model overwrites described in service config file. 

https://github.com/valory-xyz/open-autonomy/blob/main/packages/valory/services/oracle_hardhat/service.yaml#L28-L100

This logic added by @8ball030 is actually a patch rather than a temporary hack

```python
        for var_name, value in agent_vars.items():
            if any([isinstance(value, list), isinstance(value, dict)]):
                agent_vars[var_name] = json.dumps(value)
```

To explain this take this model override for example

```yaml
  - price_api:
      args:
        parameters: '[["ids", "bitcoin"], ["vs_currencies", "usd"]]'
 ```

or

```yaml
  - price_api:
      args:
        parameters:
         - - ids
           - bitcoin
         - - vs_currencies
           - usd
```

This will be stored as 

```python
{
  "SKILL_ORACLE_ABCI_MODELS_PRICE_API_ARGS_PARAMETERS": [["ids", "bitcoin"], ["vs_currencies", "usd"]]
 }
```

In the `agent_vars` object, so when we write this in a `docker-compose.yaml` file, we need to dump this as json representation rather than python string rep.

Karrenbelt: hmm. I'm not well familiar with the yaml library in python, but can't we use `yaml.safe_load`?

DavidMinarsch (commented):
Ok understand. Now please also look at `tests/helpers/deployments/test_deployments.py`. It contains a similarly encoded string and therefore the test needs updating to cover the intended behaviour. Please also add a second test to show that a wrongly configured service file (with string rather than list) will not be converted appropriately.

angrybayblade:
> Please also add a second test to show that a wrongly configured service file (with string rather than list) will not be converted appropriately.

This is not possible unless we use schema validation. Since the overrides are mostly skipped while schema validation their value can be anything the user provides. If we add validation in service loader it'll be specific to this service only since other services might have different list parameters.

DavidMinarsch (approved):

DavidMinarsch (approved):

PR #1036, opened 2022-06-29
[DO NOT MERGE]: Experiment with the e2e reset tests

Adamantios:
## Proposed changes
Experiment with the e2e tests for the reset, based on [this](https://trello.com/c/hGlIZxPP) ticket's description:

- Increase the check frequency for the `stop_string` in catchup e2e tests. We may be able to solve the flakiness if we tweak the e2e tests in such a way that the interrupted agent does not enter the reset round before the other agents enter.
- Do not raise `AddBlockError`. Log instead, because an `AddBlockError` may occur while resetting if an agent receives a request from its node after resetting, but it was sent before the reset.

## Fixes

n/a

## Types of changes

## Checklist

## Further comments

n/a


DavidMinarsch:
> * Increase the check frequency for the `stop_string` in catchup e2e tests. We may be able to solve the flakiness if we tweak the e2e tests in such a way that the interrupted agent does not enter the reset round before the other agents enter.

Ok

> * Do not raise `AddBlockError`. Log instead, because an `AddBlockError` may occur while resetting if an agent receives a request from its node after resetting, but it was sent before the reset.

Can we instead set a flag when we reset and then use this to discriminate whether to raise or log? Otherwise we will start missing other issues unrelated to reset with `AddBlockError`.

Adamantios:
> > * Increase the check frequency for the `stop_string` in catchup e2e tests. We may be able to solve the flakiness if we tweak the e2e tests in such a way that the interrupted agent does not enter the reset round before the other agents enter.
> 
> Ok
> 
> > * Do not raise `AddBlockError`. Log instead, because an `AddBlockError` may occur while resetting if an agent receives a request from its node after resetting, but it was sent before the reset.
> 
> Can we instead set a flag when we reset and then use this to discriminate whether to raise or log? Otherwise we will start missing other issues unrelated to reset with `AddBlockError`.

Yes, this was only the fastest way to experiment with that change. However, it caused Tendermint to panic so I need to further think about whether my initial assumption was correct and if this should be performed or not.

Adamantios:
Closed as we do not need to further experiment for the moment.

PR #1039, opened 2022-06-29
Temporary workaround for non indexed block response from SpookySwap's subgraph

Adamantios:
## Proposed changes

Temporarily workaround the non-indexed block response from SpookySwap's subgraph when trying to retrieve the ETH price for a block, by using the latest indexed bock instead. This is not optimal for the prediction.

## Fixes

n/a

## Types of changes

## Checklist

## Further comments

n/a


packages/valory/skills/apy_estimation_abci/behaviours.py
```diff
@@ -275,52 +292,130 @@ def _fetch_batch(self) -> Generator[None, None, None]:
...
+            res_context="indexing error that will be attempted to be handled",
+            keys=(0, "message"),
+            subgraph=self.context.spooky_subgraph,
+        )
+        if latest_indexed_block_error is None:
+            return None
+
+        match = re.match(NON_INDEXED_BLOCK_RE, latest_indexed_block_error)
+        if match is None:
+            return None
```
Karrenbelt, line 336: lets log this

packages/valory/skills/apy_estimation_abci/behaviours.py
```diff
@@ -275,52 +292,130 @@ def _fetch_batch(self) -> Generator[None, None, None]:
...
+
+        latest_indexed_block_number = match.group(1)
+        # we set the last digit to 0, so that it is more possible for the agents to reach to a consensus
+        # without requiring an extra round.
+        latest_indexed_block_number = int(
+            math.floor(float(latest_indexed_block_number) * 0.1) * 10
```
Karrenbelt, line 342: this I don't understand. Why zero? And what if the last digit is already zero? We cannot do a `-1` because? 

I like your proper mathematical programmer approach. Here's the python way 
`int(str(latest_indexed_block_number)[:-1] + '0')`

I would suggest: latest_indexed_block_number - latest_indexed_block_number % 10

tests/test_skills/test_apy_estimation_abci/test_apis.py
```diff
@@ -97,7 +100,43 @@ def test_eth_price(spooky_specs: SpecsType, eth_price_usd_q: str) -> None:
...
+        error_message = (
+            "Failed to decode `block.number` value: `subgraph QmPJbGjktGa7c4UYWXvDRajPxpuJBSZxeQK5siNT3VpthP has only "
+            "indexed up to block number 3730367 and data for block number 3830367 is therefore not yet available`"
+        )
+        match = re.match(NON_INDEXED_BLOCK_RE, error_message)
+        assert match is not None
+        assert match.groups() == ("3730367",)
+
+        error_message = "new message 3730367"
+        assert re.match(NON_INDEXED_BLOCK_RE, error_message) is None
```
Karrenbelt, line 139: can we also have a test that checks (in several attempts) to verify that the regex works on the current data we're retrieving from spookyswap? To ensure that we don't end up in a situation where they change their pattern without us noticing immediately

Karrenbelt (changes requested):
So I think this looks good. Would like one additional test I think for the regex and some more logging of reasons for returning `None`'s.

packages/valory/skills/apy_estimation_abci/behaviours.py
Adamantios: We cannot do a `-1` because the relative difference of the results will not change.

tests/test_skills/test_apy_estimation_abci/test_apis.py
Adamantios: This is what the above method does. I will refactor it to use the `NON_INDEXED_BLOCK_RE`.

packages/valory/skills/apy_estimation_abci/behaviours.py
Adamantios: I think this was the only place that returns `None` and does not immediately follow a `_handle_response` method.

Please notice how `_handle_response` will always log the error if `None` is returned.

Karrenbelt (approved):
LGTM

DavidMinarsch (approved):

PR #1042, opened 2022-07-01
Hard reset safely

Adamantios:
## Proposed changes

We need to check if the `unsafe-reset-all` exited successfully before returning a successful response.

## Fixes

n/a

## Types of changes

## Checklist

## Further comments

Some things were extracted into a `conftest.py` file from the `test_app.py`.


deployments/Dockerfiles/localnode/app.py
```diff
@@ -218,7 +218,10 @@ def hard_reset() -> Tuple[Any, int]:
             if IS_DEV_MODE:
                 period_dumper.dump_period()
 
-            tendermint_node.prune_blocks()
+            return_code = tendermint_node.prune_blocks()
+            if return_code:
+                tendermint_node.start(start_monitoring=perform_monitoring)
+                raise RuntimeError("Could not perform `unsafe-reset-all` successfully!")
```
Adamantios, line 224: We now raise if `prune_block` fails and restart the node. The error will be handled below and the status will be `False`, accompanied by a failure message and the exception's message.

tests/test_deployments/test_app.py
```diff
@@ -254,13 +221,16 @@ class TestTendermintHardResetServer(BaseTendermintServerTest):
...
         """Test hard reset"""
-        with self.app.test_client() as client:
+        with self.app.test_client() as client, mock.patch.object(
+            TendermintNode, "prune_blocks", return_value=int(prune_fail)
+        ):
             response = client.get("/hard_reset")
             data = response.get_json()
             assert response.status_code == 200
-            assert data["status"] is True
+            assert data["status"] is not prune_fail
```
Adamantios, line 233: Test now also covers prune failure.

deployments/Dockerfiles/localnode/app.py
Karrenbelt, line 224: why start the Tendermint node right before raising a `RuntimeError`?

Adamantios: The fact that we did not manage to perform the `unsafe-reset-all` does not mean that we should let the node stopped.

tests/test_deployments/conftest.py
```diff
@@ -0,0 +1,78 @@
+# -*- coding: utf-8 -*-
+# ------------------------------------------------------------------------------
```
Karrenbelt, line 2: is it customary / acceptable / good practice to have nested conftest files? I simply don't know but I guess so since that is where `autouse=True` goes to look for fixtures.

DavidMinarsch (approved):

PR #1043, opened 2022-07-01
extending simple ABCI tests

Karrenbelt:
## Proposed changes

We had no catch up behaviour test for the simple ABCI skill. Currently I am debugging the Tendermint issue we have with catchup behaviour, for which this test is useful as opposed to having to run a more complicated application.

I have also implemented tests with multiple agents restarting, and who's restart can be delayed. Outside of further testing the robustness of the system these will be useful to test shutdown of the ABCIApplication that we are about to implement.

## Fixes

## Types of changes

## Checklist

## Further comments


tests/test_agents/base.py
```diff
@@ -400,21 +400,7 @@ def check_aea_messages(self) -> None:
...
-        self.health_check(
-            max_retries=self.HEALTH_CHECK_MAX_RETRIES,
-            sleep_interval=self.HEALTH_CHECK_SLEEP_INTERVAL,
-        )
-        self.check_aea_messages()
-        self.terminate_processes()
-
-
-class BaseTestEnd2EndAgentCatchup(BaseTestEnd2End):
+class BaseTestEnd2EndExecution(BaseTestEnd2End):
```
Karrenbelt, line 403: the CatchUp test was merely an extension of the normal execution test.

tests/test_agents/base.py
```diff
@@ -434,29 +420,43 @@ class BaseTestEnd2EndAgentCatchup(BaseTestEnd2End):
...
-        if not hasattr(self, "stop_string"):
-            pytest.fail("'stop_string' is a mandatory argument.")
+    nb_nodes: int = 0  # number of agents with tendermint nodes
+
+    # configuration specific for agent restart
+    stop_string: str  # mandatory argument if n_terminal > 0
+    n_terminal: int = 0  # number of agents to be restarted
+    wait_to_kill: int = 0  # delay the termination event
+    restart_after: int = 60  # how long to wait before restart
+    wait_before_stop: int = 15  # how long to check logs
```
Karrenbelt, line 430: the only new features here are `n_terminal` and `wait_to_kill`

tests/test_agents/test_simple_abci.py
```diff
@@ -35,41 +35,78 @@
...
+# normal execution
+class BaseSimpleABCITest(BaseTestEnd2EndExecution):
     """Test the ABCI simple_abci skill with only one agent."""
 
     agent_package = "valory/simple_abci:0.1.0"
     skill_package = "valory/simple_abci:0.1.0"
     wait_to_finish = 80
     happy_path = HAPPY_PATH
     strict_check_strings = STRICT_CHECK_STRINGS
-    use_benchmarks = True
```
Karrenbelt, line 49: this was not used anywhere

tests/test_agents/base.py
DavidMinarsch, line 403: Don't rename it please; unnecessary diffs

tests/test_agents/test_simple_abci.py
```diff
@@ -35,41 +35,78 @@
...
+    """Test simple_abci skill with four agents; one restarting on `select_keeper_at_startup`."""
+
+    stop_string = "select_keeper_at_startup"
+
+
+@pytest.mark.parametrize("nb_nodes", (4,))
+class TestSimpleABCIFourAgentsCatchupResetAndPause(BaseSimpleABCITestCatchup):
+    """Test simple_abci skill with four agents; one restarting on `reset_and_pause`."""
+
+    stop_string = "reset_and_pause"
```
DavidMinarsch, line 97: there's no catchup here?

tests/test_agents/test_simple_abci.py
```diff
@@ -35,41 +35,78 @@
...
+    """Test the ABCI simple_abci skill with four agents and two restarting."""
+
+    n_terminal = 2
+
+
+@pytest.mark.parametrize("nb_nodes", (4,))
+class TestSimpleABCIFourAgentsFourAgentCatchup(BaseSimpleABCITestCatchup):
+    """Test the ABCI simple_abci skill with four agents and four restarting."""
+
+    n_terminal = 4
```
DavidMinarsch, line 112: not sure this makes much sense. The Tendermint chain will not make progress if less than BFT threshold is present.

Karrenbelt: this is true, but that is exactly the scenario we're facing when we (after prolonged period of stalling) want to consider a shutdown. I will remove the one with 4 agent restarting since it offers nothing additional beyond the two agents restarting

tests/test_agents/base.py
```diff
@@ -434,29 +420,43 @@ class BaseTestEnd2EndAgentCatchup(BaseTestEnd2End):
...
+
+    def check(self) -> None:
+        """Check pre-conditions of the test"""
+        if self.n_terminal > self.nb_nodes:
+            fail_msg = "Cannot terminate {nb_nodes} out of {n_terminal} agents:"
+            pytest.fail(
+                fail_msg.format(nb_nodes=self.nb_nodes, n_terminal=self.n_terminal)
+            )
```
Adamantios, line 438: We should instead check that `n_terminal` is not bigger than the faulty threshold.

tests/test_agents/base.py
```diff
@@ -465,22 +465,27 @@ def _stop_and_restart_last_agent(self, nb_agents: int) -> None:
...
 
-        # restart agent
-        logging.info("Restart the agent")
-        self._launch_agent_i(-1, nb_agents)
-        self.check_aea_messages()
-        self.terminate_processes()
+        # restart agents
+        for i in range(self.n_terminal):
+            logging.info(f"Restart the agent {~i}")
+            self._launch_agent_i(~i, self.nb_nodes)
```
Adamantios, line 491: Why not `-i`?

tests/test_agents/base.py
```diff
@@ -465,22 +465,27 @@ def _stop_and_restart_last_agent(self, nb_agents: int) -> None:
...
 
-        # immediately terminate the agent
-        self.terminate_agents(self.processes[-1], timeout=0)
+        # immediately terminate the agents - sequentially
         # don't pop before termination, seems to lead to failure!
-        self.processes.pop()
+        for _ in range(self.n_terminal):
+            self.terminate_agents(self.processes[-1], timeout=0)
+            self.processes.pop()
+            logging.info("Last agent stopped")
```
Adamantios, line 480: We may want to stop popping the last agent. It only adds unnecessary complexity. We could instead:

```suggestion
        for i in range(self.n_terminal):
            self.terminate_agents(self.processes[i], timeout=0)
            self.processes.pop(0)
            logging.info(f"Agent {i} stopped")
```

Karrenbelt: I agree with that, but I did not want too change too much at once in fear or breaking things and slowing down

tests/test_agents/base.py
Karrenbelt: not exactly the same, would need to `i - 1`. Best idea, though, is to not loop backwards in the first place, as you suggested we don't need to start at the end, or use reverse(range(1, len(x) + 1)) or so (which is also not nice).

tests/test_agents/base.py
```diff
@@ -252,34 +245,35 @@ def prepare(self, nb_nodes: int) -> None:
...
 
         # run 'aea install' in only one AEA project, to save time
-        self.set_agent_context(self.__get_agent_name(0, nb_nodes))
+        self.set_agent_context(self.__get_agent_name(0))
         self.run_install()
 
     def prepare_and_launch(self, nb_nodes: int) -> None:
         """Prepare and launch the agents."""
-        self.processes = []
+        self.processes = dict.fromkeys(range(nb_nodes))
```
Karrenbelt, line 253: I refactored this such that processes are now a mapping and we don't allow for negative indexing anymore. This will ensure that whatever order you decide to terminate agents, and whatever other order you decide to restart them, the processes should be in the correct order. 

To illustrate the issue, terminating multiple agents and subsequently restarting them, either the one that terminated first will need to be restarted last, or the order of self.processes will be mixed up - e.g. when agent 4 terminates, then 3, then 4 restarts, then 3, we end up with processes [0, 1, 3, 2] because of appending to the list. Also popping first agents first one would then need to prepend to the list, and in case of multiple same issue might arise, using a mapping prevents such occurrence.

tests/test_agents/base.py
```diff
@@ -216,20 +216,13 @@ def __set_configs(self, i: int, nb_agents: int) -> None:
...
         """Get the ith agent's name."""
-        if i < 0:
-            i = nb_agents + i
-        if i < 0:
-            raise ValueError(
-                f"Incorrect negative indexing. {i} was given, but {nb_agents} agents are available!"
-            )
-        agent_name = f"agent_{i:05d}_{nb_agents}agents_run"
-        return agent_name
+        return f"agent_{i:05d}"
```
Karrenbelt, line 221: we don't allow for negative indexing anymore

tests/test_agents/base.py
```diff
@@ -434,53 +413,73 @@ class BaseTestEnd2EndAgentCatchup(BaseTestEnd2End):
...
+            logging.info("Waiting to terminate agents")
+            time.sleep(self.wait_to_kill)
+
+        # terminate the agents - sequentially
         # don't pop before termination, seems to lead to failure!
-        self.processes.pop()
+        for i in range(self.n_terminal):
+            self.terminate_agents(self.processes[i], timeout=0)
+            self.processes.pop(i)
+            logging.info("Last agent stopped")
```
Karrenbelt, line 474: need to change this still...

tests/test_agents/base.py
```diff
@@ -434,53 +413,75 @@ class BaseTestEnd2EndAgentCatchup(BaseTestEnd2End):
...
+            time.sleep(self.wait_to_kill)
+
+        # terminate the agents - sequentially
         # don't pop before termination, seems to lead to failure!
-        self.processes.pop()
+        for i in range(self.n_terminal):
+            agent_name = self.__get_agent_name(i)
+            self.terminate_agents(self.processes[i], timeout=0)
+            self.processes.pop(i)
+            logging.info(f"Terminated {agent_name}")
```
Karrenbelt, line 475: we now iterate starting from 0. We pop from a dict.

tests/test_agents/base.py
```diff
@@ -434,53 +413,75 @@ class BaseTestEnd2EndAgentCatchup(BaseTestEnd2End):
...
-        # restart agent
-        logging.info("Restart the agent")
-        self._launch_agent_i(-1, nb_agents)
-        self.check_aea_messages()
-        self.terminate_processes()
+        # restart agents
+        for i in range(self.n_terminal):
+            agent_name = self.__get_agent_name(i)
+            self._launch_agent_i(i)
+            logging.info(f"Restarted {agent_name}")
```
Karrenbelt, line 487: here we now also iterate from 0. Since we use a mapping the "order" of the processes is guaranteed to remain correct.

tests/test_agents/base.py
```diff
@@ -363,16 +357,17 @@ def missing_from_output(  # type: ignore
 
         return missing_strict_strings, missing_round_strings
 
-    @staticmethod
     def __check_missing_strings(
-        missing_strict_strings: List[str], missing_round_strings: List[str], i: int
+        self,
```
Adamantios, line 361: Do we need `self`? Isn't `__get_agent_name` static?

Adamantios (commented):
LGTM, let's wait for the e2e tests to run first.

tests/test_agents/base.py
Karrenbelt: it very well might be, but it's (needlessly) part of a class so need it to reference it using `self` or `cls`

Karrenbelt: alright so name mangling got me good. Let's not use it unless we have a good reason for it :smiling_face_with_tear:

Karrenbelt:
still no good
```
2022-07-05T20:46:00.9611020Z FAILED tests/test_agents/test_oracle_abci.py::TestAgentCatchup::test_run[4]
2022-07-05T20:46:00.9611315Z FAILED tests/test_agents/test_register_reset_abci.py::TestTendermintResetInterruptNoRejoin::test_run[4]
2022-07-05T20:46:00.9611576Z FAILED tests/test_agents/test_register_reset_abci.py::TestTendermintResetInterrupt::test_run[4]
2022-07-05T20:46:00.9611810Z FAILED tests/test_agents/test_register_reset_abci.py::TestTendermintReset::test_run[4]
```

tests/test_agents/test_register_reset_abci.py
```diff
@@ -35,8 +32,11 @@
 )
 
 
+@pytest.mark.e2e
+@pytest.mark.integration
+@pytest.mark.flaky(reruns=3)
```
Karrenbelt, line 37: marked all register reset tests as flaky because they are

tests/test_agents/test_register_reset_abci.py
```diff
@@ -117,5 +128,5 @@ class TestTendermintResetInterruptNoRejoin(TestTendermintResetInterrupt):
     wait_to_finish = 200
     # set the restart to a value so that the agent never rejoins, in order to test the impact to the rest of the agents
     restart_after = wait_to_finish
-    # check if we manage to reset with Tendermint with the rest of the agents; 3rd agent will not rejoin in this test
-    exclude_from_checks = [3]
+    # check if we manage to reset with Tendermint with the rest of the agents; first agent will not rejoin in this test
+    exclude_from_checks = [0]
```
Karrenbelt, line 132: we check for the `stop_string` in the last agent, and once that has appeared we restart starting from 0 (hence in this test specifically [0] is excluded from the `check_aea_messages`.

tests/test_agents/base.py
```diff
@@ -216,20 +216,13 @@ def __set_configs(self, i: int, nb_agents: int) -> None:
...
         """Get the ith agent's name."""
-        if i < 0:
-            i = nb_agents + i
-        if i < 0:
-            raise ValueError(
-                f"Incorrect negative indexing. {i} was given, but {nb_agents} agents are available!"
-            )
-        agent_name = f"agent_{i:05d}_{nb_agents}agents_run"
-        return agent_name
+        return f"agent_{i:05d}"
```
DavidMinarsch, line 221: why are we no longer logging total number of agents too?

tests/test_agents/test_register_reset_abci.py
DavidMinarsch: Please open an issue!

DavidMinarsch (approved):

tests/test_agents/base.py
Karrenbelt: because it is defined at the test level, it will be reported in the name of the test already and adding it to the agent name itself we need to pass around another variable

Adamantios (approved):

angrybayblade (approved):
LGTM

PR #1045, opened 2022-07-01
feat: add-bfs-check

jmoreira-valory:
## Proposed changes

Added a BFS-like check for FSM specifications. This is required because the previous checks did not catch all possible erroneous specifications. Namely, a specification like

start_states: [A]

(A, ev1): B
(B, ev1): C
(D, ev1): A
(X, ev1): Y
(Y, ev1): X

would have passed the checks, but it clearly contains 2 disjoint components {A,B,C} and {X,Y}, and the latter one is unreachable, yet it does not have any orphan state. Thus, I removed the concept of "orphan state", as it is grouped into the more general concept of "reachable state".

## Fixes

## Types of changes

## Checklist

## Further comments

### DIFF
--- a/autonomy/analyse/abci/app_spec.py
+++ b/autonomy/analyse/abci/app_spec.py
@@ -23,6 +23,7 @@ import json
 import logging
 import re
 import sys
+from collections import defaultdict
 from itertools import product
 from pathlib import Path
 from typing import Any, Dict, List, OrderedDict, Set, TextIO, Tuple
@@ -63,12 +64,13 @@ class DFA:
         transition_func_states = transition_func_in_states.copy()
         transition_func_states.update(transition_func.values())  # type: ignore
 
-        orphan_states = states - (start_states | set(transition_func.values()))
+        reachable_states = DFA._get_reachable_states(start_states, transition_func)
+        unreachable_states = states - reachable_states
 
         error_strings = []
-        if orphan_states:
+        if unreachable_states:
             error_strings.append(
-                f" - DFA spec. contains orphan states: {orphan_states}."
+                f" - DFA spec. contains unreachable states: {unreachable_states}."
             )
         if not transition_func_states.issubset(states):
             error_strings.append(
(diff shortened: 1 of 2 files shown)


autonomy/analyse/abci/app_spec.py
```diff
@@ -205,6 +207,32 @@ def generate(self) -> Dict[str, Any]:
...
+        graph: Dict[str, Set[str]] = defaultdict(set)
+        for (s1, _), s2 in transition_func.items():
+            graph[s1].add(s2)
+
+        visited: Set[str] = start_states.copy()
+        queue: List[str] = []
+        queue.extend(list(start_states))
+
+        while queue:
+            s = queue.pop(0)
```
Karrenbelt, line 227: nice nice. Popping first element of a list is no good in python though. Use an actual queue: 
```
from collections import deque
queue = deque(list(start_states))
while queue:
    node = queue.popleft()
```

or, just call it `stack` and `stack.pop()` the last element as the ordering here is irrelevant.

Karrenbelt (approved):
LGTM

so I don't know how much more checks one might want to add in the future, but its good to be aware that a tool like [NetworkX](https://networkx.org/) or so might come in handy for graph related analyses. It is way too much for what we have here now, just want to point it out as it contains a lot of useful functionalities.

docs/autonomy.md
```diff
@@ -199,7 +199,6 @@ Usage: autonomy analyse abci check-app-specs [OPTIONS]
 Options:
   --check-all          Check all available definitions.
   --packages-dir PATH  Path to packages directory; Use with `--check-all` flag
-  --mermaid            Mermaid file.
```
DavidMinarsch, line 202: Do we no longer support this?

DavidMinarsch (approved):

docs/autonomy.md
jmoreira-valory: We support yaml, json or mermaid for **output**, and yaml json for input. Thus the removal of this line here.

autonomy/analyse/abci/app_spec.py
jmoreira-valory: Thanks for the advice. Will fix in an upcoming PR.

PR #1049, opened 2022-07-04
Fix agent sync mode

Adamantios:
## Proposed changes

When sync mode ends and the agent is successfully synced, we should update the app's state using the retrieved app hash.

## Fixes

n/a

## Types of changes

## Checklist

## Further comments

n/a


packages/valory/skills/abstract_round_abci/behaviour_utils.py
```diff
@@ -95,6 +95,8 @@
 HEIGHT_OFFSET_MULTIPLIER = 0.01
 NON_200_RETURN_CODE_DURING_RESET_THRESHOLD = 3
 GENESIS_TIME_FMT = "%Y-%m-%dT%H:%M:%S.%fZ"
+ROOT_HASH = "726F6F743A3"
+RESET_HASH = "72657365743A3"
```
Adamantios, line 99: The root hash is of the form:
`726F6F743A3{ROUND_COUNT}72657365743A3{RESET_INDEX}`
which is the hash that Tendermint produces from:
`root:{ROUND_COUNT}reset:{RESET_INDEX}`

tests/test_skills/test_abstract_round_abci/test_behaviours_utils.py
```diff
@@ -580,25 +610,52 @@ def test_async_act_wrapper_communication(
...
+        with mock.patch.object(logging, "info") as log_mock, mock.patch.object(
+            BaseBehaviour,
+            "_get_status",
+            _get_status_patch_wrapper(0, expected_round_count, expected_reset_index),
+        ):
             gen = self.behaviour.async_act_wrapper()
-            try_send(gen)
-            try_send(gen)
+            for __ in range(3):
+                try_send(gen)
```
Karrenbelt, line 641: why 3?

Adamantios: Notice how the `yield` was incorrectly placed [here](https://github.com/valory-xyz/open-autonomy/pull/1049/files?file-filters%5B%5D=.py&show-viewed-files=true#diff-63b131ca6614d6666a3eeb925cb9b9e71b63f27a68ea748d9db7f8bae21f0bc5L375-L378). Now it is placed before the `return`, is reachable, and another `try_send` is required.

Karrenbelt:
we check_sync, then if we are synced then we still need to sync state by setting the round count? So we sync up in terms of block height. And then we sync up state in terms of round count and reset index. Should we do the same on the Terdermint-side though? In the Tendermint flask server we set fast-sync (now named block-sync in 35.7) to false, and state sync we don't touch but defaults to false (source: [Tendermint documentation](https://docs.tendermint.com/v0.34/tendermint-core/configuration.html))
can you tell me how these two features, on the Tendermint side and our our ABCI Application side, are related?

Adamantios:
> we check_sync, then if we are synced then we still need to sync state by setting the round count? So we sync up in terms of block height. And then we sync up state in terms of round count and reset index. Should we do the same on the Terdermint-side though? In the Tendermint flask server we set fast-sync (now named block-sync in 35.7) to false, and state sync we don't touch but defaults to false (source: [Tendermint documentation](https://docs.tendermint.com/v0.34/tendermint-core/configuration.html)) can you tell me how these two features, on the Tendermint side and our our ABCI Application side, are related?

The fact that we sync up with the blockchain does not mean that we also have the correct application state. The application state is defined by the abci app's developer and also determines the generation of the app hash. When we sync with the blockchain, it means that we no longer lag behind the other agents and the application's state should have been updated.

We have chosen a simple application state for the time being, which is a combination of the round count and the times we have reset so far.

Fast sync (I think renamed to block sync after your tm update branch gets merged) and state sync are not to be confused with our state. They are different methods to sync faster with the blockchain.

Karrenbelt:
> > we check_sync, then if we are synced then we still need to sync state by setting the round count? So we sync up in terms of block height. And then we sync up state in terms of round count and reset index. Should we do the same on the Terdermint-side though? In the Tendermint flask server we set fast-sync (now named block-sync in 35.7) to false, and state sync we don't touch but defaults to false (source: [Tendermint documentation](https://docs.tendermint.com/v0.34/tendermint-core/configuration.html)) can you tell me how these two features, on the Tendermint side and our our ABCI Application side, are related?
> 
> The fact that we sync up with the blockchain does not mean that we also have the correct application state. The application state is defined by the abci app's developer and also determines the generation of the app hash. When we sync with the blockchain, it means that we no longer lag behind the other agents and the application's state should have been updated.
> 
> We have chosen a simple application state for the time being, which is a combination of the round count and the times we have reset so far.
> 
> Fast sync (I think renamed to block sync after your tm update branch gets merged) and state sync are not to be confused with our state. They are different methods to sync faster with the blockchain.

so that sounds exactly as I imaged it, thus happy to hear. Though that said, I think it would be good to add some documentation to the code to clarify this because someone else will wonder the same and might get confused (I was myself confronted with such scenario when faced with the question why the TendermintHandler I wrote does not use the ABCI connection)

Adamantios:
Need to investigate the catchup test.

packages/valory/skills/abstract_round_abci/behaviour_utils.py
```diff
@@ -98,7 +98,7 @@
 GENESIS_TIME_FMT = "%Y-%m-%dT%H:%M:%S.%fZ"
 ROOT_HASH = "726F6F743A3"
 RESET_HASH = "72657365743A3"
-APP_HASH_RE = fr"{ROOT_HASH}(\d+){RESET_HASH}(\d+)"
+APP_HASH_RE = fr"{ROOT_HASH}\d+{RESET_HASH}(\d+)"
```
Karrenbelt, line 101: oof, I should have caught this!

Karrenbelt (approved):
LGTM

There are 3 failing e2e tests:
```
 FAILED tests/test_agents/test_register_reset_abci.py::TestTendermintResetInterrupt::test_run[4]
FAILED tests/test_agents/test_registration.py::TestRegistrationStartUpFourAgents::test_run[4]
FAILED tests/test_agents/test_registration.py::TestRegistrationStartUpFourAgentsCatchUp::test_run[4]
```

the first one I believe to be flaky, the other two are ACN-related

packages/valory/skills/abstract_round_abci/behaviour_utils.py
```diff
@@ -634,6 +638,40 @@ def async_act_wrapper(self) -> Generator:
...
+        which is a combination of the round count and the times we have reset so far.
+        The round count will be automatically updated by the framework while replaying the missed rounds,
+        however, the reset index needs to be manually updated.
+
+        Tendermint's block sync and state sync are not to be confused with our application's state;
+        they are different methods to sync faster with the blockchain.
+
+        :param app_hash: the app hash from which the state will be updated.
+        """
+        if app_hash == "":
```
DavidMinarsch, line 660: Can we turn this into a constant? Otherwise this magic value is too implicit imo

DavidMinarsch (approved):
Didn't check the tests but rest lgtm

packages/valory/skills/abstract_round_abci/behaviour_utils.py
Adamantios: Done in 0a7eecae6aea2e70f4d0cbb9abc6af4a499c6eaa.

DavidMinarsch (approved):

PR #1051, opened 2022-07-05
[WIP]: Add a margin in the consensus threshold for the reset and pause round

Adamantios:
## Proposed changes

We may want to be more strict for the reset round and not start the round once the majority threshold is hit, but when the majority threshold + margin is hit, in order to allow for random interruptions. Margin should be relative to the number of participants and is the number of agents that we can afford to lose while resetting, before losing the majority. Otherwise, starting the reset round with a number of agents equal to the majority seems risky.

## Fixes

n/a

## Types of changes

## Checklist

## Further comments

[Trello ticket](https://trello.com/c/hGlIZxPP).


Adamantios:
3150e14 must be reverted when https://github.com/valory-xyz/open-autonomy/pull/1052 is merged.

Adamantios:
This PR is not relevant anymore, since the reset and pause e2e tests were failing because of an incorrect configuration of the persistent peers. This has now been fixed in #1062.

Issue #1003, opened 2022-06-23
custom __hash__ implementations


Karrenbelt:
### Subject of the issue

We overwrite the `__hash__` method on payloads in a problematic way. The hash of an object should never be allowed to change during the lifetime of the instance. We use it to obtain a hashable representation of the content of the payload. However, if ever any content in the payload were to change while it is already part of a hashset of hashmap, then this value will change, thus violating the premise. Potential problems can be hard to debug.

affects:
`BaseLiquidityRebalancingPayload`
`BaseAPYPayload`
`BaseTxPayload`
`BaseHelloWorldAbciPayload`
`BaseSimpleAbciPayload`

I see no need for us to overwrite it and suggest we implement a regular custom method

References:
[python documentation](https://docs.python.org/3/reference/datamodel.html#object.__hash__)
[example](https://eng.lyft.com/hashing-and-equality-in-python-2ea8c738fb9d)

### Your environment
- Python version: [e.g. 3.7.2]
- Package Version [e.g. 0.1.2]
- Anything else you consider helpful.

### Steps to reproduce
Tell us how to reproduce this issue.

### Expected behaviour
Tell us what should happen

### Actual behaviour
Tell us what happens instead

PR #1052, opened 2022-07-06
Test for `check_majority_possible`

Adamantios:
## Proposed changes

Adds a test for the `check_majority_possible` classmethod when the attribute values of a payload we are interested in are equal.

## Fixes

n/a

## Types of changes

## Checklist

## Further comments

n/a

### DIFF
--- a/tests/test_skills/test_abstract_round_abci/test_base.py
+++ b/tests/test_skills/test_abstract_round_abci/test_base.py
@@ -111,7 +111,7 @@ class PayloadC(BasePayload):
 
 
 class PayloadD(BasePayload):
-    """Payload class for payload type 'C'."""
+    """Payload class for payload type 'D'."""
 
     transaction_type = PayloadEnumB.A
 
--- a/tests/test_skills/test_abstract_round_abci/test_base.py
+++ b/tests/test_skills/test_abstract_round_abci/test_base.py
@@ -72,6 +72,7 @@ class PayloadEnum(Enum):
     A = "A"
     B = "B"
     C = "C"
+    DUMMY = "DUMMY"
 
     def __str__(self) -> str:
         """Get the string representation."""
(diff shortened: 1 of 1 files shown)


tests/test_skills/test_abstract_round_abci/test_base.py
```diff
@@ -914,6 +941,26 @@ def test_check_majority_possible_passes_when_vote_set_nonempty_and_check_passes(
...
+        - 3 voters have the same attribute value in their payload
+        """
+        AbstractRound.check_majority_possible(
+            {
+                "voter_1": DummyPayload("voter_1", 0),
+                "voter_2": DummyPayload("voter_2", 0),
+                "voter_3": DummyPayload("voter_3", 0),
+            },
+            4,
+        )
```
Adamantios, line 962: Shouldn't this pass? We have 3/4 agents who have voted for the same thing, but we get:
```
packages.valory.skills.abstract_round_abci.base.ABCIAppException: cannot reach quorum=3, number of remaining votes=1, number of most voted item's votes=1
```

Karrenbelt: yes. The error is here

https://github.com/valory-xyz/open-autonomy/blob/8ca6c61ccc810e47f6dc1731bcda3257be63486e/packages/valory/skills/abstract_round_abci/base.py#L1029

and should be 
`largest_nb_votes = vote_count.most_common()[0][1]`

Karrenbelt: ok so I jumped the gun. The issue arised because of `__eq__`  here, which includes the sender attribute, and hence they do not compare equal

https://github.com/valory-xyz/open-autonomy/blob/8ca6c61ccc810e47f6dc1731bcda3257be63486e/packages/valory/skills/abstract_round_abci/base.py#L263-L272

solution thus is to access the .data attribute. 
`Counter([tuple(sorted(v.data)) for v in  votes_by_participant.values()])`

Please do **not** use the `__hash__` method like such, as it is scheduled to be removed
`Counter(map(hash, votes_by_participant.values()))`

Karrenbelt: ticket is here https://github.com/valory-xyz/open-autonomy/issues/1003

Karrenbelt:
reasoning about the consequence of this bug, it is obvious that `largest_nb_votes` was always `1` because both the `.sender` and the `.id_` attribute were present on payloads. In consequence `check_majority_possible` would be triggered to raise more often than it should, because we misjudged the possibility of being able to obtain a majority.

Karrenbelt:
failures on macOS (varying per run):
```
error An unexpected error occurred: "https://registry.npmjs.org/rxjs/-/rxjs-6.6.7.tgz: ESOCKETTIMEDOUT".
```
```
ERROR tests/test_deployments/test_app.py::TestTendermintHardResetServer::test_hard_reset[True]
ERROR tests/test_deployments/test_app.py::TestTendermintHardResetServer::test_hard_reset[True]
```
```
ERROR tests/test_deployments/test_app.py::TestTendermintGentleResetServer::test_gentle_reset
```

DavidMinarsch (approved):

PR #1053, opened 2022-07-06
Fix: payload majority count

Karrenbelt:
## Proposed changes

A bug was detected by the great @Adamantios regarding the counting of payloads in `AbstractRound.check_majority_possible`. It came about by the fact that the `__eq__` operator on `BaseTxPayload` includes the sender, as well as a unique `id_`, and hence every element was unique and the `largest_nb_votes` equal to `1`. 

## Fixes

We use `sorted(tuple(sorted(payload.data.items()))` 

Furthermore also changed the typing from `Dict[Any, Any]` to `Dict[str, BaseTxPayload]`. Several tests were using a string as a payload and hence the bug went undetected, since mypy did not flag the strings being passed.

## Types of changes

## Checklist

## Further comments


packages/valory/skills/abstract_round_abci/base.py
```diff
@@ -944,9 +944,9 @@ def check_allowed_tx_type(self, transaction: Transaction) -> None:
     @classmethod
     def check_majority_possible_with_new_voter(
         cls,
-        votes_by_participant: Dict[Any, Any],
-        new_voter: Any,
-        new_vote: Any,
+        votes_by_participant: Dict[str, BaseTxPayload],
+        new_voter: str,
+        new_vote: BaseTxPayload,
```
Karrenbelt, line 949: added specific types

packages/valory/skills/abstract_round_abci/base.py
```diff
@@ -944,9 +944,9 @@ def check_allowed_tx_type(self, transaction: Transaction) -> None:
...
         nb_participants: int,
         exception_cls: Type[ABCIAppException] = ABCIAppException,
     ) -> None:
```
Karrenbelt, line 952: added specific types here and in matching types in `check_majority_possible` and `is_majority_possible`

packages/valory/skills/abstract_round_abci/base.py
```diff
@@ -1025,7 +1025,8 @@ def check_majority_possible(
         if len(votes_by_participant) == 0:
             return
 
-        vote_count = Counter(votes_by_participant.values())
+        votes = votes_by_participant.values()
+        vote_count = Counter(tuple(sorted(v.data.items())) for v in votes)
```
Karrenbelt, line 1029: we're still depending on the fact that the `.data` attribute is correctly implemented by the developer, as the base class `BaseTxPayload` always returns an empty dict.  I am not a fan of this empty dict, but remember there was a reason for this - please inform me of this reason so we can reconsider.

tests/test_skills/test_abstract_round_abci/test_base.py
```diff
@@ -939,7 +939,9 @@ def test_check_majority_possible_passes_when_vote_set_nonempty_and_check_passes(
         - the threshold is 2
         - the other voter can vote for the same item of the first voter
         """
-        AbstractRound.check_majority_possible({"voter": "item"}, 2)
+        votes = {"alice": DummyPayload("alice", True)}
+        votes_by_participants: Dict[str, BaseTxPayload] = votes
+        AbstractRound.check_majority_possible(votes_by_participants, 2)
```
Karrenbelt, line 944: here and below I have changed and payload that was a string into a `DummyPayload`

tests/test_skills/test_abstract_round_abci/test_base.py
```diff
@@ -939,7 +939,9 @@ def test_check_majority_possible_passes_when_vote_set_nonempty_and_check_passes(
         - the threshold is 2
         - the other voter can vote for the same item of the first voter
         """
-        AbstractRound.check_majority_possible({"voter": "item"}, 2)
+        votes = {"alice": DummyPayload("alice", True)}
+        votes_by_participants: Mapping[str, BaseTxPayload] = votes
+        AbstractRound.check_majority_possible(votes_by_participants, 2)
```
Karrenbelt, line 944: here and below I have changed and payload that was a string into a DummyPayload

Adamantios (changes requested):
LGTM except for the fact that we should also account for empty dicts returned by `data`.

dvilelaf (commented):
Looks good. I agree with Adamantios comment.

packages/valory/skills/abstract_round_abci/base.py
Karrenbelt: Yes there are payloads that may not have any data at some point. I don't see how that warrants returning an empty dict from `BaseTxPayload`, if such is the case the developer can implement it themselves. Would it not be better to make the `data` an abstractmethod to force the developer to implement it on concrete classes, instead of allowing for the possibility to forget to implement it, hence return an empty dict and potentially run into unexpected behaviour much further downstream? 

Also I don't quite follow the final remark. If we use `.keys()`  then the values could still be different and we would equate such payloads equal?

Adamantios: Some payloads return empty dicts on purpose. I agree with your suggestion on enforcing the implementation of `data`, but this is a separate issue that does not concern this PR. However, for this PR to be complete, I believe that we should also account for empty dicts in this method.

Re the `.keys()` I have misunderstood.

DavidMinarsch (approved):
LGTM

packages/valory/skills/abstract_round_abci/base.py
Karrenbelt: ok you asked in PM and was settled bu works the same for empty dict

PR #1064, opened 2022-07-12
Fix tendermint process termination on `TendermintNode` class

### LINKED ISSUE #1037: Add proper exception handling when starting and killing Tendermint process.
```output
    self = <deployments.Dockerfiles.localnode.tendermint.TendermintNode object at 0x12cd607f0>
    def _stop_tm_process(self) -> None:
        """Stop a Tendermint node process."""
        if self._process is None:
            return
        if platform.system() == "Windows":  # pragma: nocover
            os.kill(self._process.pid, signal.CTRL_C_EVENT)  # type: ignore  # pylint: disable=no-member
        else:
>           os.killpg(os.getpgid(self._process.pid), signal.SIGTERM)
E           ProcessLookupError: [Errno 3] No such process
deployments/Dockerfiles/localnode/tendermint.py:195: ProcessLookupError
```

From this CI run: https://github.com/valory-xyz/open-autonomy/runs/7110583220?check_suite_focus=true

angrybayblade:
## Proposed changes

This PR fixes tendermint process termination on `TendermintNode` class.

## Fixes

#1037 

## Types of changes

## Checklist


deployments/Dockerfiles/localnode/tendermint.py
```diff
@@ -189,13 +189,22 @@ def _stop_tm_process(self) -> None:
...
+            self._process.send_signal(signal.CTRL_C_EVENT)  # type: ignore  # pylint: disable=no-member
         else:
-            os.killpg(os.getpgid(self._process.pid), signal.SIGTERM)
+            self._process.send_signal(signal.SIGTERM)
+
+        self._process.wait(timeout=30)
+        poll = self._process.poll()
+        if poll is None:  # pragma: nocover
+            self._process.terminate()
+            self._process.wait(3)
```
Adamantios, line 201: Does this mean that we might wait for up to 33 secs? If yes, then this could be an issue if the reset and pause timeout is <= 33 secs. It might be worth mentioning it in the docs or guarding against it.

Also, why 30 and 3? 
Could we create constants for these values?

angrybayblade: So I addressed this in https://github.com/valory-xyz/open-autonomy/pull/1064/commits/0bf09bd28834b2e5636e1a51da1efe0a45692bc3 , I set the wait for exit timeout to 5 seconds. 30 or 5 and 3 are the values we are also using in some other places, there's no particular reason for them.

Adamantios (approved):
LGTM, failing tests will be resolved after merging #1077.

.github/workflows/main_workflow.yml
```diff
@@ -196,6 +196,12 @@ jobs:
           docker pull node:16.7.0
           docker pull trufflesuite/ganache:beta
 
+          # build required images
+          pip install .
+          autonomy deploy build image valory/oracle_hardhat
+          autonomy deploy build image valory/oracle_hardhat --dev
+          autonomy deploy build image valory/oracle_hardhat --dependencies
```
angrybayblade, line 203: This fixes https://github.com/valory-xyz/open-autonomy/issues/1067

DavidMinarsch (approved):

PR #1068, opened 2022-07-13
Fix `verify_contract` method

Adamantios:
## Proposed changes

The `verify_contract` method was returning a hardcoded `True` value, preventing us from failing early in the corresponding unit test. Therefore, we were detecting issues with contract verification later, in the e2e tests.

## Fixes

n/a

## Types of changes

## Checklist

## Further comments

n/a


Adamantios:
`test_verify_contract` will now fail, because of this update. It will pass again when we sync with the service registry updates.

tests/test_contracts/test_service_registry/test_contract.py
```diff
@@ -137,16 +139,22 @@ def test_deployment(self) -> None:
...
             ledger_api=self.ledger_api,
-            contract_address=self.contract_address,
+            contract_address=contract_address,
         )
 
-        assert result["verified"] is True, result
+        if valid_address:
+            assert result["verified"] is True, result
+        else:
+            assert result["verified"] is False
```
Karrenbelt, line 157: `assert result["verified"] is valid_address, result`

DavidMinarsch (approved):

Issue #1080, opened 2022-07-15
Registration startup entered twice


Adamantios:
### Subject of the issue
The registration startup round is entered twice as observed in the logs:
```
[2022-07-14 11:00:47,658] [WARNING] [agent] The kwargs={'log_dir': '/logs'} passed to benchmark_tool have not been set!
[2022-07-14 11:00:47,779] [WARNING] [agent] The kwargs={'log_dir': '/benchmarks'} passed to benchmark_tool have not been set!
[2022-07-14 11:00:47,955] [INFO] [agent] Entered in the 'registration_startup' round for period 0
[2022-07-14 11:00:47,956] [INFO] [agent] Entered in the 'registration_startup' round for period 0
[2022-07-14 11:00:47,957] [INFO] [agent] Start processing messages...
[2022-07-14 11:00:47,958] [INFO] [agent] Entered in the 'registration_startup' behaviour
[2022-07-14 11:00:47,958] [INFO] [agent] Checking sync...
[2022-07-14 11:00:47,958] [INFO] [agent] Checking status @ http://localhost:26657/status
[2022-07-14 11:00:47,964] [INFO] [agent] Entered in the 'registration_startup' behaviour
```

This is probably related to the repeated transaction errors in our logs:
```
[2022-07-14 11:01:17,155] [ERROR] [agent] TransactionNotValidError: sender 0xAD56A50628D4915E841b7C2eCE13b33793b365d0 has already sent value for round: registration_startup
[2022-07-14 11:01:17,158] [ERROR] [agent] TransactionNotValidError: sender 0xAD56A50628D4915E841b7C2eCE13b33793b365d0 has already sent value for round: registration_startup
[2022-07-14 11:01:17,160] [ERROR] [agent] TransactionNotValidError: sender 0xAD56A50628D4915E841b7C2eCE13b33793b365d0 has already sent value for round: registration_startup
[2022-07-14 11:01:17,165] [ERROR] [agent] TransactionNotValidError: sender 0xAD56A50628D4915E841b7C2eCE13b33793b365d0 has already sent value for round: registration_startup
```

This is not a terminal error but is certainly worth addressing.

### Steps to reproduce
Run any e2e test.

### Expected behaviour
The registration round should be entered once.

### Actual behaviour
The registration round is entered twice.

marcofavorito:
[...]

PR #1086, opened 2022-07-17
Add check that FSM chaining only involve abstract skills (fix #1085)

### LINKED ISSUE #1085: Check FSM chaining only involves abstract skills
**Is your feature request related to a problem? Please describe.**
As shown in https://github.com/valory-xyz/open-autonomy/issues/1080#issuecomment-1186296993, there is no check whether a chaining of ABCI skills will contain some non-abstract skills. This can cause problems as the non-abstract skill of the ABCI component being chained might interfere with the chaining skill, e.g. by scheduling a concurrent behaviour that makes the main behaviour failing.

**Describe the solution you'd like**
Implement a check in the chaining which assures non of the FSMs which are being chained are concrete.

**Describe alternatives you've considered**
n/a

**Additional context**
n/a

marcofavorito:
## Proposed changes

This PR implements a check in the `packages/valory/skills/abstract_round_abci/abci_app_chain.py::chain` function that requires operands to be _abstract_ AbciApp. This is done by adding a private class attribute `_is_abstract: bool`, which defaults to `True`, but it is set from the `SharedState` of the ABCI skill that implements it.

This is useful to prevent a bad omission caused by `registration_abci` skill not being abstract, as shown in #1080 .

## Fixes

Fix #1085 

## Types of changes

## Checklist

- [X] I have read the [CONTRIBUTING](../blob/main/CONTRIBUTING.md) doc
- [X] I am making a pull request against the `main` branch (left side). Also you should start your branch off our `main`.
- [X] Lint and unit tests pass locally with my changes
- [X] I have added tests that prove my fix is effective or that my feature works
- [X] I have locally run services that could be impacted and they do not present failures derived from my changes

## Further comments

n.a


DavidMinarsch (approved):

tests/test_skills/base.py
```diff
@@ -376,3 +377,14 @@ def teardown(cls) -> None:
...
+
+class DummyContext:
+    """Dummy Context class for testing shared state initialization."""
+
+    class params:
+        """Dummy param variable."""
+
+        round_timeout_seconds: float = 1.0
+
+    _skill: MagicMock = MagicMock()
```
Adamantios, line 390: :+1:

packages/valory/skills/abstract_round_abci/models.py
```diff
@@ -101,11 +101,20 @@ def _ensure(cls, key: str, kwargs: Dict) -> Any:
...
+        *args: Any,
+        abci_app_cls: Type[AbciApp],
+        skill_context: SkillContext,
+        **kwargs: Any,
+    ) -> None:
         """Initialize the state."""
         self.abci_app_cls = self._process_abci_app_cls(abci_app_cls)
+        self.abci_app_cls._is_abstract = (
+            skill_context._skill.configuration.is_abstract_component  # pylint: disable=protected-access
+        )
```
Adamantios, line 115: @angrybayblade I'd suggest that we create a property for `is_abstract_component` at `open-aea`'s `SkillContext`.

Adamantios (approved):
LGTM, only some minor comments.

Also, let's first see if e2e tests still timeout after merging `test/registration` on this branch.

packages/valory/skills/abstract_round_abci/models.py
marcofavorito: the problem is that `is_abstract` should not be modified arbitrarily, but only propertly by `SharedState`, hence it was a protected class attributes.

We could have a public access method though.

Adamantios: #1093.

PR #1082, opened 2022-07-15
Fix/monitoring flag

dvilelaf:
## Proposed changes

Set the monitoring flag to true as default in the abci connection

## Fixes
n/a

## Types of changes

What types of changes does your code introduce? (A **breaking change** is a fix or feature that would cause existing functionality and APIs to not work as expected.)
_Put an `x` in the box that applies_

- [x] Non-breaking fix (non-breaking change which fixes an issue)
- [ ] Breaking fix (breaking change which fixes an issue)
- [ ] Non-breaking feature (non-breaking change which adds functionality)
- [ ] Breaking feature (breaking change which adds functionality)
- [ ] Refactor (non-breaking change which changes implementation)
- [ ] Messy (mixture of the above - requires explanation!)

## Checklist

_Put an `x` in the boxes that apply._

- [x] I have read the [CONTRIBUTING](../blob/main/CONTRIBUTING.md) doc
- [x] I am making a pull request against the `main` branch (left side). Also you should start your branch off our `main`.
- [x] Lint and unit tests pass locally with my changes
- [ ] I have added tests that prove my fix is effective or that my feature works
- [ ] I have locally run services that could be impacted and they do not present failures derived from my changes

PR #1087, opened 2022-07-17
Replace Popen.wait with Popen.communicate in `abci` connection (fix #1073)

### LINKED ISSUE #1073: Tendermint stops after a while when start_monitoring=False
### Subject of the issue
During ElCol QA we found that the default value (False) for `start_monitoring` in `abci/connection.py` makes the e2e tests fail. It looks like some buffer is filled and blocks it making Tendermint stop producing blocks. 

From Ardian:
`(Coming from google?) all PIPEs will deadlock when one of the PIPEs' buffer gets filled up and not read. e.g. stdout deadlock when stderr is filled. Never pass a PIPE you don't intend read`

We should write a test and then fix this.

### Your environment
- OS: Ubuntu 20.04
- Python version: 3.10
- Package Version: 0.1.2

### Steps to reproduce
Run a long e2e test with start_monitoring=False

### Expected behaviour
e2e test should run

### Actual behaviour
e2e fails

marcofavorito:
## Proposed changes

This PR replaces `Popen.wait` with `Popen.communicate` in `abci` connection (fix #1073)

This fix is about a known problem with `subprocesses.Popen` when `stdout=PIPE` or `stderr=PIPE` is used:
https://docs.python.org/3/library/subprocess.html#subprocess.Popen.wait

> Note: This will deadlock when using `stdout=PIPE` or `stderr=PIPE` and the child process generates enough output to a pipe such that it blocks waiting for the OS pipe buffer to accept more data. Use [`Popen.communicate()`](https://docs.python.org/3/library/subprocess.html#subprocess.Popen.communicate) when using pipes to avoid that.

And indeed ABCI connection uses `Popen.wait` to wait for the subprocess termination.

On the other hand, the Python documentation of `subprocess` gives another warning related to the use of `Popen.communicate()` (https://docs.python.org/3/library/subprocess.html#subprocess.Popen.communicate):

> Note: The data read is buffered in memory, so do not use this method if the data size is large or unlimited.

We should be OK for our use case; still, good to keep it in mind.

## Fixes

Fix #1073

## Types of changes

## Checklist

- [X] I have read the [CONTRIBUTING](../blob/main/CONTRIBUTING.md) doc
- [X] I am making a pull request against the `main` branch (left side). Also you should start your branch off our `main`.
- [X] Lint and unit tests pass locally with my changes
- [X] I have added tests that prove my fix is effective or that my feature works
- [X] I have locally run services that could be impacted and they do not present failures derived from my changes

## Further comments

n/a


packages/valory/connections/abci/connection.py
```diff
@@ -489,7 +489,7 @@ def init(self) -> None:
         cmd = self._build_init_command()
         subprocess.call(cmd)  # nosec
 
-    def start(self, start_monitoring: bool = True) -> None:
```
dvilelaf, line 492: I think we need to revert deployments/Dockerfiles/localnode/tendermint.py as well.

marcofavorito: sorry I did not mean to undo [e7d6981](https://github.com/valory-xyz/open-autonomy/pull/1082/commits/e7d6981c8ff115470696cf8f63e9ebdd1eeb708e); restoring `True`

dvilelaf (commented):
Looks good, but it would be nice to have a test that shows that it solves 1073.

marcofavorito:
> Looks good, but it would be nice to have a test that shows that it solves 1073.

Tests got still stuck... there is some other problem. We could try to temporarily remove `PIPE` so to be sure that indeed that is the problem.

dvilelaf:
> > Looks good, but it would be nice to have a test that shows that it solves 1073.
> 
> Tests got still stuck... there is some other problem. We could try to temporarily remove `PIPE` so to be sure that indeed that is the problem.

Testing it here: https://github.com/valory-xyz/open-autonomy/pull/1092

dvilelaf:
#1073 is being fixed here: https://github.com/valory-xyz/open-autonomy/pull/1112

dvilelaf:
Not sure why/how i closed this but this could be a different issue, could you reopen (I can't)?

PR #1097, opened 2022-07-19
feat: check unused events

jmoreira-valory:
## Proposed changes

Added a check to verify that all events are used in the AbciApp.
It checks that for each transition of the form
(roundX, eventY) -> roundZ

the source code of roundX (or its superclasses) reference "eventY" in the sourcecode.

Note that this check is independent from the spec file (YAML). However, it is launched when verifying the AbciApp against its specification.

Note: the regex to identify events can be polished in a future PR to avoid catching events in comments or docstrings.

## Fixes

## Types of changes

## Checklist

## Further comments


autonomy/analyse/abci/app_spec.py
```diff
@@ -338,6 +343,40 @@ def abci_to_dfa(cls, abci_app_cls: Any, label: str = "") -> "DFA":
...
 
 
+event_pattern = re.compile(r"Event\.(\w+)", re.DOTALL)
+
+
+def _check_unreferenced_events(abci_app_cls: AbciApp) -> None:
+    """Checks that events defined in the AbciApp transition function are referenced in the source code of the coresponding round or its superclasses."""
+
+    error_strings = []
+    timeout_events = set([k.name for k in abci_app_cls.event_to_timeout.keys()])
```
Karrenbelt, line 353: apart from list comprehension, you can use generator, set and dictionary comprehensions in python. It's more effective to use them directly, because now you generate the list first, then the set, which is wasteful.

this would make the linter happy
`timeout_events = {k.name for k in abci_app_cls.event_to_timeout}`

jmoreira-valory: Fixed.

autonomy/analyse/abci/app_spec.py
```diff
@@ -338,6 +343,40 @@ def abci_to_dfa(cls, abci_app_cls: Any, label: str = "") -> "DFA":
...
+        if trf_events.symmetric_difference(referenced_events):
+            error_strings.append(
+                f" - {round_cls.__name__}: transition function events {trf_events} do not match referenced events {referenced_events}."
+            )
+
+    if len(error_strings) > 0:
+        raise DFASpecificationError(
+            f"{abci_app_cls} ABCI App has the following issues:\n"
+            + "\n".join(error_strings)
+        )
```
Karrenbelt, line 377: I like this, we return all error in the ABCIApp, not Error out on the first, which would make this a game of iterative and incremental fixes

Karrenbelt (commented):
1. ~~why must we walk the MRO? We don't, since all events are defined locally. Yes, we might return `self.done_event` from a parent class, but that requires us to assign it as (e.g.) a class variable here and hence you will have parsed it.~~  
2. Since we parse the class, we cannot differentiate between the an assignment to a class attribute, a local variable, or a direct return of Event.x. This is fine as I understand that using AST offers no easy solution either, however I do think we should inform the other developer of as much in the check here; it requires documentation.

autonomy/analyse/abci/app_spec.py
jmoreira-valory: I am glad you like it, as you suggested this approach in an earlier PR :)

jmoreira-valory:
> ~~why must we walk the MRO? We don't, since all events are defined locally. Yes, we might return self.done_event from a parent class, but that requires us to assign it as (e.g.) a class variable here and hence you will have parsed it. ~~

Because there are instances where the superclass is explicitly called, and we have no way to control if "other" events are launched. See for example `packages.valory.skills.transaction_settlement_abci.rounds.SynchronizeLateMessagesRound`


> Since we parse the class, we cannot differentiate between the an assignment to a class attribute, a local variable, or a direct return of Event.x. This is fine as I understand that using AST offers no easy solution either, however I do think we should inform the other developer of as much in the check here; it requires documentation.

Will update the docstring.

autonomy/analyse/abci/app_spec.py
```diff
@@ -338,6 +343,51 @@ def abci_to_dfa(cls, abci_app_cls: Any, label: str = "") -> "DFA":
...
+            str(e).rsplit(".", 1)[1] for e in round_transitions
+        } - timeout_events
+        referenced_events = set()
+
+        for base in filter(
+            lambda x: x.__class__.__module__ != "builtins",
+            inspect.getmro(round_cls),
+        ):
+            src = textwrap.dedent(inspect.getsource(base))
+            referenced_events.update(event_pattern.findall(src))
```
Adamantios, line 377: I disagree with this. Just the fact that the `src` contains an `Event.SOME_EVENT` does not mean that we also return this event from a round. For example, take a look at the APY's `APYEstimationAbstractRound` which is inherited by all the rounds:

https://github.com/valory-xyz/open-autonomy/blob/9ffbe860bc521842c882b33dd9bbebcac8747422/packages/valory/skills/apy_estimation_abci/rounds.py#L142-L164

Your script will require `NO_MAJORITY` and `FILE_ERROR` for all the rounds.

autonomy/analyse/abci/app_spec.py
```diff
@@ -338,6 +343,51 @@ def abci_to_dfa(cls, abci_app_cls: Any, label: str = "") -> "DFA":
...
+    :raises DFASpecificationError: If there are unfererenced events in the AbciApp.
+    """
+
+    error_strings = []
+    timeout_events = {k.name for k in abci_app_cls.event_to_timeout.keys()}
+
+    for round_cls, round_transitions in abci_app_cls.transition_function.items():
+        trf_events = {
+            str(e).rsplit(".", 1)[1] for e in round_transitions
+        } - timeout_events
```
Adamantios, line 369: Why do we remove timeout events?

autonomy/analyse/abci/app_spec.py
```diff
@@ -338,6 +343,51 @@ def abci_to_dfa(cls, abci_app_cls: Any, label: str = "") -> "DFA":
         )
 
 
+event_pattern = re.compile(r"Event\.(\w+)", re.DOTALL)
+
+
+def _check_unreferenced_events(abci_app_cls: AbciApp) -> None:
```
Adamantios, line 349: Why keep this out of the method?

autonomy/analyse/abci/app_spec.py
Adamantios: A simple solution would be to remove these two helper methods, but the script would not be 100% correct.

DavidMinarsch: The problem is we can rely on static analysis only. We could make it a convention to adopt the policies checked by the script. It would keep the code easier to audit as well (don't have any code which isn't used in all inherited classes in the base class).

autonomy/analyse/abci/app_spec.py
DavidMinarsch: I assume because it is trivially present in all cases?

autonomy/analyse/abci/app_spec.py
Adamantios: Refactored APY in #1119, to remove code in the base class which isn't used in all inherited classes.

autonomy/analyse/abci/app_spec.py
jmoreira-valory: Because they are not returned explicitly in the source code of the rounds.

autonomy/analyse/abci/app_spec.py
jmoreira-valory: To avoid compiling the regular expression (as the method is called multiple times).

autonomy/analyse/abci/app_spec.py
```diff
@@ -18,18 +18,23 @@
...
 import sys
+import textwrap
 from collections import defaultdict, deque
 from itertools import product
 from pathlib import Path
 from typing import Any, Dict, List, OrderedDict, Set, TextIO, Tuple
 
 import yaml
 
+from packages.valory.skills.abstract_round_abci.base import AbciApp
```
DavidMinarsch, line 36: This is an absolute no-go ...this doesn't necessarily exist from perspective of the autonomy

DavidMinarsch: It's (somewhat hackily) solved in `check_one` method

autonomy/analyse/abci/app_spec.py
Adamantios: Ok, I updated it in ee862018b729dae611a4842f16091339d97df304.

DavidMinarsch (approved):
Pending tests passing - lgtm

PR #1104, opened 2022-07-21
Fix/docs hash check script

dvilelaf:
## Proposed changes

Generalize the hash checker script to detect more commands

## Fixes

n/a

## Types of changes

## Checklist


scripts/check_doc_ipfs_hashes.py
```diff
@@ -30,7 +30,7 @@
 
 
 IPFS_HASH_REGEX = r"bafybei[A-Za-z0-9]{52}"
-AEA_COMMAND_REGEX = r"(?P<full_cmd>(?P<cli>aea|autonomy) (?P<cmd>fetch|add .*) (?:(?P<vendor>.*)\/(?P<package>.[^:]*):(?P<version>\d+\.\d+\.\d+)?:?)?(?P<hash>bafybei[A-Za-z0-9]{52}))"
+AEA_COMMAND_REGEX = fr"(?P<full_cmd>(?P<cli>aea|autonomy) (?P<cmd>.*) (?:(?P<vendor>.*)\/(?P<package>.[^:]*):(?P<version>\d+\.\d+\.\d+)?:?)?(?P<hash>{IPFS_HASH_REGEX}))"
```
dvilelaf, line 33: Using a f-regex-string now to avoid regex duplication

scripts/check_doc_ipfs_hashes.py
```diff
@@ -91,13 +91,12 @@ def __init__(self, package_line: str) -> None:
...
-            return (
-                f"aea fetch {self.vendor}/{self.name}:{self.last_version}:{self.hash}"
-            )
-        return f"aea add {self.type} {self.vendor}/{self.name}:{self.last_version}:{self.hash}"
+    def get_command(self, cmd, include_version=True) -> str:
+        """Get the corresponding command"""
+        version = (
+            ":" + self.last_version if include_version and self.last_version else ""
+        )
+        return f"autonomy {cmd} {self.vendor}/{self.name}{version}:{self.hash}"
```
dvilelaf, line 99: Now passing the command so we can rebuild the complete string for any command

scripts/check_doc_ipfs_hashes.py
```diff
@@ -159,7 +159,18 @@ def get_hash_by_package_line(
...
+                package_type = "service"
+            if d["cmd"] == "fetch":
+                package_type = "agent"
+            if d["cmd"].startswith("add"):
+                package_type = d["cmd"].split(" ")[-1]  # i.e.: aea add connection
+            if not package_type:
+                raise ValueError(
+                    f"Docs [{md_file}]: could not infer the package type for line '{package_line}'"
+                )
+
```
dvilelaf, line 173: Better detect the package type

scripts/check_doc_ipfs_hashes.py
angrybayblade: You can use the public id regex here

dvilelaf: Ok, I've incorporated some already existing regexes but since i needed capturing groups i still had to add some custom ones.

angrybayblade (approved):
The failures are mostly network failures on macOS, but this one is a bit different: https://github.com/valory-xyz/open-autonomy/runs/7451315965?check_suite_focus=true

DavidMinarsch (approved):

PR #1112, opened 2022-07-22
Fix/tendermint monitoring

### LINKED ISSUE #1073: Tendermint stops after a while when start_monitoring=False

dvilelaf:
## Proposed changes

Adds tests and a fix to the tendermint monitoring issue that makes the node timeout. 
Props to Ardian for all the help with the tests.

## Fixes

#1073 

## Types of changes

## Checklist


deployments/Dockerfiles/localnode/app.py
```diff
@@ -122,7 +122,7 @@ def dump_period(self) -> None:
 
 
 def create_app(
-    dump_dir: Optional[Path] = None, perform_monitoring: bool = True
+    dump_dir: Optional[Path] = None, perform_monitoring: bool = True, debug: bool = False
```
dvilelaf, line 125: We're now passing this debug flag so we can set it from the test and generate a lot more info to make the test fail under certain conditions.

deployments/Dockerfiles/localnode/tendermint.py
```diff
@@ -143,40 +146,39 @@ def init(self) -> None:
...
+        cmd = self._build_node_command(debug)
+        kwargs = {
+            "bufsize": 1,
+            "universal_newlines": True,
+        }
+
+        # Only redirect stdout and stderr if we're going to read
+        if monitoring:
+            kwargs["stdout"] = subprocess.PIPE
+            kwargs["stderr"] = subprocess.STDOUT
```
dvilelaf, line 168: Here's the actual fix

tests/test_deployments/test_app.py
```diff
@@ -269,3 +275,56 @@ def get_missing(messages: List[str]) -> List[str]:
...
+        try:
+            # Give the test 30 seconds for it to throw a timeout
+            for _ in range(30):
+                # If it hangs for 5 seconds, we assume it's not working.
+                # Increasing the timeout should have no effect,
+                # the node is unresponsive at this point.
+                requests.get("http://localhost:26657/status", timeout=5)
+                time.sleep(1)
+        except requests.exceptions.Timeout:
+            assert False, "Tendermint has timed out"
```
dvilelaf, line 299: This will fail without the fix and passes with it

Karrenbelt (commented):
[These tests](https://github.com/valory-xyz/open-autonomy/blob/089e0d7da889c4baabb69ce7e6ac352c18249b8b/tests/test_deployments/test_deployment_code_identical.py) will fail

tests/test_deployments/test_app.py
```diff
@@ -269,3 +275,56 @@ def get_missing(messages: List[str]) -> List[str]:
...
+        except requests.exceptions.Timeout:
+            raise AssertionError("Tendermint has timed out")
+
+    @classmethod
+    def teardown_class(cls) -> None:
+        """Teardown the test."""
+        cls.app_context.pop()
+        # Stopping doesn't work when monitoring is off
+        if cls.perform_monitoring:
+            cls.tendermint_node.stop()
```
Adamantios, line 307: Could you explain this comment?

dvilelaf: That comes from Ardian. I think he means that when this test fails due to the node timing out, we cannot call stop anymore, so we only call stop when we are monitoring.

dvilelaf:
Some failing test with the message:` bind: address already in use"`
This might indicate that we are not properly closing the connection.

tests/test_deployments/test_app.py
dvilelaf: I had to remove this as tendermint was not being stopped properly and made other tests fail.

DavidMinarsch (changes requested):
Tests currently time out

dvilelaf:
Unrelated failing e2e: registration and APY.

tests/test_deployments/test_app.py
```diff
@@ -269,3 +275,54 @@ def get_missing(messages: List[str]) -> List[str]:
...
+        try:
+            # Give the test 30 seconds for it to throw a timeout
+            for _ in range(30):
+                # If it hangs for 5 seconds, we assume it's not working.
+                # Increasing the timeout should have no effect,
+                # the node is unresponsive at this point.
+                requests.get("http://localhost:26657/status", timeout=5)
+                time.sleep(1)
+        except requests.exceptions.Timeout:
+            raise AssertionError("Tendermint has timed out")
```
Karrenbelt, line 299: can't we just fill up the buffer in one step? What is the size limit? lets use that in test if we can?

pseudo code:
```
maxsize = (1 << 31) - 1
test_tendermint_buffer(maxsize) 
# next test or flush buffer here
with pytest.raises(requests.exceptions.Timeout)
    test_tendermint_buffer(maxsize + 1)
```

tests/test_deployments/test_app.py
```diff
@@ -269,3 +275,54 @@ def get_missing(messages: List[str]) -> List[str]:
...
+
+    @classmethod
+    def teardown_class(cls) -> None:
+        """Teardown the test."""
+        cls.app_context.pop()
+        cls.tendermint_node.stop()
+        shutil.rmtree(cls.tm_home, ignore_errors=True, onerror=readonly_handler)
```
Karrenbelt, line 306: I think `ignore_errors` overwrites `onerror`. 

Not available yet until next release on open-aea side, but have a look at [`remove_test_directory`](https://github.com/valory-xyz/open-aea/blob/6395b608560d6e7f2d1ed043087dd4c375cb3b39/tests/conftest.py#L363-L385) and let me know if you think we should open a ticket to (attempt) replacing `shutil.rmtree` with this - any suggestions on the implementation  more than welcome too!

tests/test_deployments/test_app.py
```diff
@@ -269,3 +275,54 @@ def get_missing(messages: List[str]) -> List[str]:
...
+    def test_tendermint_buffer(self) -> None:
+        """Test Tendermint buffer"""
+
+        try:
+            # Give the test 30 seconds for it to throw a timeout
+            for _ in range(30):
+                # If it hangs for 5 seconds, we assume it's not working.
+                # Increasing the timeout should have no effect,
+                # the node is unresponsive at this point.
+                requests.get("http://localhost:26657/status", timeout=5)
```
Karrenbelt, line 296: let's define this URI, `"http://localhost:26657/status"`, at the class level and reference it here and where ever else we use it

tests/test_deployments/test_app.py
```diff
@@ -269,3 +275,54 @@ def get_missing(messages: List[str]) -> List[str]:
...
+        # Give the test 60 seconds for it to work
+        for _ in range(60):
+            try:
+                res = requests.get("http://localhost:26657/status", timeout=5)
+                # We expect all responses to be OK
+                assert res.status_code == 200
+            except Exception as e:
+                raise AssertionError(e)
+
+            time.sleep(1)
```
Karrenbelt, line 328: so I'm gonna comment it here but holds for the test above as well. We try 60 times, if not timeout for each attempt (might take upto 5 sec) then we're good. So can take upto 300 seconds. Is that what we want? then clarify in comments. 

However, I don't think the test is clear, I will explain in the review comment up next.

Karrenbelt (commented):
I don't understand the tests, specifically the one asserting failure. The clearest way would to once fill the buffer to the rim, assert this works, and once add one byte more, assert this raises an exception.

We need:
- one test passing (filling till rim)
- one test failing. This should use `with pytest.raises(match: "<buffer timeout of sorts>")`

To clarify my miscomprehension: How is the test for failure asserting failure?

tests/test_deployments/test_app.py
0xArdi: To clarify what we have here, in my understanding when running the agent with `use_tendermint: True` we get the following log flow between tendermint and the agent: 
tendermint -> PIPE -> aea.

The [pipe has limited capacity](https://man7.org/linux/man-pages/man7/pipe.7.html#:~:text=A%20pipe%20has%20a%20limited,process%20does%20not%20remain%20blocked.), typically 16KB. When this limit is reached the process that writes to it *might* be blocked. 
It depends on the writing application (in our case tendermint), OS and the config of the `subprocess.popen()`call whether & when that might happen.
In our case we see this problem, at least on Debian distros, I haven't tried with windows. This blocking happens when the reader doesn't read and the pipe limit is reached. 

What we are trying to achieve with these tests is to show that when the pipe limit is reached, and the reader (AEA) isn't reading tendermint blocks.
So, what we do is we run the config as it is, with the small difference that we configure tendermint to use debug logging, because it will cause it to spit out more logs, which would fill the pipe faster, and ultimately cause the tendermint process to block faster. On my machine the failure happens in about ~6s. 

When reading is turned off (`perform_monitoring = False`), we give 30 seconds for tendermint to reach the pipe limit and block, we check whether tendermint has blocked if it doesn't return a 200 on the status check, we give 5s time for it to respond, which should be more than enough (it normally takes milliseconds).
The test shows this, tendermint takes more than 5s to respond, and we assume its blocked. Increasing the timeout should have no effect, although it would make the tests take longer, you can try out with a bigger timeout.

When reading is turned on (`perform_monitoring = True`), the pipe shouldn't get filled and tendermint shouldn't be blocked. To prove that this is the case, we give tendermint twice the time to reach the pipe limit and block. We can switch to 30s, that should have no difference, I chose for 60s just make sure.

Karrenbelt (commented):
what i don't understand, and would like answered, is how this assesses failure. As far as I can tell it doesn't

https://github.com/valory-xyz/open-autonomy/blob/6a50a20cf457e6fea8bf4df7d5b48bc003590478/tests/test_deployments/test_app.py#L281-L307

dvilelaf:
> what i don't understand, and would like answered, is how this assesses failure. As far as I can tell it doesn't
> 
> https://github.com/valory-xyz/open-autonomy/blob/6a50a20cf457e6fea8bf4df7d5b48bc003590478/tests/test_deployments/test_app.py#L281-L307

Have you seen Ardian's reply? Initially I experimented with the `pytest.raises()` and worked nice: the test failed and the raises() statement got it right. The problem came after we introduced the bug fix (conditional PIPE on line 166): once fixed, we could not reproduce it anymore so the pytest.raises didn't work and I had to roll it back and keep the test as a simple one that will fail if the buffer is filled (we could have something like a testing parameter that is passed but that felt hacky), but that won't happen with the fix (or at least we haven't seen it fail). 

TLDR: I don't see how, after fixing the bug, I can force the error to happen without being hacky. Ideas welcome!

dvilelaf:
Only `tests/test_agents/test_register_reset_abci.py::TestTendermintStartup::test_run[4]` failing ATM.

Karrenbelt:
> 

Yes I read his elaborate response - doesn't hurt to learn a bit more about the details. However, my comment related to the test whose name is no longer appropriate, as it does not test the failure. It now tests that there is no longer a failure. 

I suppose you could set `self.tendermint_node._process` anew with the faulty parameters? `monitoring = False` to the setup and this `._process = subprocessPopen(stdout=subprocess.PIPE, stderr=subprocess.STDOUT`?

Karrenbelt:
I suggest a refactor since single responsibility principle is violated.

- remove `kwargs` local variable container outside of the scope of this method. In the current implementation they are constant as you cannot change them because of this, which in this particular case makes it difficult to patch them for testing purposes to assess failure when writing but not reading from buffer.
https://github.com/valory-xyz/open-autonomy/blob/bbb00868aa4aadd13905ae12203bb4383515aa53/deployments/Dockerfiles/localnode/tendermint.py#L155-L177

It doesn't need another class, just a dict is fine. E.g. pass them on `__init__` as default. You can 
```
def _get_popen_kwargs():
    return{
            "bufsize": 1,
            "universal_newlines": True,
            "stdout": None,
            "stderr": None,
        }
...
...
    self.popen_kwargs = _get_popen_kwargs()
```

- move the content of `_build_node_command` to `TendermintParams`, which is clearly where it should reside. While at it you may replace `TendermintParams.__str__` special method to return `" ".join(self._build_node_command())`
https://github.com/valory-xyz/open-autonomy/blob/bbb00868aa4aadd13905ae12203bb4383515aa53/deployments/Dockerfiles/localnode/tendermint.py#L85-L96

Now the entire command can be obtained using `self.params._build_node_command() + some_func(self.kwargs)`, you can pass `monitoring` parameter to `start` command to overwrite if you want, or not and patch it on the object to simulate the failure.

tests/test_deployments/test_app.py
```diff
@@ -271,3 +279,78 @@ def get_missing(messages: List[str]) -> List[str]:
...
+    """Test Tendermint buffer"""
+
+    perform_monitoring = False  # Setting this flag to False and not monitoring makes the buffer to fill and freeze Tendermint
+    debug_tendermint = True  # This will cause more logging -> faster failure
+
+    @wait_for_node_to_run
+    @mock.patch(
+        "autonomy.data.Dockerfiles.tendermint.tendermint.TendermintParams.get_node_command_kwargs",
+        mock_get_node_command_kwargs,
+    )
```
dvilelaf, line 313: This patch is not working. I've tried several approaches, but it seems that the node does not exists at this point so I can't use mock.patch.object. Any idea @Karrenbelt ?

Karrenbelt: can try something like 
```
    @classmethod
    def setup_class(cls) -> None:
        with mock.patch.object(TendermintParams, "get_node_command_kwargs", return_value=mock_get_node_command_kwargs()):
            super().setup_class()
```

autonomy/data/Dockerfiles/tendermint/tendermint.py
```diff
@@ -98,6 +98,45 @@ def __str__(self) -> str:
...
+        if debug:
+            cmd.append("--log_level=debug")
+
+        if self.home is not None:  # pragma: nocover
+            cmd += ["--home", self.home]
+        return cmd
+
+    @staticmethod
+    def get_node_command_kwargs(monitoring: bool = False) -> Dict:
+        """Get the node command kwargs"""
```
dvilelaf, line 122: Moved these two methods here as proposed by @Karrenbelt to facilitate the mocking.

tests/test_deployments/test_app.py
```diff
@@ -271,3 +280,83 @@ def get_missing(messages: List[str]) -> List[str]:
...
+
+    @classmethod
+    def setup_class(cls) -> None:
+        """Setup the test."""
+        with mock.patch.object(
+            TendermintParams,
+            "get_node_command_kwargs",
+            return_value=mock_get_node_command_kwargs(),
+        ):
+            super().setup_class()
```
dvilelaf, line 318: We needed to mock on the `setup_class` as a normal mocking was not working. Thanks for the help here @Karrenbelt

dvilelaf:
@Karrenbelt I've applied the suggested changes with the exception of the small `TendermintParams.__str__` refactor as it was outside the scope of this PR.

Karrenbelt:
I think the buffer is still full which causes the timeout. Need to flush it or perhaps just kill the subprocess entirely in teardown.

bind the process to the instance here: `self.process = ...`
https://github.com/valory-xyz/open-autonomy/blob/c545639d15cb2c1b89bac1f33f1f7081582632f3/tests/test_deployments/test_app.py#L81

then in teardown try:
`os.killpg(os.getpgid(self.process.pid), signal.SIGTERM)`

I don't know if this works the same on windows

Karrenbelt (dismissed):
only two unrelated flaky tests remain
```
2022-08-03T17:24:46.5234072Z =========================== short test summary info ============================
2022-08-03T17:24:46.5234344Z FAILED tests/test_agents/test_registration.py::TestRegistrationStartUpFourAgents::test_run[4]
2022-08-03T17:24:46.5234558Z FAILED tests/test_agents/test_register_reset_abci.py::TestTendermintResetInterrupt::test_run[4]
```

tests/test_deployments/test_app.py
```diff
@@ -271,3 +280,89 @@ def get_missing(messages: List[str]) -> List[str]:
...
+            try:
+                # Give the test 30 seconds for it to throw a timeout
+                for _ in range(30):
+                    # If it hangs for 5 seconds, we assume it's not working.
+                    # Increasing the timeout should have no effect,
+                    # the node is unresponsive at this point.
+                    requests.get(self.tm_status_endpoint, timeout=5)
+                    time.sleep(1)
+            except requests.exceptions.Timeout:
+                raise AssertionError("Tendermint has timed out")
```
Adamantios, line 338: Why not:

```python
with pytest.raises(requests.exceptions.Timeout):
...
```
?

Adamantios (approved):

DavidMinarsch (approved):

PR #1113, opened 2022-07-25
Feat/goerli service specification

8ball030:
## Proposed changes

Adds in Goerli network.

## Fixes

## Types of changes

## Checklist

## Further comments


packages/valory/services/oracle_goerli/service.yaml
```diff
@@ -0,0 +1,98 @@
...
+  - server_api:
+      args: *id002
+  3:
+  - price_api:
+      args:
+        url: https://api.binance.com/api/v3/ticker/price
+        api_id: binance
+        parameters:
+        - - symbol
+          - BTCUSDT
```
Karrenbelt, line 88: why do we use USDT over USD? It's not the same and we use USD for the others

8ball030: image

8ball030: binance dont have that pair

Karrenbelt: https://bitcoin.stackexchange.com/questions/95445/why-does-binance-use-usdt-instead-of-usd

yes suboptimal. So what happens when it loses it's peg? It doesn't matter much for now though, but I dislike this because it's incorrect, rather would compare all to USDT or USDC then. Since this is only minor and not PR related, I will approve of the PR.

Karrenbelt (commented):
Will approve if you rerun tests and they pass

packages/valory/services/oracle_goerli/service.yaml
8ball030: we already use this pair for all of our oracles, 
I have created an issue for this.
https://github.com/valory-xyz/open-autonomy/issues/1114

scripts/check_doc_links.py
```diff
@@ -78,7 +78,7 @@ def main() -> None:  # pylint: disable=too-many-locals
                 error = True
 
             # Check for broken links: 200 and 403 codes are admitted
-            if url in url_skips:
+            if url in url_skips + http_skips:
```
8ball030, line 81: skips http and urls correctly

DavidMinarsch (approved):

PR #1119, opened 2022-07-26
Handle unused events

Adamantios:
## Proposed changes

Handle the unused events detected by the script added in #1097.

## Fixes

n/a

## Types of changes

## Checklist

## Further comments

n/a


packages/valory/skills/price_estimation_abci/rounds.py
```diff
@@ -193,6 +194,7 @@ class PriceAggregationAbciApp(AbciApp[Event]):
         },
         EstimateConsensusRound: {
             Event.DONE: TxHashRound,
+            Event.NONE: CollectObservationRound,
```
Adamantios, line 197: Transition was missing.

packages/valory/skills/simple_abci/rounds.py
```diff
@@ -237,6 +238,7 @@ class SimpleAbciApp(AbciApp[Event]):
     transition_function: AbciAppTransitionFunction = {
         RegistrationRound: {
             Event.DONE: RandomnessStartupRound,
+            Event.NO_MAJORITY: RegistrationRound,
```
Adamantios, line 241: Transition was missing.

packages/valory/skills/liquidity_rebalancing_abci/rounds.py
```diff
@@ -188,16 +187,14 @@ def end_block(self) -> Optional[Tuple[BaseSynchronizedData, Event]]:
...
-            elif strategy["action"] == StrategyType.SWAP_BACK.value:
-                event = Event.DONE_SWAP_BACK
-            return synchronized_data, event
+                return synchronized_data, Event.DONE
+            if strategy["action"] == StrategyType.ENTER.value:
+                return synchronized_data, Event.DONE_ENTER
+            if strategy["action"] == StrategyType.EXIT.value:
+                return synchronized_data, Event.DONE_EXIT
+            if strategy["action"] == StrategyType.SWAP_BACK.value:
+                return synchronized_data, Event.DONE_SWAP_BACK
```
Adamantios, line 197: Not sure why we explicitly returned a timeout event here.

packages/valory/skills/hello_world_abci/rounds.py
```diff
@@ -75,14 +75,6 @@ def synchronized_data(self) -> SynchronizedData:
...
         return cast(SynchronizedData, self._synchronized_data)
 
-    def _return_no_majority_event(self) -> Tuple[SynchronizedData, Event]:
-        """
-        Trigger the NO_MAJORITY event.
-
-        :return: a new synchronized data and a NO_MAJORITY event
-        """
-        return self.synchronized_data, Event.NO_MAJORITY
-
```
Adamantios, line 85: https://github.com/valory-xyz/open-autonomy/pull/1097#discussion_r929790977

packages/valory/skills/apy_estimation_abci/rounds.py
```diff
@@ -155,14 +155,6 @@ def _return_no_majority_event(self) -> Tuple[SynchronizedData, Event]:
...
         return self.synchronized_data, Event.NO_MAJORITY
 
-    def _return_file_error(self) -> Tuple[SynchronizedData, Event]:
-        """
-        Trigger the FILE_ERROR event.
-
-        :return: a new synchronized data and a FILE_ERROR event
-        """
-        return self.synchronized_data, Event.FILE_ERROR
-
```
Adamantios, line 165: https://github.com/valory-xyz/open-autonomy/pull/1097#discussion_r929790977

packages/valory/skills/liquidity_rebalancing_abci/rounds.py
Karrenbelt: not sure why we use `.value` here either, inconsistent with the rest

packages/valory/skills/oracle_abci/fsm_specification.yaml
```diff
@@ -75,6 +75,7 @@ transition_func:
     (DeploySafeRound, DONE): ValidateSafeRound
     (DeploySafeRound, FAILED): SelectKeeperSafeRound
     (EstimateConsensusRound, DONE): TxHashRound
+    (EstimateConsensusRound, NONE): CollectObservationRound
```
Karrenbelt, line 78: actual fix!

packages/valory/skills/apy_estimation_abci/behaviours.py
```diff
@@ -128,7 +128,7 @@ class APYEstimationBaseBehaviour(BaseBehaviour, ABC):
 
     @property
     def synchronized_data(self) -> SynchronizedData:
-        """Return the synchronized data."""
+        """Return the synchronized data data."""
```
DavidMinarsch, line 131: Typo?

Adamantios: Yes

packages/valory/skills/price_estimation_abci/fsm_specification.yaml
```diff
@@ -19,6 +19,7 @@ transition_func:
     (CollectObservationRound, NO_MAJORITY): CollectObservationRound
     (CollectObservationRound, ROUND_TIMEOUT): CollectObservationRound
     (EstimateConsensusRound, DONE): TxHashRound
+    (EstimateConsensusRound, NONE): CollectObservationRound
```
Karrenbelt, line 22: what is the difference between returning None and `Event.NONE`? Should we make it such that we always return `Event.NONE` in case we don't branch into another execution path first?

packages/valory/skills/liquidity_rebalancing_abci/rounds.py
Adamantios: I guess so that we do not pass the enum object through the payload. Instead, we pass the value and compare it with the other values.

A mapping from value to event would simplify the code here, but this is out of the scope of this PR.

packages/valory/skills/price_estimation_abci/fsm_specification.yaml
Adamantios: Returning `None` means that the execution of the round is still going on. We simply re-enter. Returning something other than `None` means that the round ended.

Karrenbelt (commented):
Looking good but test failures need resolving. Please [report flaky tests here](https://docs.google.com/document/d/1f7DrPdTXaLWJ1icTZ1Mt2S3z3ZdzCodrxAG6d1tI3o0/edit) if you find any

packages/valory/skills/apy_estimation_abci/rounds.py
```diff
@@ -177,7 +169,7 @@ def end_block(self) -> Optional[Tuple[BaseSynchronizedData, Event]]:
         """Process the end of the block."""
         if self.threshold_reached:
             if self.most_voted_payload is None:
-                return self._return_file_error()
+                return self.synchronized_data, Event.FILE_ERROR
```
jmoreira-valory, line 172: Why is it required to change this?

packages/valory/skills/hello_world_abci/rounds.py
```diff
@@ -122,7 +114,7 @@ def end_block(self) -> Optional[Tuple[BaseSynchronizedData, Event]]:
         if not self.is_majority_possible(
             self.collection, self.synchronized_data.nb_participants
         ):
-            return self._return_no_majority_event()
+            return self.synchronized_data, Event.NO_MAJORITY
```
jmoreira-valory, line 117: Same question as above, why is this required?

Karrenbelt: yep, we impose the additional constraint we discussed earlier as well @jmoreira-valory

jmoreira-valory: After discussing with @Karrenbelt and reviewing the comment from 1097:

Note that the script will examine the parent classes. In this particular case, it will catch this "NO_MAJORITY" event. So I assume that the refactoring here is **not** to fix a problem in this point of the code. Rather this refactoring is to avoid interpreting that other Rounds return the NO_MAJORITY event (because these rounds will not have that event on their parent class).

This observation applies also to the remaining similar refatorings.

DavidMinarsch (approved):

Karrenbelt (approved):
LGTM

PR #1120, opened 2022-07-27
test: replace check string

Adamantios:
## Proposed changes

The registration e2e tests sometimes fail. The issue is because of the way that we check if the test succeeded, and not because the test does not really succeed. We are checking the logs of the agent. However, if the execution reaches a degenerate round, then it starts looping and logging very fast the same exception. This results in the test not detecting the expected logs - probably because of a buffer filling - even though they have been logged, and falsely reporting the e2e test to have failed.

Therefore, we replace the check string with another one that happens a bit before reaching the degenerate round but still ensures that the test has succeeded.

However, this solution does not guarantee that the e2e test will not falsely report a failure again but decreases the chances.

Two more appropriate solutions would be to:
 1. Decrease the interval of checking the logs.
 2. Raise and stop the execution when a degenerate round is reached instead of looping.

## Fixes

n/a

## Types of changes

## Checklist

## Further comments

n/a

### DIFF
--- a/tests/test_agents/test_registration.py
+++ b/tests/test_agents/test_registration.py
@@ -54,7 +54,7 @@ STRICT_CHECK_STRINGS = (
     log_messages.collection_complete.value,
     log_messages.request_update.value,
     log_messages.response_update.value,
-    "Entered in the 'finished_registration' round for period 0",
+    "local height == remote height; continuing execution...",
 )
 
 


tests/test_agents/test_registration.py
```diff
@@ -54,7 +54,7 @@ def slow_down_tests() -> Generator:
     log_messages.collection_complete.value,
     log_messages.request_update.value,
     log_messages.response_update.value,
-    "Entered in the 'finished_registration' round for period 0",
+    "local height == remote height; continuing execution...",
```
Karrenbelt, line 57: So we report this when `reset_tendermint_with_wait` is called and executed successfully. 
Is this guaranteed to only occur right before degenerate round, not ever before, on successful execution? And is this reliably picked up on, in that it does not suffer from the same hiatus as the other message does?

https://github.com/valory-xyz/open-autonomy/blob/8a86d041a885fd0cd7dc460cfc1aa8165d09763f/packages/valory/skills/abstract_round_abci/behaviour_utils.py#L1705-L1709

Adamantios: This message occurs earlier than the previous one we had but still ensures that we succeeded in the e2e test. It just decreases the chances. We still enter the degenerate round but we may have picked up the log message before the buffer fills since it is logged earlier.

Karrenbelt (commented):
> Two more appropriate solutions would be to:
> 
>     1. Decrease the interval of checking the logs.
> 
>     2. Raise and stop the execution when a degenerate round is reached instead of looping.

in case of "stop the execution", we would need to make sure that any "catch up" test does not rely on the other agents being present (as in registration with ACN is the case). Since the degenerate round is always an end state from which agents do not recover (to the best of my understanding), we could implement a mechanism for them to "slumber" and report less frequently. We could also consider a test-specific degenerate round, but I am not sure this is needed or even desirable. I think opening a ticket on this may be warranted.

Adamantios:
> > Two more appropriate solutions would be to:
> > ```
> > 1. Decrease the interval of checking the logs.
> > 
> > 2. Raise and stop the execution when a degenerate round is reached instead of looping.
> > ```
> 
> in case of "stop the execution", we would need to make sure that any "catch up" test does not rely on the other agents being present (as in registration with ACN is the case). Since the degenerate round is always an end state from which agents do not recover (to the best of my understanding), we could implement a mechanism for them to "slumber" and report less frequently. We could also consider a test-specific degenerate round, but I am not sure this is needed or even desirable. I think opening a ticket on this may be warranted.

Opened #1121.

Karrenbelt (approved):
LGTM

tests/test_agents/test_registration.py
DavidMinarsch, line 57: Ok - the actual problem is this:
DegenerateRound should never be allowed to be part of a composed, non-abstract app. So we should do two things:
1. Change the registration app to end in a non-degenerate round which is simply an "idle" round that is never being left.
2. Add a check in the chaining logic that after chaining no degenerate rounds are present in the final app.

DavidMinarsch (approved):

tests/test_agents/test_registration.py
Adamantios: 1. The registration app needs to end in a degenerate round in order to be chainable. The e2e test checks the registration skill, which is not chained and this is why we end up in a degenerate round when it finishes. 
2. I think we already check for this indirectly because we check for orphans.

Issue #1126, opened 2022-07-28
Tendermint reset edge case


Adamantios:
https://github.com/valory-xyz/open-autonomy/pull/1122#issuecomment-1197976368

Adamantios:
Has not occurred again. Maybe we should close this issue.

0xArdi:
It occurred again [here](https://github.com/valory-xyz/open-autonomy/actions/runs/3539758889/attempts/1).

Adamantios:
@0xArdi I believe that this is not relevant anymore since we are now syncing the agent's state during registration and recovery.

0xArdi:
@Adamantios agreed, we can close this.

PR #1134, opened 2022-07-29
Add a timeout at the optimization procedure itself

Adamantios:
## Proposed changes

This attempts to make the e2e tests run faster and not timeout during the optimization round, by enforcing the optimization study to timeout and return random hyperparams.

Please also take a look at #1164 in order to understand the reasoning of this PR.

## Fixes

n/a

## Types of changes

## Checklist

## Further comments

n/a


tests/test_agents/test_apy_estimation_abci.py
```diff
@@ -68,6 +68,10 @@ class BaseTestABCIAPYEstimationSkillNormalExecution(BaseTestEnd2EndExecution):
             "value": 200,
             "type_": "float",
         },
+        {
+            "dotted_path": f"{__args_prefix}.optimizer.timeout",
+            "value": 1,
+        },
```
Adamantios, line 74: If the first trial requires more time than the round's timeout, then the study will never be interrupted, and the round will timeout. This is because `optuna` checks for the timeout after every trial finishes, and therefore it is possible for it to be exceeded. 

Therefore, we increase the round's timeout and also decrease the optimizer's timeout for the e2e tests.

Adamantios:
- Failing unit test is unrelated and resolved in #1165.
- Failing e2 test related to #1126.

dvilelaf (approved):
LGTM. Is this one flaky? TestSubgraphs::test_top_n_pairs

PR #1142, opened 2022-07-30
Feat/port

DavidMinarsch:
## Proposed changes

Ports all the extensions Ardian developed on academy-1 repo here.

Sibling PR: https://github.com/valory-xyz/agent-academy-1/pull/93

## Fixes

-

## Types of changes

## Checklist

## Further comments

-


packages/valory/connections/abci/connection.py
```diff
@@ -489,7 +1124,7 @@ def init(self) -> None:
         cmd = self._build_init_command()
         subprocess.call(cmd)  # nosec
 
-    def start(self, start_monitoring: bool = False) -> None:
```
DavidMinarsch, line 492: Revert changes here and below

tests/helpers/docker/ganache.py
```diff
@@ -106,3 +106,21 @@ def wait(self, max_attempts: int = 15, sleep_rate: float = 1.0) -> bool:
...
+    BLOCK_NUMBER: int = 11290000
+
+    def _build_command(self) -> List[str]:
+        """Build command."""
+
+        cmd = [
+            "--wallet.deterministic=true",
+            f"--fork.network={self.NETWORK}",
+            f"--fork.blockNumber={self.BLOCK_NUMBER}",
+        ]
```
marcofavorito, line 124: wrt parent class, we do not set gas limit and account/balances:

https://github.com/valory-xyz/open-autonomy/blob/35a90b02d2b30371994796a0935bb2568ef4f83e/tests/helpers/docker/ganache.py#L73-L81

Probably the solution is to call `super()._build_command()` and then append the new options?

DavidMinarsch: @0xArdi agree with @marcofavorito here, any reason you see why we shouldn't do that?

DavidMinarsch (commented):
Fails on same tests as main, so merging now

tests/helpers/docker/ganache.py
0xArdi: `--wallet.accounts` takes precedence over `--wallet.deterministic=true`, but that's not a problem.

PR #1144, opened 2022-07-31
Add `autonomy.test_tools`

marcofavorito:
## Proposed changes

TODO

## Fixes

n/a

## Types of changes

What types of changes does your code introduce? (A **breaking change** is a fix or feature that would cause existing functionality and APIs to not work as expected.)
_Put an `x` in the box that applies_

- [ ] Non-breaking fix (non-breaking change which fixes an issue)
- [ ] Breaking fix (breaking change which fixes an issue)
- [ ] Non-breaking feature (non-breaking change which adds functionality)
- [ ] Breaking feature (breaking change which adds functionality)
- [ ] Refactor (non-breaking change which changes implementation)
- [ ] Messy (mixture of the above - requires explanation!)

## Checklist

_Put an `x` in the boxes that apply._

- [ ] I have read the [CONTRIBUTING](../blob/main/CONTRIBUTING.md) doc
- [ ] I am making a pull request against the `main` branch (left side). Also you should start your branch off our `main`.
- [ ] Lint and unit tests pass locally with my changes
- [ ] I have added tests that prove my fix is effective or that my feature works
- [ ] I have locally run services that could be impacted and they do not present failures derived from my changes

## Further comments

n/a

PR #1146, opened 2022-07-31
chore: move base test classes into skill for easier reuse

DavidMinarsch:
## Proposed changes

Continues the work of #1144 and moves test tools for skill testing into the `abstract_round_abci`. This makes it easier to re-use them for now.

Eventually, a cleanup needs to be done to extract as much as possible into `autonomy` and keep the skills light.

## Fixes

-

## Types of changes

## Checklist

## Further comments

-


packages/valory/skills/abstract_round_abci/test_tools/integration.py
```diff
@@ -120,6 +120,8 @@ class IntegrationBaseCase(FSMBehaviourBaseCase):
         "0x1CBd3b2770909D4e10f157cABC84C7264073C9Ec": "0x47c99abed3324a2707c28affff1267e45918ec8c3f20b8aa892e8b065d2942dd",
     }
     current_agent: Address
+    ROOT_DIR: str
+    make_ledger_api_connection_callable: Callable
```
DavidMinarsch, line 124: Stop-gap for easy porting.

packages/valory/skills/abstract_round_abci/test_tools/integration.py
```diff
@@ -366,7 +368,7 @@ def setup(cls, **kwargs: Any) -> None:
         super().setup()
 
         # register gnosis contract
```
DavidMinarsch, line 370: Technically this pollutes the concerns. But we can clean this up in a next step.

tests/test_skills/test_liquidity_provision_abci/test_behaviours_integration.py
```diff
@@ -104,6 +106,7 @@ class LiquidityProvisionIntegrationBaseCase(
     exit_nonce: int
     swap_back_nonce: int
     default_synchronized_data_hash: LiquidityRebalancingSynchronizedSata
+    ROOT_DIR = ROOT_DIR
```
DavidMinarsch, line 109: Missing make_ledger_api_connection_callable

tests/test_skills/test_transaction_settlement_abci/test_failure_modes.py
```diff
@@ -103,6 +105,7 @@ class TransactionSettlementIntegrationBaseCase(
     """Base case for integration testing TransactionSettlement FSM Behaviour."""
 
     price_estimation_synchronized_data: PriceEstimationSynchronizedSata
+    ROOT_DIR = ROOT_DIR
```
DavidMinarsch, line 109: Missing make_ledger_api_connection_callable

tests/test_skills/test_transaction_settlement_abci/test_failure_modes.py
```diff
@@ -326,6 +329,8 @@ def _test_same_keeper(self) -> None:
 class TestKeepers(OracleBehaviourBaseCase, IntegrationBaseCase):
     """Test the keepers related functionality for the tx settlement skill."""
 
+    ROOT_DIR = ROOT_DIR
+
```
DavidMinarsch, line 333: Same

dvilelaf (commented):
I don't see any problem with the code, but have some questions:
- I would like to understand the benefit of moving all this code into the skills. How does that help us?
- I'm surprised of the amount of nosec statements that are needed. What is the problem here?
- Looking at the changes, I guess test failing are unrelated, but AFAIK TestABCIPriceEstimationFourAgents has been working OK lately, so that one might be related after all.

DavidMinarsch:
> I don't see any problem with the code, but have some questions:
> 
> * I would like to understand the benefit of moving all this code into the skills. How does that help us?
Because other users of the framework need to otherwise manually copy and paste from our repo. Eventually this should be further refactored into the base and plugins.
> * I'm surprised of the amount of nosec statements that are needed. What is the problem here?
The test tools need asserts and since they are not used to be in prod we add the ignores.
> * Looking at the changes, I guess test failing are unrelated, but AFAIK TestABCIPriceEstimationFourAgents has been working OK lately, so that one might be related after all.
This is a flaky test.

dvilelaf (approved):

PR #1155, opened 2022-08-02
Introduce proper deployment flow

angrybayblade:
## Proposed changes

This PR introduces the deployment flow proposed in the deployment flow doc

## Types of changes

## Checklist


angrybayblade:
As mentioned in the doc, the deployments flow will look like this

```
$ autonomy fetch some_service
$ cd some_service
$ autonomy build-images
$ autonomy deploy build
$ cd abci_build
$ docker-compose up --force-recreate
```

@DavidMinarsch It seems docker compose provides a python SDK on installation, we can write a wrapper around this to implement `autonomy build run` command (Just an idea)

For one click deployments 

`$autonomy deploy run TOKEN_ID PATH_TO_KEYS`

This will resolve the token id from on-chain contracts, fetch the relevant service package, build the images and build the deployment.

autonomy/cli/deploy.py
```diff
@@ -120,168 +102,107 @@ def build_group() -> None:
...
         shutil.rmtree(build_dir)
         raise click.ClickException(str(e)) from e
 
 
-def _build_images(
-    build_dir: Path, version: Optional[str], dev_mode: bool = False
+@deploy_group.command(name="from-token")
+@click.argument("token_id", type=int)
+@click.argument("keys_file", type=click.Path())
+@registry_flag()
```
angrybayblade, line 147: Build deployments for on-chain services using their token ids

angrybayblade (commented):
### Update

I've implemented `autonomy deploy from-token` command so It would be less confusing. Using this command we can build deployments for on-chain services using their token ids.

```
$ autonomy deploy from-token 2
```

Now only issue with the flow is images, right now we build images on the runtime which is not a good thing. Both of the dependency images does not require building every time we build a deployment we can just build and push them with every autonomy package release.

The agent runtime image has some issues as well, for example it needs to fetch the required agent twice. First when building the image and second on runtime. When building the image we fetch the agent to install the required dependencies and then on the runtime we fetch the agent for actual use. We also install base dependencies like open-aea and open-autonomy when building an image for an agent. I propose we separate it into 2 images, one is simple base autonomy image which comes with pre-installed base dependencies and derive an agent image from it. The agent image will fetch the agent, install agent dependencies, configure agent and run the agent.

.github/workflows/main_workflow.yml
```diff
@@ -198,12 +198,15 @@ jobs:
...
           pip install .
-          autonomy deploy build deployment valory/oracle_hardhat deployments/keys/hardhat_keys.json --local
+          autonomy fetch --local --service valory/hello_world
+          
+          cd hello_world
+          autonomy build-images
+          autonomy build-images --dev
+          autonomy build-images --dependencies
+          autonomy build-images --dependencies --version 1.0.0
+          autonomy deploy build ../deployments/keys/hardhat_keys.json
```
8ball030, line 208: nice nice nice thank you for this

8ball030 (approved):

setup.py
```diff
@@ -74,6 +74,7 @@ def _get_docker_dependency() -> str:
     "pandas<1.4,>=1.3.4",
     "watchdog >=2.1.6",
     "pytest==7.0.0",
+    "web3==5.25.0",
```
DavidMinarsch, line 77: I prefer we make `open-aea-ledger-ethereum` the dependency. Then we don't need to pay attention between the web3 version here and the one in the plugin.

autonomy/deploy/chain.py
```diff
@@ -0,0 +1,66 @@
...
+CHAIN_CONFIG = {
+    "staging": {
+        "rpc": "https://chain.staging.autonolas.tech",
+        "address": "0x0DCd1Bf9A1b36cE34237eEaFef220932846BCD82",
+    },
+    "mainnet": {
+        "rpc": "https://eth-rpc.gateway.pokt.network",
+        "address": "0x48b6af7B12C71f09e2fC8aF4855De4Ff54e775cA",
+    },
+    "testnet": {
```
DavidMinarsch, line 39: This should be `goerli`. We will eventually support more than 1

autonomy/deploy/chain.py
```diff
@@ -0,0 +1,66 @@
...
+        "rpc": "https://goerli.infura.io/v3/1622a5f5b56a4e1f9bd9292db7da93b8",
+        "address": "0x1cEe30D08943EB58EFF84DD1AB44a6ee6FEff63a",
```
DavidMinarsch, line 41: We really need to stop hardcoding api keys; it's very bad practice!

.github/workflows/main_workflow.yml
DavidMinarsch, line 208: What does `autonomy deploy build ..` do?

autonomy/cli/utils/click_utils.py
```diff
@@ -65,3 +66,22 @@ def wrapper(f: Callable) -> Callable:
         return f
 
     return wrapper
+
+
+def chain_selection_flag(
```
DavidMinarsch, line 71: What if my chain is not supported? There should be an easy way to pass in a set of full chain configs!

DavidMinarsch: Effectively a dict of service contract address and rpc

autonomy/deploy/chain.py
```diff
@@ -0,0 +1,88 @@
...
+    },
+    "ethereum": {
+        "rpc": os.environ.get("ETHEREUM_CHAIN_RPC"),
+        "address": os.environ.get(
+            "SERVICE_ADDRESS_ETHEREUM", "0x48b6af7B12C71f09e2fC8aF4855De4Ff54e775cA"
+        ),
+    },
+    "goerli": {
+        "rpc": os.environ.get("GOERLI_CHAIN_RPC"),
+        "address": os.environ.get(
```
DavidMinarsch, line 48:
```suggestion
        "service_address": os.environ.get(
```
otherwise this is too ambiguous

DavidMinarsch:
> ### Update
> I've implemented `autonomy deploy from-token` command so It would be less confusing. Using this command we can build deployments for on-chain services using their token ids.
> 
> ```
> $ autonomy deploy from-token 2
> ```
> 
neat!

> Now only issue with the flow is images, right now we build images on the runtime which is not a good thing. Both of the dependency images does not require building every time we build a deployment we can just build and push them with every autonomy package release.
> 
Yes, but let's treat this in a separate PR.

> The agent runtime image has some issues as well, for example it needs to fetch the required agent twice. First when building the image and second on runtime. When building the image we fetch the agent to install the required dependencies and then on the runtime we fetch the agent for actual use. We also install base dependencies like open-aea and open-autonomy when building an image for an agent. I propose we separate it into 2 images, one is simple base autonomy image which comes with pre-installed base dependencies and derive an agent image from it. The agent image will fetch the agent, install agent dependencies, configure agent and run the agent.
The problem is that the agent dependencies are specific to an agent. So you cannot get all dependencies in the base by design.

autonomy/cli/utils/click_utils.py
angrybayblade: I added flags `--rpc` and `--sca` (Service contract address) to facilitate this, now you can effectively pass the rpc and address as command line arguments.

.github/workflows/main_workflow.yml
DavidMinarsch: @angrybayblade

angrybayblade: `autonomy deploy build ../deployments/keys/hardhat_keys.json` is the whole command, `autonomy deploy build` builds deployments for the available service.

autonomy/cli/build_images.py
```diff
@@ -0,0 +1,92 @@
...
+@click.option(
+    "--build-dir",
+    type=click.Path(dir_okay=True),
+    help="Path to build dir.",
+)
+@click.option(
+    "--agent",
+    type=PublicIdParameter(),
+    help="Public Id of the agent to use in the docker image.",
+)
```
angrybayblade, line 48: @8ball030 Now you can build images without fetching service components. Just pass the public ID for the agent you want to build the image for.

autonomy/configurations/base.py
```diff
@@ -353,13 +353,11 @@ def try_to_process_nested_fields(  # pylint: disable=too-many-locals
...
-            if len(nums) != self.number_of_agents:
-                raise ValueError("Not enough items in override")
-
-            if nums != set(range(0, self.number_of_agents)):
-                raise ValueError("Overrides incorrectly indexed")
+            override_index = set(field_override.keys())
+            if self.number_of_agents > len(override_index):
+                raise ValueError(
+                    f"Not enough items in override, Number of agents = {self.number_of_agents}; Number of overrides provided = {len(override_index)}"
+                )
```
angrybayblade, line 360: Raising with proper error message

autonomy/cli/deploy.py
```diff
@@ -120,168 +108,207 @@ def build_group() -> None:
...
+            dev_mode=False,
+            force_overwrite=True,
+            number_of_agents=n,
+            version=DEFAULT_IMAGE_VERSION,
+        )
 
-    packages_dir = Path(packages_dir).absolute()
-    skaffold_dir = Path(skaffold_dir).absolute()
+    click.echo("Service build successful.")
+    run_deployment(build_dir)
```
angrybayblade, line 278: Deploying from token

1. Resolve token metadata from chain
2. Fetch required service
3. Build images (Optional)
4. Build deployment
5. Run deployment(API wrapper around `docker-compose'` python SDK)

autonomy/deploy/chain.py
```diff
@@ -0,0 +1,95 @@
...
+            "SERVICE_ADDRESS_ETHEREUM", "0x48b6af7B12C71f09e2fC8aF4855De4Ff54e775cA"
+        ),
+    },
+    "goerli": {
+        "rpc": os.environ.get("GOERLI_CHAIN_RPC"),
+        "service_contract_address": os.environ.get(
+            "SERVICE_ADDRESS_GOERLI", "0x1cEe30D08943EB58EFF84DD1AB44a6ee6FEff63a"
+        ),
+    },
+}
```
8ball030, line 52: Missing polygon

docs/api/cli/deploy.md
```diff
@@ -89,50 +68,85 @@ Build an agent service deployment.
...
+
+<a id="autonomy.cli.deploy.run_deployment_from_token"></a>
+
+#### run`_`deployment`_`from`_`token
+
+```python
+@deploy_group.command(name="from-token")
+@click.argument("token_id", type=int)
+@click.argument("keys_file", type=click.Path())
+@click.option("--rpc", "rpc_url", type=str, help="Custom RPC URL")
```
8ball030, line 124: Nice

autonomy/deploy/chain.py
angrybayblade: Do we have contracts deployed on polygon?

DavidMinarsch: Yes we do. But I think we already support it via the flags I requested. We should keep this restricted to a single mainnet and a single testnet to avoid hardcoded config as much as possible.

8ball030 (approved):
LGTM -> Ticket to update workflow commands to use the new autonomy commands.
https://github.com/valory-xyz/internal-infrastructure/issues/221

dvilelaf (approved):

PR #1165, opened 2022-08-04
Add a retry strategy for the subgraphs tests

### LINKED ISSUE #1128: Subgraph tests
### Subject of the issue
We have some tests in order to test the subgraphs' queries that we are using. However, these tests depend on a) network and b) an external resource, and therefore, they may occasionally fail. For example, test `tests/test_skills/test_apy_estimation_abci/test_apis.py::TestSubgraphs::test_top_n_pairs[UniswapSubgraph-uni_specs_pairs_extended]` fails in [main_workflow #4429](https://github.com/valory-xyz/open-autonomy/runs/7563825270?check_suite_focus=true).

There are 3 potential solutions:
1. The subgraphs cannot change once deployed, therefore we may skip the tests. That would be a problem if we updated a query incorrectly.
2. We may add a retry strategy to these tests.
3. We may mark all the tests that depend on network connectivity/external resources.

Solution (2.) might be the most appropriate at this point.

Adamantios:
## Proposed changes

Adds a retry strategy for the subgraphs tests which depend on the network and may fail occasionally.

## Fixes

Closes #1128.

## Types of changes

## Checklist

## Further comments

n/a

### DIFF
--- a/tests/test_skills/test_apy_estimation_abci/test_apis.py
+++ b/tests/test_skills/test_apy_estimation_abci/test_apis.py
@@ -21,6 +21,7 @@
 import ast
 import logging  # noqa: F401
 import re
+import time
 from typing import Dict, List, Tuple, Type, cast
 
 import pytest
@@ -49,7 +50,7 @@ ResponseItemType = List[Dict[str, str]]
 SubgraphResponseType = Dict[str, ResponseItemType]
 
 
-def make_request(
+def __make_request_no_retries(
     api_specs: Dict, query: str, raise_on_error: bool = True
 ) -> requests.Response:
     """Make a request to a subgraph.
@@ -94,6 +95,27 @@ def make_request(
         raise ValueError(f"Unknown error encountered!\nRaw response: {res}")
 
     return r
+
+
+def make_request(
+    api_specs: Dict,
+    query: str,
+    raise_on_error: bool = True,
+    allowed_retries: int = 5,
+    backoff_factor: int = 1,
+) -> requests.Response:
+    """Make a request using retries."""
+    for i in range(allowed_retries):
+        n_tries = i + 1
+        try:
+            return __make_request_no_retries(api_specs, query, raise_on_error)
+        except (ValueError, ConnectionError) as e:
+            retry_in = backoff_factor * n_tries
+            print(
+                f"Trial {n_tries}/{allowed_retries} failed. Retrying the request in {retry_in} secs. "
+                f"Received an exception:\n{e}"
+            )
+            time.sleep(retry_in)
 
 
 class TestSubgraphs:


tests/test_skills/test_apy_estimation_abci/test_apis.py
```diff
@@ -49,7 +50,7 @@
 SubgraphResponseType = Dict[str, ResponseItemType]
 
 
-def make_request(
+def __make_request_no_retries(
```
Karrenbelt, line 53: for a function this does not matter, but please refrain from using double underscores generally, single underscore marks private and should be clear to other devs. Particularly can be annoying when something if used in methods without actual need for such, leads to name mangling during inheritance (you're basically saying, even child classes should not access this)

Karrenbelt (approved):
This should really be generalized and put into a class with generic helper functions, as I've seen this same logic implemented several places (and yes, I am also a "contributor" to that).

However we can do better than what we have so far done and use built-in libraries entirely. Please open a ticket if you agree with my suggestion, link this PR so we can replace and discard this implementation and others (like my own) after

```
import requests
from requests.adapters import HTTPAdapter, Retry

def requests_retry_session(...):
    s = requests.Session()
    retries = Retry(total=5, backoff_factor=0.1, status_forcelist=[404, ... ])
    s.mount('http://', HTTPAdapter(max_retries=retries))
    s.get(api_specs["url"])
```

https://urllib3.readthedocs.io/en/latest/reference/urllib3.util.html#module-urllib3.util.retry
https://www.peterbe.com/plog/best-practice-with-retries-with-requests

Adamantios:
> This should really be generalized and put into a class with generic helper functions, as I've seen this same logic implemented several places (and yes, I am also a "contributor" to that).
> 
> However we can do better than what we have so far done and use built-in libraries entirely. Please open a ticket if you agree with my suggestion, link this PR so we can replace and discard this implementation and others (like my own) after
> 
> ```
> import requests
> from requests.adapters import HTTPAdapter, Retry
> 
> def requests_retry_session(...):
>     s = requests.Session()
>     retries = Retry(total=5, backoff_factor=0.1, status_forcelist=[404, ... ])
>     s.mount('http://', HTTPAdapter(max_retries=retries))
>     s.get(api_specs["url"])
> ```
> 
> https://urllib3.readthedocs.io/en/latest/reference/urllib3.util.html#module-urllib3.util.retry https://www.peterbe.com/plog/best-practice-with-retries-with-requests

I think we cannot because if you notice the implementation, we get status 200 and a specific error message in the returned dictionary which we need to handle.

DavidMinarsch (approved):

Karrenbelt:
> > This should really be generalized and put into a class with generic helper functions, as I've seen this same logic implemented several places (and yes, I am also a "contributor" to that).
> > However we can do better than what we have so far done and use built-in libraries entirely. Please open a ticket if you agree with my suggestion, link this PR so we can replace and discard this implementation and others (like my own) after
> > ```
> > import requests
> > from requests.adapters import HTTPAdapter, Retry
> > 
> > def requests_retry_session(...):
> >     s = requests.Session()
> >     retries = Retry(total=5, backoff_factor=0.1, status_forcelist=[404, ... ])
> >     s.mount('http://', HTTPAdapter(max_retries=retries))
> >     s.get(api_specs["url"])
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > https://urllib3.readthedocs.io/en/latest/reference/urllib3.util.html#module-urllib3.util.retry https://www.peterbe.com/plog/best-practice-with-retries-with-requests
> 
> I think we cannot because if you notice the implementation, we get status 200 and a specific error message in the returned dictionary which we need to handle.

yes so we generalize it one step further and wrap this basic requests retry with back off method such that it is itself retried when specific message is returned.

PR #1166, opened 2022-08-04
[WIP][Do not merge]Feat/contract exporter

dvilelaf:
## Proposed changes

Experimental work with cookiecutter to export contract packages from contract build in 3 steps:
- [x] Export 1 uniswap erc20 method: approve
- [x] Export all uniswap erc20 methods
- [x] Generalize for other contracts

Note: this tool now correclty generates both Uniswap ERC20 and Router02 contracts, as well as Multisend and OffchainAggregator. Note that for now it just generates the methods corresponding to the contract methods. For extra methods like get_raw_transaction, get_raw_message, get_state we need to further improve the tool.

## Fixes

n/a

## Types of changes

## Checklist


contract_template/cookiecutter.json
```diff
@@ -0,0 +1,9 @@
+{
+    "project_slug": "default_project_slug",
+    "contract_file_name": "default_contract_file_name",
+    "contract_class_name": "default_class_name",
+    "read_functions": {},
+    "write_functions": {},
+    "_jinja2_env_vars": {},
+    "year": "default_year"
+}
```
dvilelaf, line 9: This is a default configuration for cookiecutter that will be overwritten when cookiecutter is called.

contract_template/{{cookiecutter.contract_file_name}}/{{cookiecutter.contract_file_name}}.py
```diff
@@ -0,0 +1,96 @@
+# -*- coding: utf-8 -*-
+# ------------------------------------------------------------------------------
+#
+#   Copyright {{cookiecutter.year}} Valory AG
```
dvilelaf, line 4: Contract template

scripts/contract_config.json
```diff
@@ -0,0 +1,7 @@
+{
+    "project_slug": "uniswap_v2_erc20_exported",
+    "build_path": "packages/valory/contracts/uniswap_v2_erc20/build/IUniswapV2ERC20.json",
+    "output_dir": "packages/valory/contracts/",
+    "contract_file_name": "IUniswapV2ERC20",
+    "contract_class_name": "UniswapV2ERC20Contract"
+}
```
dvilelaf, line 7: This needs to be set by the user. The script will read it, add some more configuration and pass it to cookiecutter.

contract_template/{{cookiecutter.contract_file_name}}/{{cookiecutter.contract_file_name}}.py
```diff
@@ -0,0 +1,96 @@
...
+#
+#   Unless required by applicable law or agreed to in writing, software
+#   distributed under the License is distributed on an "AS IS" BASIS,
+#   WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+#   See the License for the specific language governing permissions and
+#   limitations under the License.
+#
+# ------------------------------------------------------------------------------
+
+"""This module contains the class to connect to a ERC20 contract."""
```
DavidMinarsch, line 20: I don't understand why you want to generate all the functions. Why not just have a sufficiently generic base class so it can interpret any ABI? Then all we need is a package with the ABI and a bit of metadata and send messages to it.

DavidMinarsch: This current feature is more of an `open-aea` than an `open-autonomy` feature in any case.

dvilelaf: You mean having just one generic template for any kind of method instead of separating by method type (view, payable...)? Yes that makes a lot of sense actually. And yes, I'll move this over to open-aea.

DavidMinarsch: I mean even simpler; at runtime feed the message to the right method (constructed on the fly effectively) and execute the call.

DavidMinarsch: Of course, someone might want to add extra methods, so a scaffolded package makes sense with the abi and a very short Contract.py. Then most methods just get auto resolved, the dev can add custom functionality. This approach would also mean there's no need for cookiecutter. Just extend our existing scaffold to handle it.

PR #1174, opened 2022-08-08
Fix/doc hash script

dvilelaf:
## Proposed changes

Better infere package type in the doc ipfs hash checker script.

## Fixes

n/a

## Types of changes

## Checklist

### DIFF
--- a/scripts/check_doc_ipfs_hashes.py
+++ b/scripts/check_doc_ipfs_hashes.py
@@ -165,17 +165,37 @@ class PackageHashManager:
                 return None
 
             # Complete command, succesfully retrieved
-            package_type = None
-            if "deployment" in d["cmd"]:
-                package_type = "service"
-            if d["cmd"] == "fetch":
-                package_type = "agent"
-            if d["cmd"].startswith("add"):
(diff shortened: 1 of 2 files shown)


scripts/check_doc_ipfs_hashes.py
```diff
@@ -165,17 +165,43 @@ def get_hash_by_package_line(
...
+
+                # Deployments are always services
+                if "deployment" in d["cmd"]:
+                    potential_package_types = ["service"]
+
+                # Add commands always specify the package type
+                if d["cmd"].startswith("add"):
+                    potential_package_types = [
+                        d["cmd"].split(" ")[-1]
+                    ]  # i.e.: aea add connection
```
dvilelaf, line 197: @angrybayblade please confirm whether these three assumptions are true

angrybayblade: When we fetch a service with `fetch` we specify it using `--service` flag so you can use that decide if the packages is an agent or a service

dvilelaf: Ok, updated

DavidMinarsch (approved):

PR #1194, opened 2022-08-16
On-chain agent instance deployments

angrybayblade:
## Proposed changes

This PR introduces support for partial deployments for on-chain agent instances.

## Types of changes

## Checklist


autonomy/deploy/base.py
```diff
@@ -69,33 +68,33 @@ def __init__(
...
-            KEY_SCHEMA_UNENCRYPTED_KEY
-            if self.private_keys_password is None
-            else KEY_SCHEMA_ENCRYPTED_KEY
-        )
         for key in keys:
-            for required_key in [KEY_SCHEMA_ADDRESS, key_schema]:
-                if required_key not in key.keys():
-                    raise ValueError("Key file incorrectly formatted.")
-            self.private_keys.append(key[key_schema])
+            if {KEY_SCHEMA_ADDRESS, KEY_SCHEMA_PRIVATE_KEY} != set(key.keys()):
```
angrybayblade, line 88: @8ball030 I don't why we store encrypted private keys as `encrypted_key` if we export them as it is and decrypt them on the agent runtime. I've update the format to simple `address` and `private_key`, Let me know if there is something that I'm missing

autonomy/deploy/base.py
```diff
@@ -126,12 +134,18 @@ def generate_common_vars(self, agent_n: int) -> Dict:
...
         """Generate next agent."""
         agent_vars = self.generate_common_vars(agent_n)
         if len(self.service.overrides) == 0:
             return agent_vars
-        overrides = self.process_model_args_overrides(agent_n)
+
+        if override_idx is None:
+            override_idx = agent_n
+
+        overrides = self.process_model_args_overrides(override_idx)
```
angrybayblade, line 148: We define override index using the agent instances list and export them using respective agent index.

autonomy/deploy/base.py
```diff
@@ -109,7 +108,16 @@ def process_model_args_overrides(self, agent_n: int) -> Dict:
...
-        return [self.generate_agent(i) for i in range(self.service.number_of_agents)]
+        if self.agent_instances is None:
+            return [
+                self.generate_agent(i) for i in range(self.service.number_of_agents)
+            ]
+
+        agent_override_idx = [
+            (i, self.agent_instances.index(kp["address"]))
+            for i, kp in enumerate(self.keys)
+        ]
```
angrybayblade, line 119: Generate override indexes from `agent_instances` list

autonomy/deploy/build.py
```diff
@@ -87,9 +90,12 @@ def generate_deployment(  # pylint: disable=too-many-arguments
...
     )
 
-    deployment.generate(image_versions).generate_config_tendermint(
-        image_versions["tendermint"]
-    ).write_config().populate_private_keys()
+    (
+        deployment.generate(image_versions)
+        .generate_config_tendermint(image_versions["tendermint"])
+        .write_config()
+        .populate_private_keys()
```
angrybayblade, line 97: Just a little bit of re-formatting, looks better this way

autonomy/deploy/chain.py
```diff
@@ -59,37 +59,52 @@ def get_abi(url: str) -> Dict:
...
 
 
-def resolve_token_id(
-    token_id: int,
-    chain_type: str = "staging",
-    rpc_url: Optional[str] = None,
-    service_contract_address: Optional[str] = None,
-) -> Dict:
-    """Resolve token id using on-chain contracts."""
+class ServiceRegistry:
```
angrybayblade, line 62: Class to interact with on-chain contracts.

angrybayblade (commented):
With this PR we can deploy instances for agents registered on the contracts. To deploy an instance you need one or more keys registered for the service. Save it in a keys.json file and just run `autonomy deploy from-token TOKEN_ID keys.json`

autonomy/deploy/base.py
8ball030: for backwards compatibility, we need to support both password and non-password encrypted keys no?

8ball030: Also, we need to co-ordinate rolling this as as i will need to update all my keys and all my backups of those keys

8ball030 (approved):
LGTMN, please co-ordinate with me when this is to be rolled out

autonomy/deploy/base.py
```diff
@@ -69,33 +68,33 @@ def __init__(
...
+            if {KEY_SCHEMA_ADDRESS, KEY_SCHEMA_PRIVATE_KEY} != set(key.keys()):
+                raise ValueError("Key file incorrectly formatted.")
+
+        if self.agent_instances is not None:
+            keys = [kp for kp in keys if kp["address"] in self.agent_instances]
+            self.service.number_of_agents = len(keys)
+
+        self.keys = keys
+        if self.service.number_of_agents > len(self.keys):
+            raise ValueError("Number of agents cannot be greater than available keys.")
```
Karrenbelt, line 97: 1. 
can an agent have multiple keys? If no, check for this as well, if yes, then this last check might lead to a situation where one agent has multiple and another none. Also now complexity is n * m.

```
agent_addresses = set(self.agent_instances)  
keys = [kp for kp in keys if kp["address"] in agent_addresses]
```

2. 
- what if number_of_agents is not set on `__init__` and no agent instances here? 
`self.service.number_of_agents > len(self.keys)` -> `None > len(self.keys)` -> will error out
- what if `self.service.number_of_agents` is already set to `int(x)` and this doesn't match `len(keys)` in the overwrite here?

autonomy/deploy/base.py
```diff
@@ -69,33 +68,33 @@ def __init__(
         keys: Path,
         number_of_agents: Optional[int] = None,
         private_keys_password: Optional[str] = None,
+        agent_instances: Optional[List[str]] = None,
```
Karrenbelt, line 71: we set information that is partially redundant here. Why do we even pass the number_of_agents as a parameter in the first place? Seems like we don't have to. At least I suppose we should check that the size of agent_instances matches the specified number_of_agents. But is there ever a situation where we pass only `number_of_agents` and not `agent_intances`? I think we may want to remove `number_of_agents`

https://github.com/valory-xyz/open-autonomy/blob/693915c76e8fe40907a66223687d2f3ebd6228da/autonomy/deploy/base.py#L65-L72

autonomy/deploy/base.py
DavidMinarsch: 1. no, an agent can have one key only atm.

autonomy/cli/deploy.py
```diff
@@ -194,39 +194,6 @@ def run(build_dir: Path, no_recreate: bool, remove_orphans: bool) -> None:
     run_deployment(build_dir, no_recreate, remove_orphans)
 
 
-def run_deployment(
```
DavidMinarsch, line 197: Was this a complete copy& paste?

DavidMinarsch (commented):
Please update documentation in docs and add tests

autonomy/deploy/base.py
angrybayblade: 1. Agents cannot have multiple keys AFAIK
2. We define `number_of_agents ` in service config, then perform an override in `__init__` and `read_keys` method of service specification object. 
3. I don't understand how it won't match, we need the number of agents to be less than or equal to the number of keys.

autonomy/deploy/base.py
angrybayblade: We create two types of deployments, on-chain deployments and service deployments. Now with on-chain deployments your argument makes sense, but with the local deployments while developing service and components a user might want to experiment with a different number of agents without updating the service config or the keys file so we provide this option here.

autonomy/deploy/base.py
Karrenbelt: not very readable and n ^ 2 complexity also. 

https://github.com/valory-xyz/open-autonomy/blob/693915c76e8fe40907a66223687d2f3ebd6228da/autonomy/deploy/base.py#L109-L120

- remove argument `override_idx` from `generate_agent`. It just reduced legibility
- why not sort agent instances as soon as they get set in accordance with the keys? 
- why do we have the keys attribute if this information is already present on the agent_instances?

```
if self.agent_instances: 
    mapping = {kp["address"]: i for i, kp in enumerate(keys)}
    indices = [mapping[x] for x in agent_instances]
else:
    indices = range(len(self.keys))
return list(map(self.generate_agent, indices))
```

autonomy/deploy/base.py
Karrenbelt: great. Lets add that in a sentence to the docstring or as a comment: "we allow configurable number of agents independent of the number of agent instances for local development purposes"

autonomy/cli/deploy.py
angrybayblade: Yeah, I just moved it below.

autonomy/deploy/base.py
angrybayblade: We still support both encrypted and non encrypted keys, just store them both as `private_key` rather than storing plain keys as `private_key` and encrypted keys as `encrypted_key`.

autonomy/cli/deploy.py
angrybayblade: For the docs, the CLI docs does not need updating but we'll have to add a tutorial on publishing services, deploying them on-chain and running agent instances. I had a talk with jose about this, I'll create appropriate tickets for him.

docs/autonomy.md
```diff
@@ -160,7 +160,7 @@ $ autonomy deploy run
 
 4. One click deployments
 
-If you have a service registered on-chain you can deploy the services directly using token IDs generated after the minting the service component.
+If you have a service registered on-chain you can deploy the services directly using token IDs generated after the minting the service component. Refer [here](https://docs.autonolas.network/package_publishing/) for more information.
```
DavidMinarsch, line 163: They should be relative links

autonomy/deploy/base.py
DavidMinarsch: @angrybayblade please add

docs/autonomy.md
DavidMinarsch: @angrybayblade

autonomy/deploy/base.py
angrybayblade: > why do we have the keys attribute if this information is already present on the agent_instances?

The reason we have both `keys` and `agent_instance` is because the `agent_instances` are the addressed allowed by the service owner to be the part of the deployment and `keys` are defined by the service operator. The `keys` are essentially some funded accounts provided by the service operator to the owner to be a part some service. But they only share the public key or the address of the account, the private keys stays with the service operator, so `agent_instances` are just a list of public keys/account addresses and we need private keys to run the agent instances which will be stored in the `keys.json` file and picked up at the run time. Most of the times the `keys.json` will only hold one account, but we also want to allow a user to run multiple agent instances for a single service so we allow multiple accounts in a single `keys.json` file.

> why not sort agent instances as soon as they get set in accordance with the keys?

This would work if `keys == agent_instances` but it won't be the case unless someone wants to run all n agent instances by themselves.

> remove argument `override_idx` from `generate_agent`. It just reduced legibility

We need that argument because of the situation I explained above, say someone has an addressed which is indexed at position 3 on the `agent_instances` array, In this case we only need to generate 1 agent with the override index being 3. If we go with your logic and just generate `abci3` node instead of `abci0` with overrides being numbered 3 we could face some issues with respective tendermint node.

If the code above seems unnecessarily complicated I can open a ticket to check if your logic affects the tendermint nodes or not if not we can go with your logic.

autonomy/deploy/base.py
```diff
@@ -69,33 +68,41 @@ def __init__(
...
+            if {KEY_SCHEMA_ADDRESS, KEY_SCHEMA_PRIVATE_KEY} != set(key.keys()):
+                raise ValueError("Key file incorrectly formatted.")
+
+        if self.agent_instances is not None:
+            keys = [kp for kp in keys if kp["address"] in self.agent_instances]
+            if not keys:
+                raise ValueError(
+                    "Cannot find the provided keys in the list of the agent instances."
+                )
+            self.service.number_of_agents = len(keys)
```
Karrenbelt, line 101: here it's the same thing right
```
mapping = {kp["address"]: kp for kp in keys}
keys = [mapping[address] for address in self.agent_instances]
```
which will error out in case of a missing key so no check needed either

autonomy/deploy/base.py
Karrenbelt: ok just implement the O(n) solution as suggested then and lets move on

DavidMinarsch (approved):

autonomy/deploy/base.py
Karrenbelt: don't do that, do this: https://github.com/valory-xyz/open-autonomy/pull/1194#discussion_r946522585

PR #1195, opened 2022-08-16
[WIP]: Fix ledger connection

### LINKED ISSUE #1173: asyncio.wait() copies the task container that is passed to it
### Subject of the issue
Contract & ledger operations can be blocked if a previously sent request to the `ledger` connection is blocked. The issue stems from a wrong usage of the `asyncio.wait()` method. 
We know for sure the `ledger` connection is affected, but any other component that uses `asyncio.wait()` in the way described below is affected by it. 

#### What asyncio.wait() does
Let `tasks` be an iterable of tasks (or futures), if we want to wait for ANY of the tasks to be completed, we can do so with `asycnio.wait()` in the following way:

 ```python
 done, pending = await asyncio.wait(tasks, return_when=asyncio.FIRST_COMPLETED)
 ```
The line above will block until at least one of the tasks is done. The completed tasks will be assigned to `done` and those that are still running will be assigned to `pending`. 
Note that its irrelevant as to what the underlying tasks are running and through which mechanism they are running, this includes tasks running on worker threads.

In the [ledger connection](https://github.com/valory-xyz/open-autonomy/blob/v0.1.6/packages/valory/connections/ledger/connection.py) we can see this being used in the `receive` method:
```python
    async def receive(self, *args: Any, **kwargs: Any) -> Optional["Envelope"]:
        """
        Receive an envelope. Blocking.
        :param args: the arguments
        :param kwargs: the keyword arguments
        :return: the envelope received, or None.
        """
        # if there are done tasks, return the result
        if len(self.done_tasks) > 0:  # pragma: nocover
            done_task = self.done_tasks.pop()
            return self._handle_done_task(done_task)

        if len(self.receiving_tasks) == 0:  # pragma: nocover
            self.event_new_receiving_task.clear()
            await self.event_new_receiving_task.wait()

        # wait for completion of at least one receiving task
        done, _ = await asyncio.wait(
            self.receiving_tasks, return_when=asyncio.FIRST_COMPLETED
        )

        # pick one done task
        done_task = done.pop()

        # update done tasks
        self.done_tasks.extend([*done])

        return self._handle_done_task(done_task)
```
 
In the `receiving_tasks` have a list of tasks that are being executed in the background, we await for at least one of them to be completed so that we send a response back:
```python
        done, _ = await asyncio.wait(
            self.receiving_tasks, return_when=asyncio.FIRST_COMPLETED
        )
```
#### Where's the problem?
`asyncio.wait()` [copies the iterable](https://github.com/python/cpython/blob/v3.10.4/Lib/asyncio/tasks.py#L374) that is being passed to it.
Consider the following:
```python
tasks = []

async def receive():
  done, _ = await asyncio.wait(
     tasks, return_when=asyncio.FIRST_COMPLETED
  )

async def send(task):
  tasks.append(task)

send(task_a)
receive_task = asyncio.create_task(receive())
send(task_b)
await receive_task
```
In the current scenario, even though we have two tasks in `tasks`, we are only waiting for `task_a` to be finished, because `asyncio.wait()` [copies the list](https://github.com/python/cpython/blob/v3.10.4/Lib/asyncio/tasks.py#L374) that is being passed to it.

In the ledger connection we have the same setup as the example above.
We have a list of [receiving_tasks](https://github.com/valory-xyz/open-autonomy/blob/v0.1.6/packages/valory/connections/ledger/connection.py#L175) that represents requests being processed. And in the send method we [append to this list](https://github.com/valory-xyz/open-autonomy/blob/v0.1.6/packages/valory/connections/ledger/connection.py#L125).

Now consider the following flow of events:
1. Assume a fresh instance of the ledger connection.
2. Assume the multiplexer has created a task `task_receive` by calling the [receive method](https://github.com/valory-xyz/open-autonomy/blob/v0.1.6/packages/valory/connections/ledger/connection.py#L172) of the ledger connection. Since the ledger connection is fresh, `task_receive` will be [awaiting a request to be received](https://github.com/valory-xyz/open-autonomy/blob/v0.1.6/packages/valory/connections/ledger/connection.py#L172).
3. A request is sent to the ledger connection. Ex: a `transaction_receipt` request. A task `task_a` is [created for this request](https://github.com/valory-xyz/open-autonomy/blob/v0.1.6/packages/valory/connections/ledger/connection.py#L125). 
4. The `event_new_receiving_task` event [is triggered](https://github.com/valory-xyz/open-autonomy/blob/v0.1.6/packages/valory/connections/ledger/connection.py#L127), and `task_receive` will now be [awaiting for `task_a`  to complete](https://github.com/valory-xyz/open-autonomy/blob/v0.1.6/packages/valory/connections/ledger/connection.py#L175).
5. Assume `task_a` is blocked. This can happen for many reasons, a bad node, crypto opts taking longer than usual etc.
6. Assume a contract request is sent to the ledger connection. A task `task_b` is created for this request, and [is added to the receiveing tasks](https://github.com/valory-xyz/open-autonomy/blob/v0.1.6/packages/valory/connections/ledger/connection.py#L125).
7. Assume `task_b` finishes 3 seconds after its creation. Because we are only waiting for `task_a` to finish, `task_b` will wait until `task_a` is finished.

The following snippet reproduces this:
```python
import asyncio


async def blocking_task():
    print("started blocking task")
    await asyncio.sleep(10000)
    print("finished blocking task")


async def normal_task():
    print("started normal task")
    await asyncio.sleep(1)
    print("finished normal task")


tasks = []


async def receive():
    done, _ = await asyncio.wait(
        tasks, return_when=asyncio.FIRST_COMPLETED
    )


def send(task):
    tasks.append(task)


async def main():
    # a blocking task is sent to the con
    send(asyncio.create_task(blocking_task()))

    # the multiplexer awaits for any task to be finished
    receive_task = asyncio.create_task(receive())
    receive_task.add_done_callback(lambda _: print("done"))

    # 3 seconds later a normal task is received
    await asyncio.sleep(3)
    send(asyncio.create_task(normal_task()))

    # the normal task will finish 1 second later but the receive task won't be completed,
    # i.e. "done" won't be printed
    await asyncio.sleep(1000)


asyncio.get_event_loop().run_until_complete(main())
```

Adamantios:
## Proposed changes

Fixes the ledger connection issue, caused by wrong usage of the `asyncio.wait()` method.

## Fixes

Closes #1173 

## Types of changes

## Checklist

## Further comments

n/a


tests/test_connections/test_ledger/test_ledger.py
```diff
@@ -0,0 +1,106 @@
...
+    task = asyncio.create_task(dummy_task())
+    return task
+
+
+async def dummy_receive(receiving_tasks: List[asyncio.Future]) -> None:
+    """A dummy receive method that simply waits for the first task to complete and returns the done and pending."""
+    await asyncio.wait(receiving_tasks, return_when=asyncio.FIRST_COMPLETED)
+
+
+class TestLedgerConnection:
```
DavidMinarsch, line 51: Haven't found the issue yet, but I suggest it would be better in any case to write a custom RequestDispatcher which implements `dummy_task` as a "performative" and is triggered via the message flow. That gets us closer to the real failure.

Adamantios: Managed to recreate the issue in a more minimal way.

Adamantios:
Replaced by #1200.

PR #1199, opened 2022-08-17
Fix/remove third party constant

angrybayblade:
## Proposed changes

Remove the need for `THIRD_PARTY` constant and updates the base image classes to set the third party contracts path on runtime.

## Types of changes

## Checklist

### DIFF
--- a/autonomy/test_tools/configurations.py
+++ b/autonomy/test_tools/configurations.py
@@ -31,8 +31,6 @@ def get_key(key_path: Path) -> str:
 
 CUR_PATH = os.path.dirname(inspect.getfile(inspect.currentframe()))  # type: ignore
 TEST_TOOLS_DIR = Path(CUR_PATH).resolve().absolute()
-ROOT_DIR = Path(CUR_PATH, "..", "..").resolve().absolute()
-THIRD_PARTY = ROOT_DIR / "third_party"
 DATA_PATH = TEST_TOOLS_DIR / "data"
 
 LOCALHOST = "localhost"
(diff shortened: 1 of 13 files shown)


autonomy/test_tools/configurations.py
```diff
@@ -31,8 +31,6 @@ def get_key(key_path: Path) -> str:
 
 CUR_PATH = os.path.dirname(inspect.getfile(inspect.currentframe()))  # type: ignore
 TEST_TOOLS_DIR = Path(CUR_PATH).resolve().absolute()
-ROOT_DIR = Path(CUR_PATH, "..", "..").resolve().absolute()
-THIRD_PARTY = ROOT_DIR / "third_party"
```
Karrenbelt, line 35: so now we always obtain it the THIRD_PARTY as a constant from the conftest.py file in the respective test directory of whatever repository autonomy is installed in, am I understanding correctly?

this then creates a constraint on any repository using the autonomy package to specify these in their `tests/conftest.py` file?

angrybayblade: Not necessarily from the conftest.py, you can set it on the test class like this

https://github.com/valory-xyz/open-autonomy/pull/1199/files#diff-7dbf9570300e3300f3e8c673ea8087d701d596adf755023c02b5035a10c9a803R75

DavidMinarsch (approved):
Let's also open an issue to fully remove the `contracts-amm` (it should not be part of autonomy) and replace the current gnosis safe third party with our own registry repo, so we have 1:1 parity with our deployments.

autonomy/test_tools/configurations.py
Karrenbelt: So now we should provide the necessary contracts ourselves in our repository. Would it be an option to link the respective contracts as submodules along with the package?

for example, `GnosisSafeNetDockerImage` replies on `safe-contracts` specifically.

Adamantios (approved):

PR #1200, opened 2022-08-17
Fix incorrect usage of `asyncio` in ledger connection

### LINKED ISSUE #1173: asyncio.wait() copies the task container that is passed to it

Adamantios:
## Proposed changes

Fixes the ledger connection issue, caused by wrong usage of the `asyncio.wait()` method.
Replaces #1195.

## Fixes

Closes #1173.

## Types of changes

## Checklist

## Further comments

n/a


packages/valory/connections/ledger/connection.py
```diff
@@ -67,9 +63,9 @@ def __init__(self, **kwargs: Any):
         )
 
     @property
-    def event_new_receiving_task(self) -> asyncio.Event:
-        """Get the event to notify the 'receive' method of new receiving tasks."""
-        return cast(asyncio.Event, self._event_new_receiving_task)
+    def response_envelopes(self) -> asyncio.Queue:
+        """Get the response envelopes. Only intended to be accessed when connected."""
+        return cast(asyncio.Queue, self._response_envelopes)
```
DavidMinarsch, line 68:
```suggestion
        if self._response_envelopes is None:
            raise ValueError("asyncio.Queue not set")
        return self._response_envelopes
```

Adamantios: Done in d2ac6547dd2d77e3e009ce936ecbe8d4a027f5a5.

DavidMinarsch (approved):

0xArdi (approved):

Karrenbelt:
```
FAILED tests/test_skills/test_counter_client/test_behaviours.py::TestIncrementerBehaviour::test_run
FAILED tests/test_autonomy/test_cli/test_scaffold_fsm.py::TestScaffoldFSM::test_run
FAILED tests/test_skills/test_counter_client/test_behaviours.py::TestHttpHandler::test_handle
FAILED tests/test_skills/test_counter_client/test_behaviours.py::TestMonitorBehaviour::test_run
```
please [add here](https://docs.google.com/document/d/1f7DrPdTXaLWJ1icTZ1Mt2S3z3ZdzCodrxAG6d1tI3o0) so we can squash them when time is due

packages/valory/connections/ledger/connection.py
```diff
@@ -162,37 +161,15 @@ async def receive(self, *args: Any, **kwargs: Any) -> Optional["Envelope"]:
...
-            await self.event_new_receiving_task.wait()
-
-        # wait for completion of at least one receiving task
-        done, _ = await asyncio.wait(
-            self.receiving_tasks, return_when=asyncio.FIRST_COMPLETED
-        )
-
-        # pick one done task
-        done_task = done.pop()
+        return await self.response_envelopes.get()
```
Karrenbelt, line 164: would it be better to use a timeout here?
 `await asyncio.wait_for(self.agent.outbox.get(), timeout=1)`

packages/valory/connections/ledger/connection.py
```diff
@@ -203,4 +180,6 @@ def _handle_done_task(self, task: asyncio.Future) -> Optional[Envelope]:
                 message=response_message,
                 context=request.context,
             )
-        return response_envelope
+
+        # not handling `asyncio.QueueFull` exception, because the maxsize we defined for the Queue is infinite
+        self.response_envelopes.put_nowait(response_envelope)
```
Karrenbelt, line 185: are you going to address the full queue issue as well? Leaving it at infinite size seems like a bad idea, not?

packages/valory/connections/ledger/connection.py
```diff
@@ -106,12 +106,12 @@ async def disconnect(self) -> None:
 
         self.state = ConnectionStates.disconnecting
 
-        for task in self.receiving_tasks:
+        for task in self.task_to_request.keys():
```
Karrenbelt, line 109: `.keys()` not needed

packages/valory/connections/ledger/connection.py
DavidMinarsch: The whole point is that it doesn't time out.

packages/valory/connections/ledger/connection.py
DavidMinarsch: That would introduce an arbitrary magic value defining the queue size. It's not a bad question, but the answer is more involved as there are other areas of the framework where queues can build up.

PR #1206, opened 2022-08-18
Create a test for receive loop hanging using the multiplexer

Adamantios:
## Proposed changes

Create a test for receive loop hanging using the multiplexer.

## Fixes

n/a

## Types of changes

## Checklist

## Further comments

n/a


tests/test_connections/test_ledger/test_ledger.py
```diff
@@ -96,3 +116,112 @@ async def test_not_hanging(self) -> None:
...
+            [cls.make_ledger_connection_callable()], loop=cls.running_loop
+        )
+        cls._multiplexer.connect()
+        # the ledger connection's connect() is called by the multiplexer
+        # once a connection is ready, `receive()` is called by the multiplexer
+        cls.ledger_connection = cast(LedgerConnection, cls._multiplexer.connections[0])
+
+    @classmethod
+    def teardown(cls) -> None:
+        """Tear down the multiplexer."""
```
DavidMinarsch, line 159: hm I think also need to call teardown on super, but check

tests/test_connections/test_ledger/test_ledger.py
```diff
@@ -69,30 +81,175 @@ async def test_not_hanging(self) -> None:
...
+
+        # the response envelopes of the ledger connection should now contain the `normal_dummy_envelope`
+        assert (
+            not self.ledger_connection.response_envelopes.empty()
+        ), "The response envelopes of the ledger connection should not be empty."
+
+        # `receive()` should be done,
+        # and multiplexer's `_receiving_loop` should have put the `normal_dummy_envelope` in the `in_queue`
+        self._multiplexer.disconnect()
+        envelope = self._multiplexer.get(block=True)
```
Adamantios, line 244: The multiplexer's `in_queue` only gets filled if we call `disconnect()`. Otherwise, the test hangs forever in the `get()`, since [this](https://github.com/valory-xyz/open-aea/blob/fd6d615eae4bd01a7dc5e045eee401b895d3f1bb/aea/multiplexer.py#L524-L526) does not return.

Does this indicate the same issue that we had with the connection in the multiplexer as well?

tests/test_connections/test_ledger/test_ledger.py
Adamantios: `BaseSkillTestCase` does not have a `teardown`.

tests/test_connections/test_ledger/test_ledger.py
DavidMinarsch: Hm I think you have to place an `asyncio.sleep()` here instead of the disconnect

Adamantios: Tried that without luck.

Adamantios: This was an error in the test's implementation. 
We are now using the connection's `send` indirectly via the multiplexer.

tests/test_connections/test_ledger/test_ledger.py
```diff
@@ -69,30 +84,193 @@ async def test_not_hanging(self) -> None:
...
+                performative=LedgerApiMessage.Performative.ERROR,  # type: ignore
+                code=1,
+                message="",
+                data=b"blocking_task",
+            ),
+        )
+        with mock.patch.object(
+            self.ledger_connection, "_schedule_request", return_value=blocking_task
+        ):
+            self.multiplexer.put(blocking_dummy_envelope)
```
Adamantios, line 222: The mocking does not work after f5a32ce2619122c68977f6b8ca74227122b7c0ed and therefore the test fails on the first assertion.
This could be because of the multiplexer running the connection's `send` on a separate thread.

tests/test_connections/test_ledger/test_ledger.py
```diff
@@ -69,30 +84,193 @@ async def test_not_hanging(self) -> None:
...
+                performative=LedgerApiMessage.Performative.ERROR,  # type: ignore
+                code=1,
+                message="",
+                data=b"normal_task",
+            ),
+        )
+        with mock.patch.object(
+            self.ledger_connection, "_schedule_request", return_value=normal_task
+        ):
+            self.multiplexer.put(normal_dummy_envelope)
```
Adamantios, line 244: Same issue as above with patching.

tests/test_connections/test_ledger/test_ledger.py
DavidMinarsch: Yes it does, that's why I think you should use the approach of a custom RequestDispatcher as mentioned before.

DavidMinarsch (approved):

PR #1207, opened 2022-08-18
fix/better gas estimation

0xArdi:


packages/valory/contracts/gnosis_safe/contract.py
```diff
@@ -430,6 +431,7 @@ def get_raw_safe_transaction(  # pylint: disable=too-many-arguments,too-many-loc
         configured_gas = base_gas + safe_tx_gas + 75000
         tx_parameters: Dict[str, Union[str, int]] = {
             "from": sender_address,
+            "gas": configured_gas,
```
0xArdi, line 434: added this to avoid gas estimation when building the transaction, which might lead to GS026 when repricing. This can happen if the node doing the estimation is using the pending state for gas estimation.

packages/valory/contracts/gnosis_safe/contract.py
```diff
@@ -449,15 +451,51 @@ def get_raw_safe_transaction(  # pylint: disable=too-many-arguments,too-many-loc
...
+    @staticmethod
+    def _estimate_gas(ledger_api: EthereumApi, tx: Dict) -> int:
+        """Estimate the gas for the given tx."""
+        estimated_gas: Optional[int] = None
+        transaction_dict = deepcopy(tx)
+        # we delete the set gas to not interfere with the estimation
+        del transaction_dict["gas"]
+
+        try:
+            estimated_gas = ledger_api.api.eth.estimate_gas(transaction_dict)
```
0xArdi, line 473: we try to estimate using the default block identifier.

packages/valory/contracts/gnosis_safe/contract.py
```diff
@@ -449,15 +451,51 @@ def get_raw_safe_transaction(  # pylint: disable=too-many-arguments,too-many-loc
...
+            _logger.warning(
+                "Failed to estimate gas with the default block identifier, "
+                f"{type(e).__name__}: {e.__str__()}"
+            )
+
+        if estimated_gas is None:
+            # estimation using default args may fail if the pending state (block)
+            # is used by the node, which might happen when repricing
+            try:
+                estimated_gas = ledger_api.api.eth.estimate_gas(
```
0xArdi, line 484: this might throw an error, as some nodes don't accept the block identifier as arg.

packages/valory/contracts/gnosis_safe/contract.py
0xArdi: We cannot supply a block identifier on `buildTransaction()`

packages/valory/contracts/gnosis_safe/contract.py
```diff
@@ -449,15 +451,51 @@ def get_raw_safe_transaction(  # pylint: disable=too-many-arguments,too-many-loc
...
+            transaction_dict,
         )
+        transaction_dict["gas"] = Wei(max(estimated_gas + 75000, configured_gas))
         transaction_dict["nonce"] = nonce  # pragma: nocover
 
         return transaction_dict
 
+    @staticmethod
+    def _estimate_gas(ledger_api: EthereumApi, tx: Dict) -> int:
+        """Estimate the gas for the given tx."""
```
Adamantios, line 466: Would it make sense to move this method to the ledger instead?

0xArdi: Yes, https://github.com/valory-xyz/open-aea/pull/268

DavidMinarsch:
Closing in favour of #1209

PR #1212, opened 2022-08-20
fix: addresses gas estimation bug during repricing

DavidMinarsch:
## Proposed changes

This PR addresses an issue with the gas estimation during repricing.

## Fixes

-

## Types of changes

## Checklist

## Further comments

-


packages/valory/connections/ledger/ledger_dispatcher.py
```diff
@@ -254,6 +251,10 @@ def get_transaction_receipt(
             if message.retry_timeout is None
             else message.retry_timeout
         )
+
```
DavidMinarsch, line 254: Rearranged and added logging

packages/valory/contracts/gnosis_safe/contract.py
```diff
@@ -450,21 +453,18 @@ def get_raw_safe_transaction(  # pylint: disable=too-many-arguments,too-many-loc
...
         ):
             tx_parameters.update(ledger_api.try_get_gas_pricing(old_price=old_price))
-        # note, the next line makes an eth_estimateGas call!
-
+        # note, the next line makes an eth_estimateGas call iff gas is not set!
         transaction_dict = w3_tx.buildTransaction(tx_parameters)
-        gas_estimate = (
-            ledger_api._try_get_gas_estimate(  # pylint: disable=protected-access
-                transaction_dict
+        if configured_gas != 1:
```
DavidMinarsch, line 458: We now either price exactly as configured or we use the estimated value. For the estimation, we need to be able to specify a fallback as the estimation can fail.

packages/valory/skills/price_estimation_abci/behaviours.py
```diff
@@ -52,11 +52,11 @@
...
 
 
-# This safeTxGas value is calculated from experimental values plus
-# a 10% buffer and rounded up. The Gnosis safe default value is 0 (max gas)
+# The Gnosis safe default value is 0 (max gas)
 # https://help.gnosis-safe.io/en/articles/4738445-advanced-transaction-parameters
+# This value allows for straightforward gas estimation.
 # More on gas estimation: https://help.gnosis-safe.io/en/articles/4933491-gas-estimation
-SAFE_TX_GAS = 120000
+SAFE_TX_GAS = 0
```
DavidMinarsch, line 59: At some point we can use exact values, but it can interfere with gas estimation and needs to be consistently applied throughout.

packages/valory/skills/transaction_settlement_abci/behaviours.py
```diff
@@ -201,6 +201,7 @@ def _get_tx_data(
         tx_data["tx_digest"] = cast(str, tx_digest)
 
         nonce = Nonce(int(cast(str, message.raw_transaction.body["nonce"])))
+        fallback_gas = message.raw_transaction.body["gas"]
```
DavidMinarsch, line 204: For the repricing, we need a gas estimate, otherwise the transaction will fail!

tests/test_skills/test_transaction_settlement_abci/test_failure_modes.py
```diff
@@ -324,6 +324,32 @@ def _test_same_keeper(self) -> None:
...
+        current implementation of the plugin does not play well with Hardhat:
+        https://github.com/valory-xyz/open-aea/blob/main/plugins/aea-ledger-ethereum/aea_ledger_ethereum/ethereum.py#L1113-L1118
+        https://github.com/NomicFoundation/hardhat/blob/d1733f9ecfe8125111a707ff0e3dea287e584caa/packages/hardhat-core/src/internal/hardhat-network/provider/modules/eth.ts#L419-L424
+        eth_estimateGas
+          Contract call:       GnosisSafeProxy#<unrecognized-selector>
+          From:                0xbcd4042de499d14e55001ccbb24a551f3b954096
+          To:                  0x68fcdf52066cce5612827e872c45767e5a1f6551
+          Value:               0 ETH
+
+          Error: VM Exception while processing transaction: reverted with reason string 'GS026'
```
DavidMinarsch, line 339: https://github.com/safe-global/safe-contracts/blob/main/docs/error_codes.md this indicates that the signature is incorrect, which here is due to the nonce (which would be invalid if the first transaction ever got mined first)

0xArdi: In my understanding the error code is correct, the nonce is increased and the signature is invalid in the pending state (block). The problem is that hardhat doesn't support changing with which state of the blockchain the estimation should be done, i.e. setting the block identifier. This is also true for some other eth node implementations.

packages/valory/skills/transaction_settlement_abci/behaviours.py
0xArdi: On second thought, this makes sense. Reprice gas estimation **must** always be manual. 

- The first instance why it might fail if a fallback is not provided is because the eth node doesnt support the using the last state/block.

- Secondly, even if the estimation during repricing goes through with the last state, we cannot expect for it to result in a different gas from what the original tx was submitted with, given that the original tx is estimated automatically.

tests/test_skills/test_transaction_settlement_abci/test_failure_modes.py
DavidMinarsch: We are saying mostly the same in different ways. From the perspective of tx execution the second tx is invalid of the first tx has been committed (which is what happens when gas estimation is run in pending state). The invalidity stems from the nonce on the safe side (not the tx nonce!) being identical for both txs.

0xArdi: I was referring to the safe nonce.

PR #1216, opened 2022-08-22
Chore/spell check

DavidMinarsch:
## Proposed changes

Fixes the spell-check so it actually exits properly on error.

## Fixes

-

## Types of changes

## Checklist

## Further comments

-

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -75,9 +75,9 @@ security:
 .PHONY: generators
 generators:
 	tox -e abci-docstrings
-	tox -e check-copyright
+	tox -e fix-copyright
 	python -m autonomy.cli hash all
-	tox -e generate_api_documentation
+	tox -e generate-api-documentation
 	tox -e fix-doc-hashes
 
 .PHONY: common-checks-1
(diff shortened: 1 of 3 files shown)


Makefile
```diff
@@ -75,9 +75,9 @@ security:
 .PHONY: generators
 generators:
 	tox -e abci-docstrings
-	tox -e check-copyright
+	tox -e fix-copyright
```
DavidMinarsch, line 78: generator should fix

tox.ini
```diff
@@ -397,10 +397,11 @@ deps =
 commands = aea generate-all-protocols --check-clean
 
 [testenv:spell-check]
+whitelist_externals = mdspell
 skipsdist = True
 usedevelop = True
 deps =
-commands = {toxinidir}/scripts/spell-check.sh
+commands = mdspell -r -n -a --en-gb '**/*.md' '!docs/api/**/*.md' '!third_party/**/*.md'
```
DavidMinarsch, line 404: without the `-r` it doesn't exit on error

PR #1220, opened 2022-08-24
Pin ML libraries

Adamantios:
## Proposed changes

Pin the ML libraries.

## Fixes

n/a

## Types of changes

## Checklist

## Further comments

n/a


DavidMinarsch:
First, we should pin exactly in the context of an application and with ranges in the context of a framework.

`abstract_round_abci` is a framework more than an application, as it is used in many fsm apps.

I propose the following:
1) Remove `ForecasterLoader`, `ForecasterStorer`, `CSVLoader` and `CSVStorer` from `abstract_round_abci` and move them into the only skill that needs them (`apy_estimation_abci`).
2) Move the dependencies `pmdarima` and `joblib` into `apy_estimation_abci` and pin exactly. Same for new `scikit-learn` dep.

(Note, in Pipfile and tox.ini we pin exactly for a reproducible test env.)

DavidMinarsch:
The failures on Windows are known. What about the e2e failures @Adamantios ?

DavidMinarsch (approved):

dvilelaf (approved):

PR #1222, opened 2022-08-24
[WIP] Introduce cluster image

angrybayblade:
## Proposed changes

Introduce cluster image to fix the build issues on kubernetes cluster.

## Types of changes

## Checklist


deployments/Dockerfiles/cluster/run.sh
```diff
@@ -0,0 +1,50 @@
+#! /bin/bash
```
DavidMinarsch, line 1: Is this identical to the other run.sh? If so, we might want to come up with a smarter approach to avoid duplication

deployments/Dockerfiles/autonomy/Dockerfile
```diff
@@ -21,5 +21,10 @@ RUN python -m pip install --upgrade pip
 WORKDIR /home/ubuntu
 
 RUN pip3 install -r requirements.txt
+RUN aea init --reset --remote --ipfs --author default_author
```
DavidMinarsch, line 24: We had decided this shouldn't be here... now it's back...

deployments/Dockerfiles/autonomy/run.sh
```diff
@@ -0,0 +1,7 @@
+#! /bin/bash
+
+sudo chmod +x ./install.sh
+sudo chmod +x ./start.sh
+
+./install.sh
+./start.sh
```
DavidMinarsch, line 7: Why do we run start in the base image?

deployments/Dockerfiles/autonomy/Dockerfile
angrybayblade: To answer this and https://github.com/valory-xyz/open-autonomy/pull/1222#discussion_r954021750 , as you said we don't need a separate image for cluster but we can't use the OAR image as well since it needs to built before running it. 

So to fix this I made the base image runnable by default. We do it by making `run.sh` the default executable for the base image 

https://github.com/valory-xyz/open-autonomy/blob/23079e6bec03f6d9386a03c3f104848914598502/deployments/Dockerfiles/autonomy/Dockerfile#L30

This means if you run the base image directly it'll both install and run image without we need to build it first. It'll need the same env vars as the OAR image to run for a particular agent. And will need `AEA_AGENT` env var by default. 

Now when we use the base image to build OAR image we run `install.sh` when building the image for a particular agent and update the default executable to be `start.sh`

https://github.com/valory-xyz/open-autonomy/blob/23079e6bec03f6d9386a03c3f104848914598502/autonomy/data/Dockerfiles/agent/Dockerfile#L15

Now the author thing, I put this 

https://github.com/valory-xyz/open-autonomy/blob/23079e6bec03f6d9386a03c3f104848914598502/deployments/Dockerfiles/autonomy/Dockerfile#L24

here because we'll need a repo config when we run the base image directly, but it'll be overwritten when we build the OAR image.

https://github.com/valory-xyz/open-autonomy/blob/23079e6bec03f6d9386a03c3f104848914598502/autonomy/data/Dockerfiles/agent/Dockerfile#L12

angrybayblade:
We don't need the cluster image since we can build the images in a job we run before running clusters.

PR #1223, opened 2022-08-24
Rename shadowing `io`

### LINKED ISSUE #901: Shadowing package name
### Subject of the issue
Current `main` (2b14bf8f23e9e02e9d17246212aabc5d714cae79) contains a [package](https://github.com/valory-xyz/consensus-algorithms/tree/2b14bf8f23e9e02e9d17246212aabc5d714cae79/packages/valory/skills/abstract_round_abci/io) which shadows `io` module. We may rename it to `io_`.

Adamantios:
## Proposed changes

Rename shadowing `io` to `io_`

## Fixes

Fixes #901.

## Types of changes

## Checklist

## Further comments

n/a


DavidMinarsch (approved):

Karrenbelt (commented):
Looks like a nonissue; why not simply do the same we do everywhere else?
`from packages.valory.skills.abstract_round_abci import io as abstract_round_abci_io`

that said I didn't even see io being imported anywhere.

Adamantios:
> Looks like a nonissue; why not simply do the same we do everywhere else? using `import x as y`
> 
> that said I didn't even see io being imported anywhere. Can you explain why this is necessary? not clear from the ticket or the code changes to me. If not I suggest closing the PR.

The fact that we did not *yet* have an issue, does not mean that we should continue having `io` shadowed. It is a bad practice in general to shadow the standard library. Also, minor, but it was causing [issues](https://youtrack.jetbrains.com/issue/PY-23615/Dont-break-when-debugging-files-with-names-shadowing-standard-modules) with PyCharm's debugger.

Karrenbelt (approved):

PR #1224, opened 2022-08-24
Move Loaders and Storers to relevant skill

Adamantios:
## Proposed changes

Move Loaders and Storers to relevant APY skill:
https://github.com/valory-xyz/open-autonomy/pull/1220#issuecomment-1225483764

## Fixes

n/a

## Types of changes

## Checklist

## Further comments

n/a


packages/valory/skills/abstract_round_abci/skill.yaml
```diff
@@ -17,9 +17,9 @@ fingerprint:
   handlers.py: bafybeidammqyvz2wvqirr3it4tfy7vbe5kxybwkz5o3y4ogo3wrbaqwcmy
   io_/__init__.py: bafybeiattfgecxl5y276etqe5jiuhqk3grcov6qky5cgokstmwhrbaodri
   io_/ipfs.py: bafybeic2lil23fk77wk3i6jrx47fdmc5wxr4dl2xbjjomgtfkz7yesz4eq
-  io_/load.py: bafybeiamag3kolsuv75skjgzfyosip3qqpdqzadet6niqej7x4wdostgdm
+  io_/load.py: bafybeifdki3qeq4aplu6chpozysmeqrt23lxhvcmlefsbesdrklh4wxu64
   io_/paths.py: bafybeicfno2l4vwtmjcm3rzpp6tqi3xlkof47pypf5teecad22d44u2ple
-  io_/store.py: bafybeiczvoewpvhlblwkewsx4b27urji73vgssrhxkkpqf4a6eesmqp62m
+  io_/store.py: bafybeihksp4ch67ae3lum5ou62mmctbl4xnodkiewdlbilmgy752m4pwc4
   models.py: bafybeifsna4hnkhgqegmhzh2uh62q6ikplvoqcsujnnvpol3wqozdhpcn4
   serializer.py: bafybeiezjdtp33nub7ual5ibt4vi7dj3ilmyw7wqq374ctub5oo7wgp7mm
```
DavidMinarsch, line 24: Also remove the dependencies below

Adamantios:
This PR will fail, but is just intended in order for the refactoring to be more transparent and separated from fixing anything that breaks. 
#1225 is intended to fix anything that breaks on this PR.

packages/valory/skills/abstract_round_abci/skill.yaml
Adamantios: Done in 814d7f16736b67c64627c48a755f8bddd33f8209.

packages/valory/skills/apy_estimation_abci/io_/load.py
```diff
@@ -0,0 +1,90 @@
...
+    def load_single_file(self, path: str) -> NativelySupportedSingleObjectType:
+        """Load a `pmdarima` forecaster.
+
+        :param path: path to store the forecaster.
+        :return: a `pmdarima.pipeline.Pipeline`.
+        """
+        try:
+            return joblib.load(path)
+        except (NotADirectoryError, FileNotFoundError) as e:  # pragma: no cover
+            raise IOError(f"Could not detect {path}!") from e
```
Karrenbelt, line 70: quite frankly just excessive code. This can all be deleted, and `pd.read_csv` and `joblib.load` used where these classes and their single method (!) are referenced. Classes with a single method that do not reference `self` should not exist. Since this is a refactor now would be the perfect time.

Bonus: we get rid of some `# pragma: no cover`

packages/valory/skills/apy_estimation_abci/io_/load.py
```diff
@@ -0,0 +1,90 @@
...
+        """Initialize a `Loader`."""
+        super().__init__(filetype, custom_loader)
+
+        self.__filetype_to_loader: Dict[ExtendedSupportedFiletype, SupportedLoaderType]
+        self.__filetype_to_loader[
+            ExtendedSupportedFiletype.PM_PIPELINE
+        ] = ForecasterLoader().load_single_file
+        self.__filetype_to_loader[
+            ExtendedSupportedFiletype.CSV
+        ] = CSVLoader().load_single_file
```
Karrenbelt, line 90: why introduce name mangling with the `__` prefix? Also this whole class is just a mapping, why not use a `dict`?

packages/valory/skills/apy_estimation_abci/io_/store.py
```diff
@@ -0,0 +1,123 @@
...
+        """Store a pmdarima Pipeline."""
+        if not isinstance(obj, Pipeline):
+            raise ValueError(  # pragma: no cover
+                f"`JSONStorer` cannot be used with a {type(obj)}! Only with a {Pipeline}"
+            )
+
+        try:
+            joblib.dump(obj, filename)
+        except (ValueError, OSError) as e:  # pragma: no cover
+            raise IOError(str(e)) from e
```
Karrenbelt, line 102: looks like we can get rid of some more `# pragma: no cover` here :upside_down_face:

packages/valory/skills/apy_estimation_abci/io_/store.py
```diff
@@ -0,0 +1,123 @@
...
+
+
+class Storer(BaseStorer):
```
Karrenbelt, line 105: another `dict`

packages/valory/skills/apy_estimation_abci/io_/load.py
Adamantios: No, we cannot just use `pd.read_csv` and `joblib.load`. We want the `IPFSInteract` class to be able to call a `store` or `load` method and look for a single exception, `IOError` (ideally should be a custom exception), and a relevant message. We need these methods in order to filter all the exceptions and raise the relevant ones.

Moreover, these classes do not have a single method. They **override** a single method.

packages/valory/skills/apy_estimation_abci/io_/load.py
Adamantios: This PR only moves things. This has been converted to `_` in #1225.

packages/valory/skills/apy_estimation_abci/io_/store.py
Adamantios: I hate them too; these classes need tests, but we should prioritize properly.

packages/valory/skills/apy_estimation_abci/io_/store.py
Adamantios: Why do you think it is a single `dict`? It has `store_single_file` and `store`.

packages/valory/skills/apy_estimation_abci/io_/load.py
Karrenbelt: thanks for clarifying. Yes custom exception for intercepting makes sense, but not obvious since that is not what is going on

Karrenbelt (approved):

DavidMinarsch (approved):

PR #1225, opened 2022-08-24
Fix Storers and Loaders

Adamantios:
## Proposed changes

Fixes the storer and Loader functionality after #1224.

## Fixes

n/a

## Types of changes

## Checklist

## Further comments

n/a


packages/valory/skills/abstract_round_abci/behaviour_utils.py
```diff
@@ -372,9 +373,11 @@ def __init__(self, **kwargs: Any):
         # If `ipfs_domain_name` is not specified for the skill, then we get a `None` default.
         # Therefore, `IPFSBehaviour` will be disabled.
         domain = getattr(self.params, "ipfs_domain_name", None)  # type: ignore  # pylint: disable=E1101
+        loader_cls = kwargs.pop("loader_cls", Loader)
+        storer_cls = kwargs.pop("storer_cls", Storer)
```
DavidMinarsch, line 377: What's this for?

Adamantios: Now we have a basic `Loader` and `Storer` on the abstract skill, which can only load and store `json` files. However, we can create any store or loader, and pass it here. This is how we pass the extended versions for the APY skill, which also support csvs and pipelines.

DavidMinarsch:
Interesting hypothesis failure `https://github.com/valory-xyz/open-autonomy/runs/8000224737?check_suite_focus=true`

dvilelaf (approved):

Adamantios:
> https://github.com/valory-xyz/open-autonomy/runs/8000224737?check_suite_focus=true

Yes, we should think of something about the `hypothesis` tests timing out, I have noticed that happening too often on CI for various tests using `hypothesis`:

```
Unreliable test timings! On an initial run, this test took 9080.57ms, which exceeded the deadline of 5000.00ms, but on a subsequent run it took 1175.99 ms, which did not. If you expect this sort of variability in your test timings, consider turning deadlines off for this test by setting deadline=None.
```

DavidMinarsch:
> > https://github.com/valory-xyz/open-autonomy/runs/8000224737?check_suite_focus=true
> 
> Yes, we should think of something about the `hypothesis` tests timing out, I have noticed that happening too often on CI for various tests using `hypothesis`:
> 
> ```
> Unreliable test timings! On an initial run, this test took 9080.57ms, which exceeded the deadline of 5000.00ms, but on a subsequent run it took 1175.99 ms, which did not. If you expect this sort of variability in your test timings, consider turning deadlines off for this test by setting deadline=None.
> ```

I didn't look into the test in detail but I think it's doing external requests? If so, then I don't think this should be done with hypothesis at all.

DavidMinarsch (approved):
LGTM

Adamantios:
> > > https://github.com/valory-xyz/open-autonomy/runs/8000224737?check_suite_focus=true
> > 
> > 
> > Yes, we should think of something about the `hypothesis` tests timing out, I have noticed that happening too often on CI for various tests using `hypothesis`:
> > ```
> > Unreliable test timings! On an initial run, this test took 9080.57ms, which exceeded the deadline of 5000.00ms, but on a subsequent run it took 1175.99 ms, which did not. If you expect this sort of variability in your test timings, consider turning deadlines off for this test by setting deadline=None.
> > ```
> 
> I didn't look into the test in detail but I think it's doing external requests? If so, then I don't think this should be done with hypothesis at all.

No, this is a simple test that uses `hypothesis` to generate a `list` of `int`s, passes the `list` to a protected attribute of the behaviour object, and then asserts that the property utilizing this attribute is correct.

Karrenbelt:
> > > > https://github.com/valory-xyz/open-autonomy/runs/8000224737?check_suite_focus=true
> > > 
> > > 
> > > Yes, we should think of something about the `hypothesis` tests timing out, I have noticed that happening too often on CI for various tests using `hypothesis`:
> > > ```
> > > Unreliable test timings! On an initial run, this test took 9080.57ms, which exceeded the deadline of 5000.00ms, but on a subsequent run it took 1175.99 ms, which did not. If you expect this sort of variability in your test timings, consider turning deadlines off for this test by setting deadline=None.
> > > ```
> > 
> > 
> > I didn't look into the test in detail but I think it's doing external requests? If so, then I don't think this should be done with hypothesis at all.
> 
> No, this is a simple test that uses `hypothesis` to generate a `list` of `int`s, passes the `list` to a protected attribute of the behaviour object, and then asserts that a property utilizing this attribute returns the expected result based on the generated `list`.

I wondered what is happening with `synchronized_data` during tests. I see you use a setup() which will run every time for each method - makes sense since you don't want the state of synchronized data set in one test to affect another.
But since setup runs each time, that may contribute to the variability.
I suggest to setup once (`setup_class`), then the only state that changes is the synchronized data (??) and you can implement `setup` or `teardown` to clear the synchronized data at the start or end of each test (method).
https://github.com/valory-xyz/open-autonomy/blob/33e4a8649af30eeadd1398d6a5746dfaddc45530/packages/valory/skills/abstract_round_abci/test_tools/base.py#L79-L81

PR #1227, opened 2022-08-25
fix: connection test with multiplexer

0xArdi:


tests/test_connections/test_ledger/test_ledger.py
```diff
@@ -281,6 +281,7 @@ def get_normal(
             dialogue.reply(
                 performative=DummyLedgerApiMessage.Performative.NORMAL,  # type: ignore
                 target_message=message,
+                data=b"normal_task",
```
0xArdi, line 284: the last assert in the file requires it

tests/test_connections/test_ledger/test_ledger.py
```diff
@@ -258,7 +258,7 @@ class DummyRequestDispatcher(RequestDispatcher):
     def __init__(self, *args: Any, **kwargs: Any) -> None:
         """Initialize the dispatcher."""
         super().__init__(*args, **kwargs)
-        self._ledger_api_dialogues = DummyLedgerApiDialogues(SOME_SKILL_ID)
+        self._ledger_api_dialogues = DummyLedgerApiDialogues(LedgerConnection.connection_id.__str__())
```
0xArdi, line 261: The dialogues sit in the connection, I think.

tests/test_connections/test_ledger/test_ledger.py
```diff
@@ -442,6 +443,7 @@ async def test_not_hanging_with_multiplexer(self) -> None:
 
         # create a blocking task lasting `blocking_time` secs
         request, _ = self.create_ledger_dialogues()
+        request._sender = SOME_SKILL_ID
```
0xArdi, line 446: the sender of the message is the skill, by default this is set to the connection. There probably is a better way to set this than like this.

tests/test_connections/test_ledger/test_ledger.py
```diff
@@ -451,6 +453,7 @@ async def test_not_hanging_with_multiplexer(self) -> None:
         await asyncio.sleep(wait_time_among_tasks)
 
         request, _ = self.create_ledger_dialogues(blocking=False)
+        request._sender = SOME_SKILL_ID
```
0xArdi, line 456: Same as above.

tests/test_connections/test_ledger/test_ledger.py
```diff
@@ -471,7 +474,6 @@ async def test_not_hanging_with_multiplexer(self) -> None:
 
         # `receive()` should be done,
         # and multiplexer's `_receiving_loop` should have put the `normal_dummy_envelope` in the `in_queue`
-        self.multiplexer.disconnect()
```
0xArdi, line 474: Feel free to add this, if necessary.

tests/test_connections/test_ledger/test_ledger.py
```diff
@@ -412,7 +414,6 @@ def create_ledger_dialogues(
                 counterparty=str(self.ledger_connection.connection_id),
                 performative=performative,  # type: ignore
                 ledger_id=EthereumCrypto.identifier,
-                address=EthereumCrypto(ETHEREUM_KEY_DEPLOYER).address,
```
0xArdi, line 415: This is not needed, it will **not** break the test.

tests/test_connections/test_ledger/test_ledger.py
Adamantios: It should not be necessary, it was just a residue from a previous commit that was demonstrating an issue.

DavidMinarsch: Yeah let's also remove the comment then :)

PR #1231, opened 2022-08-25
fix: enables registry flag support for fsm scaffold and pins exact ar…

DavidMinarsch:
…_abci skill

## Proposed changes

This PR fixes two issues identified today with @Bapi-Reddy and @dvilelaf:
1. `autonomy scaffold fsm` does not support registry flags
2. The `abstract_round_abci` dependency is not explicitly tagged, so there's no support for remote registry.

With the fix the command now works with both `--remote/local` flags and also fetches the exact dependency required.

## Fixes

-

## Types of changes

## Checklist


DavidMinarsch (commented):
@angrybayblade would appreciate if you could add a test for the registry flag case

PR #1232, opened 2022-08-26
[WIP] Add tests for remote repo in fsm scaffold tool

angrybayblade:
## Proposed changes

This PR adds tests for remote repo in fsm scaffold tool

## Types of changes

## Checklist

## Further comments

### DIFF
--- a/tests/test_autonomy/test_cli/test_scaffold_fsm.py
+++ b/tests/test_autonomy/test_cli/test_scaffold_fsm.py
@@ -21,6 +21,7 @@
 from pathlib import Path
 from typing import Tuple
 
+import pytest
 from aea.configurations.constants import PACKAGES, SKILLS
 from aea.test_tools.test_cases import AEATestCaseEmpty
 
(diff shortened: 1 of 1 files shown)


DavidMinarsch (changes requested):
That won’t work. This would only pass each time if we pushed abstract round abci skill and all it’s dependencies for each PR, not feasible. The remote case needs to be mocked

angrybayblade:
> That won’t work. This would only pass each time if we pushed abstract round abci skill and all it’s dependencies for each PR, not feasible. The remote case needs to be mocked

If this is the case then is a bigger issue than just a test failing in the CI since the package hash is hardcoded for `abstract_round_abci` and will be only updated at the release. So if the test fails in the CI the users won't be able to use the `--remote` flag when scaffolding a FSM. 

We can fix this by either making the hash constant configurable using env var or adding a CLI arg to pass down the latest hash for abstract round package

DavidMinarsch:
> > That won’t work. This would only pass each time if we pushed abstract round abci skill and all it’s dependencies for each PR, not feasible. The remote case needs to be mocked
> 
> If this is the case then is a bigger issue than just a test failing in the CI since the package hash is hardcoded for `abstract_round_abci` and will be only updated at the release. So if the test fails in the CI the users won't be able to use the `--remote` flag when scaffolding a FSM.
> 
> We can fix this by either making the hash constant configurable using env var or adding a CLI arg to pass down the latest hash for abstract round package

No it's no bigger issue. The package is hard coded (there is no other way) and during development no-one needs to be able to fetch it remotely. At release the package is available remotely. We do the same in `open-aea` for `open_aea/signing` protocol.

No-one wants to have to find out what hash to use - horrible ux

PR #1243, opened 2022-08-26
Fix/hash check flags

dvilelaf:
## Proposed changes

Fix the hash check regex to also get command flags after the hash.

## Fixes

n/a

## Types of changes

## Checklist


scripts/check_doc_ipfs_hashes.py
```diff
@@ -35,8 +35,9 @@
 VENDOR_REGEX = rf"(?P<vendor>{SIMPLE_ID_REGEX})"
 PACKAGE_REGEX = rf"(?P<package>{SIMPLE_ID_REGEX})"
 VERSION_REGEX = r"(?P<version>\d+\.\d+\.\d+)"
+FLAGS_REGEX = r"(?P<flags>(\s--.*)?)"
 
-AEA_COMMAND_REGEX = rf"(?P<full_cmd>{CLI_REGEX} {CMD_REGEX} (?:{VENDOR_REGEX}\/{PACKAGE_REGEX}:{VERSION_REGEX}?:?)?(?P<hash>{IPFS_HASH_REGEX}))"
+AEA_COMMAND_REGEX = rf"(?P<full_cmd>{CLI_REGEX} {CMD_REGEX} (?:{VENDOR_REGEX}\/{PACKAGE_REGEX}:{VERSION_REGEX}?:?)?(?P<hash>{IPFS_HASH_REGEX}){FLAGS_REGEX})"
```
dvilelaf, line 40: This regex wasnt accounting for flags after the hash, so when checking if "--service" was present in the command, it could find it so it guessed it was an agent.

scripts/check_doc_ipfs_hashes.py
```diff
@@ -232,11 +236,13 @@ def check_ipfs_hashes(  # pylint: disable=too-many-locals,too-many-statements
...
-        for match in re.findall(AEA_COMMAND_REGEX, content):
+        for match in [m.groupdict() for m in re.finditer(AEA_COMMAND_REGEX, content)]:
             matches += 1
-            doc_full_cmd = match[0]
-            doc_cmd = match[2]
-            doc_hash = match[-1]
+            doc_full_cmd = match["full_cmd"]
+            doc_cmd = match["cmd"]
+            doc_hash = match["hash"]
+            flags = match["flags"]
```
dvilelaf, line 244: We now use groupdict, as it is more secure. The old method returned duplicated items so accesing by index was not safe.

DavidMinarsch (approved):

scripts/check_doc_ipfs_hashes.py
angrybayblade: This assumes we only use flags at the end or after public id argument, but we can also use flags before public id. This will work for now, but please open an issue to address this issue.

dvilelaf: Not exactly: we were already accounting for flags before the public_id (we get that on the CMD_REGEX) so it shouldnt be a problem, but that one is not a very good regex for now. Actually I'm having similar issues with the other script I'm working on to validate commands. First i need to understand the command requirements, and then I'll improve this and the other script.

angrybayblade (approved):

PR #1256, opened 2022-08-30
Add support for nested overrides on connection config

angrybayblade:
## Proposed changes

This PR adds support for nested overrides on connection config.

## Types of changes

## Checklist

## Further comments


angrybayblade:
Although this patch works, we need to keep it in the mind that this is a patch and needs to be addressed properly on open-aea. Right now only skill config has support for nested overrides which should not be the case, I'll open a PR on open-aea to address this later this week.

autonomy/configurations/base.py
```diff
@@ -44,6 +41,19 @@
...
+    FIELDS_WITH_NESTED_FIELDS: FrozenSet[str] = frozenset(
+        [
+            "config",
+        ]
+    )
+    NESTED_FIELDS_ALLOWED_TO_UPDATE: FrozenSet[str] = frozenset(
+        ["ledger_apis", "ethereum"]
+    )
+
+
```
8ball030, line 56: How come this was not needed for models?

autonomy/configurations/base.py
```diff
@@ -366,12 +366,6 @@ def try_to_process_nested_fields(  # pylint: disable=too-many-locals
                         nested_override_key,
                         nested_override_value,
                     ) in nested_value.items():
-                        if (
-                            nested_override_key
-                            not in config_class.NESTED_FIELDS_ALLOWED_TO_UPDATE  # type: ignore
-                        ):
-                            raise ValueError("Trying to override non-nested field.")
```
angrybayblade, line 373: We don't need to check for `NESTED_FIELDS_ALLOWED_TO_UPDATE`, we're planning to remove this variable in one of the coming releases

Adamantios (commented):
Can we add a test for this?

autonomy/configurations/base.py
```diff
@@ -391,4 +385,5 @@ def try_to_process_nested_fields(  # pylint: disable=too-many-locals
 PACKAGE_TYPE_TO_CONFIG_CLASS = {
     **_PACKAGE_TYPE_TO_CONFIG_CLASS,
     PackageType.SERVICE: Service,
+    PackageType.CONNECTION: ConnectionConfig,
```
DavidMinarsch, line 388: I think this can go. `PackageType.CONNECTION` is already specified in `_PACKAGE_TYPE_TO_CONFIG_CLASS`

DavidMinarsch (changes requested):

autonomy/configurations/base.py
angrybayblade: https://github.com/valory-xyz/open-aea/pull/298/files#diff-9d90463edcd0573846e5906abe4e7def07e18ad71ed8b7ae1da9f8e33330ec4bR33-R39

tests/test_base.py
```diff
@@ -87,9 +88,10 @@ def test_load_all_packages(component_type: PackageType, config_file_path: str) -
     )
     with open_file(config_file_path) as fp:
         configuration_object = configuration_loader.load(fp)
-        directory = Path(config_file_path).parent
-        configuration_object.directory = directory
-        load_aea_package(configuration_object)
+        if component_type != PackageType.SERVICE:
+            directory = Path(config_file_path).parent
+            configuration_object.directory = directory
+            load_aea_package(configuration_object)
```
DavidMinarsch, line 94: I think this line should not be nested?

Why are services not being assigned a directory?

angrybayblade: Sorry, for not providing context. The nested logic also loads dependency packages which is not required in service component. We only want to check if we can load a service component config without any error.

DavidMinarsch: This comment should really be there in code. I will merge into the base branch as you need to fix things there but please add it.

PR #1255, opened 2022-08-29
Chore: port from PR Tendermint v0.35.7

Karrenbelt:
## Proposed changes

Porting a few changes from PR #1012 since the upgrade to v0.35.7 is discontinued
- fix: `TestTendermintServerApp.test_files_exist`
- - fix: `ABCIHandler.apply_snapshot_chunk` tuple instead of list to `refetch_chunks` and `reject_senders`
- tests: check all proto files before raising on any
- tests: `_TendermintABCISerializer.encode_varint`
- docs: abci protocol references
- feat: `EncodeVarintError` on out-of-bounds uint64

~~TODO:~~ 
- `ConsensusParams` import

This is not an issue here, only became one after moving to v0.35.7
 [v0.34.19/proto/tendermint/abci/types.proto](https://github.com/tendermint/tendermint/blob/v0.34.19/proto/tendermint/abci/types.proto#L284) --- VS --- [v0.35.7/proto/tendermint/types/params.proto](https://github.com/tendermint/tendermint/blob/v0.35.7/proto/tendermint/types/params.proto#L13)
[example use in `RequestInitChain` v0.34.19](https://github.com/tendermint/tendermint/blob/2f231ceb952a2426cf3c0abaf0b455aadd11e5b2/proto/tendermint/abci/types.proto#L64) --- VS --- [example use in `RequestInitChain` v0.35.7](https://github.com/tendermint/tendermint/blob/9daea43375e06816a44c30bdb7b02701df4a5f76/proto/tendermint/abci/types.proto#L57)

## Fixes

If it fixes a bug or resolves a feature request, be sure to link to that issue.

## Types of changes

What types of changes does your code introduce? (A **breaking change** is a fix or feature that would cause existing functionality and APIs to not work as expected.)
_Put an `x` in the box that applies_

- [ ] Non-breaking fix (non-breaking change which fixes an issue)
- [ ] Breaking fix (breaking change which fixes an issue)
- [ ] Non-breaking feature (non-breaking change which adds functionality)
- [ ] Breaking feature (breaking change which adds functionality)
- [ ] Refactor (non-breaking change which changes implementation)
- [ ] Messy (mixture of the above - requires explanation!)

## Checklist

_Put an `x` in the boxes that apply._

- [ ] I have read the [CONTRIBUTING](../blob/main/CONTRIBUTING.md) doc
- [ ] I am making a pull request against the `main` branch (left side). Also you should start your branch off our `main`.
- [ ] Lint and unit tests pass locally with my changes
[...]

PR #1218, opened 2022-08-23
Fix Dockerfiles and docker images

angrybayblade:
## Proposed changes

This PR fixes several docker image related issues

- [x] Remove support for dependency image builds from the framework 
- [ ] Separate agent image into base autonomy image and runtime image

## Types of changes

What types of changes does your code introduce? (A **breaking change** is a fix or feature that would cause existing functionality and APIs to not work as expected.)
_Put an `x` in the box that applies_

- [ ] Non-breaking fix (non-breaking change which fixes an issue)
- [x] Breaking fix (breaking change which fixes an issue)
- [ ] Non-breaking feature (non-breaking change which adds functionality)
- [ ] Breaking feature (breaking change which adds functionality)
- [ ] Refactor (non-breaking change which changes implementation)
- [ ] Messy (mixture of the above - requires explanation!)

## Checklist

_Put an `x` in the boxes that apply._

- [x] I have read the [CONTRIBUTING](../blob/main/CONTRIBUTING.md) doc
- [x] I am making a pull request against the `main` branch (left side). Also you should start your branch off our `main`.
- [ ] Lint and unit tests pass locally with my changes
- [ ] I have added tests that prove my fix is effective or that my feature works
- [ ] I have locally run services that could be impacted and they do not present failures derived from my changes

PR #1265, opened 2022-09-01
Fix Tendermint node's logs

Adamantios:
## Proposed changes

I had to study the logs from the nodes' containers but they were filled with a repetitive error message. Therefore, this PR adds a check that was missing so that the logs are useful again.

The logs were reporting the following message:

```
'NoneType' object has no attribute 'readline'Error!: 'NoneType' object has no attribute 'readline'Error!
```

## Fixes

n/a

## Types of changes

## Checklist

## Further comments

In order to confirm that this works, it is necessary to re-build the tendermint image locally before running an e2e test. This can be done with the following command from the root:

```bash
docker build ./deployments/Dockerfiles/tendermint/ -t valory/open-autonomy-tendermint:latest
```


plugins/aea-test-autonomy/aea_test_autonomy/docker/tendermint.py
```diff
@@ -134,28 +133,6 @@ def __init__(  # pylint: disable=too-many-arguments
...
-            cmd = [
-                "autonomy",
-                "deploy",
-                "build",
-                "image",
-                "valory/oracle_hardhat",
-                "--dependencies",
-            ]
-            subprocess.run(cmd)  # nosec   # pylint: disable=subprocess-run-check
-            os.chdir(cwd)
```
Adamantios, line 158: Removed since we now do not run the tests with local images' versions.

Karrenbelt (commented):
Looks good, but CI errors out on `check-doc-links-hashes`

packages/valory/connections/abci/connection.py
```diff
@@ -1236,18 +1236,21 @@ def check_server_status(
...
-                    "RPC HTTP server stopped",  # this occurs when we lose connection from the tm side
-                    "Stopping abci.socketClient for error: read message: EOF module=abci-client connection=",  # this occurs when we lose connection from the AEA side.
-                ]:
-                    if line.find(trigger) >= 0:
-                        self._stop_tm_process()
-                        self._start_tm_process()
-                        self.write_line(
-                            f"Restarted the HTTP RPC server, as a connection was dropped with message:\n\t\t {line}\n"
-                        )
+                if self._process is not None and self._process.stdout is not None:
```
DavidMinarsch, line 1239: How is this suddenly an issue - makes me feel uneasy. We didn't change this recently..

Also, how do we recover from either of them being `None` with this logic?

I suspect it might be related to https://github.com/valory-xyz/open-autonomy/pull/1255 and the change this introduced to abci protocol

plugins/aea-test-autonomy/aea_test_autonomy/docker/tendermint.py
DavidMinarsch: @angrybayblade safe to remove?

deployments/Dockerfiles/tendermint/tendermint.py
```diff
@@ -245,18 +245,21 @@ def check_server_status(
...
-                    "RPC HTTP server stopped",  # this occurs when we lose connection from the tm side
-                    "Stopping abci.socketClient for error: read message: EOF module=abci-client connection=",  # this occurs when we lose connection from the AEA side.
-                ]:
-                    if line.find(trigger) >= 0:
-                        self._stop_tm_process()
-                        self._start_tm_process()
-                        self.write_line(
-                            f"Restarted the HTTP RPC server, as a connection was dropped with message:\n\t\t {line}\n"
-                        )
+                if self._process is not None and self._process.stdout is not None:
```
DavidMinarsch, line 248: see below

plugins/aea-test-autonomy/aea_test_autonomy/docker/tendermint.py
Adamantios: It is safe because we now pull from the hub if the tag is not found locally. Either way, the command had not been updated and does not work anymore (`--dependencies` does not exist). It would fail and since there is no check for the return code, we would never know.

packages/valory/connections/abci/connection.py
Adamantios: I checked the GH history and this started occurring after #1218.

This specific part of the code is just monitoring so I am not sure if we should do something to recover.

DavidMinarsch (approved):

packages/valory/connections/abci/connection.py
0xArdi: It seems the problem was introduced here: 
https://github.com/valory-xyz/open-autonomy/pull/1218/commits/48e138aaef4da999631be50f0807a29e2b1ce05a

0xArdi: I will make a couple of draft PR to showcase this, and make sure that is the case and something funny is not going on my local.

0xArdi: #1268 is at the last commit without the error.
#1269 is at the commit that I think introduced the problem.

0xArdi: So it is not possible to run at the commit that #1268, without making changes that might effect it. So I closed it.

Regardless, it seems changing from image tag `0.1.0` -> 'latest' seemed to have caused the problem.
In #1269 there are 2 successful runs:
- One with tag "latest" and the endline [error present](https://github.com/valory-xyz/open-autonomy/runs/8155945324?check_suite_focus=true).
- One with tag "0.1.0" and the endline [error not present](https://github.com/valory-xyz/open-autonomy/runs/8155945312?check_suite_focus=true). 

Not sure what the difference between this images is. Any idea @angrybayblade ?

PR #1270, opened 2022-09-02
fix/flaky registries tests

0xArdi:
## Fixes
The test at `tests/test_autonomy/test_cli/test_deploy/test_from_token.py::TestFromToken::test_from_token` is flaky.
Ex: https://github.com/valory-xyz/open-autonomy/runs/8112005341?check_suite_focus=true
## Types of changes

## Checklist

## Further comments

Judging from the logs on the [prev CI runs](https://github.com/valory-xyz/open-autonomy/runs/8112005341?check_suite_focus=true) it seems that the container is running correctly. However, it seems the test decides to kill it. 
The number of retries is quite close to the max (10).


.github/workflows/main_workflow.yml
```diff
@@ -155,7 +155,6 @@ jobs:
     continue-on-error: True
     needs:
       - lock_check
-      - copyright_doc_and_pipfile_check
```
0xArdi, line 158: removing this temporarily in order to see the results

Adamantios: You may cherry pick 218f1ebd4898e774611534702c0c5595f76affdc instead.

0xArdi:
After taking another look at it the problem is that the gateway sometimes responds with 504, ex: 
```
2022-09-02 16:38:51 [   DEBUG] https://gateway.autonolas.tech:443 "GET /ipfs/f01701220d913b5bf68193dfacb941538d5900466c449c9ec8121153f152de2e026fa7f3a HTTP/1.1" 504 160 (connectionpool.py:456)
```

autonomy/deploy/chain.py
```diff
@@ -104,6 +104,11 @@ def resolve_token_id(
     ) -> Dict:
         """Resolve token id using on-chain contracts."""
         url = self.contract.functions.tokenURI(token_id).call()
+        return self._resolve_from_ipfs(url)
+
+    @staticmethod
+    def _resolve_from_ipfs(url: str) -> Dict:
+        """Resolves from ipfs given an URL."""
```
0xArdi, line 111: for easier mocking

tests/test_autonomy/test_cli/test_deploy/test_from_token.py
```diff
@@ -62,6 +67,9 @@ def test_from_token(
         ), mock.patch(
             "autonomy.cli.fetch.get_ipfs_node_multiaddr",
             new=lambda: "/dns/registry.autonolas.tech/tcp/443/https",
+        ), mock.patch(
+            "autonomy.deploy.chain.ServiceRegistry._resolve_from_ipfs",
+            return_value=mock_ipfs_response,
```
0xArdi, line 72: we mock the ipfs http request. I don't see a point why this should be using a real gateway, just introduces an external dependency.

tests/test_autonomy/test_cli/test_deploy/test_from_token.py
```diff
@@ -52,7 +51,13 @@ def test_from_token(
...
     ) -> None:
         """Run test."""
-
+        mock_ipfs_response = {
+            "name": "valory/oracle_hardhat",
+            "description": "Oracle service.",
+            "code_uri": "ipfs://bafybeiansmhkoovd6jlnyurm2w4qzhpmi43gxlyenq33ioovy2rh4gziji",
+            "image": "bafybeiansmhkoovd6jlnyurm2w4qzhpmi43gxlyenq33ioovy2rh4gziji",
+            "attributes": [{"trait_type": "version", "value": "0.1.0"}],
+        }
```
0xArdi, line 60: this is the response from an actual request the command makes

0xArdi:
This is from a [previous run](https://github.com/valory-xyz/open-autonomy/runs/8159601598)  that is not visible now because to the force push.

.github/workflows/main_workflow.yml
0xArdi: This is now resolved by cherry picking. Thanks @Adamantios . 
Force pushed to not pollute the repo with unnecessary commits and their reverts.

tests/test_autonomy/test_cli/test_deploy/test_from_token.py
DavidMinarsch: The problem is this will end up outdated

angrybayblade: >The problem is this will end up outdated

In a way this is not a problem, since we just want to check if the on-chain token resolution is working or not.  I think we can introduce integration tests where we actually interact with external services and keep this as an unit test.

0xArdi: I just realized that there is a ipfs download going on after this ipfs response is returned. I assume you mean that it will get outdated because we download from ipfs with that specific hash. 
What if we just mock the download to copy from the local service package instead of actually trying to download form ipfs?

DavidMinarsch: This is an e2e test, so we should use the external dependency imo. The mocking will just get out of sync and be hard to maintain.

0xArdi: Okay, so we have to "fixes" in this PR.
1. Increased max retries.
2. Mocking of ipfs req.

I can revert 2, but the test will remain flaky.

DavidMinarsch: Yes let's revert 2 for now and instead increase retries (done) and also add flaky test marker so it reruns the test on failure

DavidMinarsch (approved):

Issue #1021, opened 2022-06-26
Hardcoded values in contracts


Karrenbelt:
not clear what these values represent; `MIN_GAS`, `MIN_GASPRICE` and ...?

https://github.com/valory-xyz/open-autonomy/blob/285b699bca07bd9ac21bf26ab7d43ac5d76f6abb/packages/valory/contracts/gnosis_safe/contract.py#L212
https://github.com/valory-xyz/open-autonomy/blob/285b699bca07bd9ac21bf26ab7d43ac5d76f6abb/packages/valory/contracts/gnosis_safe/contract.py#L430
https://github.com/valory-xyz/open-autonomy/blob/285b699bca07bd9ac21bf26ab7d43ac5d76f6abb/packages/valory/contracts/gnosis_safe/contract.py#L456
https://github.com/valory-xyz/open-autonomy/blob/285b699bca07bd9ac21bf26ab7d43ac5d76f6abb/packages/valory/contracts/gnosis_safe_proxy_factory/contract.py#L147
https://github.com/valory-xyz/open-autonomy/blob/495561dd996aeb1149b06083ff135c4c3c6bd4ea/packages/valory/contracts/multisend/contract.py#L150

PR #1271, opened 2022-09-02
Use local registry for registration tests

Adamantios:
## Proposed changes

Use the local registry for registration tests and stop using the staging chain.

## Fixes

n/a

## Types of changes

## Checklist

## Further comments

Blocked by https://github.com/valory-xyz/autonolas-registries/pull/47.


packages/valory/contracts/service_registry/contract.py
```diff
@@ -31,8 +31,8 @@
 
 # test_contract and test_agents, respectively...
 DEPLOYED_BYTECODE_MD5_HASH = (
-    "d518af84b4ce10c7611ba3ad749cb3907d225a8a66061bd871f490800c57ea41"
-    "186febbada22648882cca80541b87fc04fa5051768f027aeccd467ac97c9554c"
+    "d8084598f884509694ab1f244cbb8e7697d8db00c241710b89b2ec3037d2edd3"
+    "b82d01f1f0ca6bd9b265b1184d9c563a6000c30958c2a7ae5a9c35e5ff2ba7de"
 )
```
Adamantios, line 36: This will be parametrized when I address #1021.

plugins/aea-test-autonomy/aea_test_autonomy/docker/registries.py
```diff
@@ -34,7 +34,6 @@
 DEFAULT_HARDHAT_ADDR = "http://127.0.0.1"
 DEFAULT_HARDHAT_PORT = 8545
 REGISTRIES_CONTRACTS_DIR = "autonolas-registries"
-CONFIG_FILE = "hardhat.config.ts"
```
Adamantios, line 37: Was not used. Also, the constants below are currently not used but might be useful in the future.

scripts/check_ipfs_hashes_pushed.py
```diff
@@ -32,7 +32,7 @@
 def check_ipfs_hash_pushed(ipfs_hash: str) -> bool:
     """Check that the given ipfs hash exists in the registry"""
     try:
-        url = f"{IPFS_ENDPOINT}/{ipfs_hash}"
+        url = f"{IPFS_ENDPOINT}/{ipfs_hash.strip()}"
```
Adamantios, line 35: Resolves issue with `check_ipfs_hashes_pushed`:
https://github.com/valory-xyz/open-autonomy/runs/8151526695?check_suite_focus=true

packages/valory/contracts/service_registry/contract.py
DavidMinarsch: It should not be.

DavidMinarsch (dismissed):

packages/valory/contracts/service_registry/contract.py
Adamantios: If not, then the real skill's bytecode and the e2e tests' bytecode would be different if the local registries image does not get updated but the `ServiceRegistry` contract does.

DavidMinarsch: It needs to be hardcoded. See what I did. We want to be in full control of the service registries being supported.

packages/valory/contracts/service_registry/contract.py
```diff
@@ -82,15 +84,23 @@ def verify_contract(
...
-        verified = DEPLOYED_BYTECODE_MD5_HASH == sha512_hash
+        verified = DEPLOYED_BYTECODE_MD5_HASH_BY_CHAIN_ID[chain_id] == sha512_hash
         if not verified:
-            log_msg = "CONTRACT NOT VERIFIED! reason: frequent changes."
-            log_msg += f". Verified: {verified}. Contract address: {contract_address}."
-            _logger.error(f"{log_msg} Address: {CHAIN_ADDRESS}, chain_id: {CHAIN_ID}")
-        return dict(verified=verified, sha512_hash=sha512_hash)
+            _logger.error(
+                f"CONTRACT NOT VERIFIED! Contract address: {contract_address}, chain_id: {chain_id}."
+            )
```
Karrenbelt, line 102: we're no longer return `True` by default as we did when the contract bytecode was continuously changing, we may actually return `False` as well, so this log message is incorrect.

Adamantios: I think you missed the `if not verified` condition above.

Karrenbelt: you are right. Lets change the message to be more explicit, I used this to indicate that verification was not performed, in this case it has been performed but failed. e.g. `BYTECODE VERIFICATION FAILED` or something like that

Adamantios: Opened #1280 because this is out of the scope of this PR.

Karrenbelt (approved):

DavidMinarsch (approved):

PR #1277, opened 2022-09-05
Fix flaky `test_fetch_behaviour_non_indexed_block` test

Adamantios:
## Proposed changes

Fixes flaky `test_fetch_behaviour_non_indexed_block` test.

Example failing run: [Use local registry for registration tests main_workflow #4973](https://github.com/valory-xyz/open-autonomy/runs/8185034106?check_suite_focus=true)

## Fixes

n/a

## Types of changes

## Checklist

## Further comments

n/a

### DIFF
--- a/tests/test_skills/test_apy_estimation_abci/test_behaviours.py
+++ b/tests/test_skills/test_apy_estimation_abci/test_behaviours.py
@@ -818,17 +818,15 @@ class TestFetchAndBatchBehaviours(APYEstimationFSMBehaviourBaseCase):
         else:
             res = {"errors": [{"message": "another error"}]}
 
-        response_kwargs["body"] = json.dumps(res).encode("utf-8")
-        behaviour.act_wrapper()
-        self.mock_http_request(request_kwargs, response_kwargs)
+        with caplog.at_level(
+            logging.WARNING,
+            logger="aea.test_agent_name.packages.valory.skills.apy_estimation_abci",
+        ):
+            response_kwargs["body"] = json.dumps(res).encode("utf-8")
+            behaviour.act_wrapper()
+            self.mock_http_request(request_kwargs, response_kwargs)
 
         if not is_non_indexed_res:
-            with caplog.at_level(
-                logging.WARNING,
-                logger="aea.test_agent_name.packages.valory.skills.apy_estimation_abci",
-            ):
-                behaviour.act_wrapper()
-
             assert (
                 "Attempted to handle an indexing error, but could not extract the latest indexed block!"
                 in caplog.text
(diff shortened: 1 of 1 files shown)


tests/test_skills/test_apy_estimation_abci/test_behaviours.py
```diff
@@ -818,17 +818,15 @@ def test_fetch_behaviour_non_indexed_block(
         else:
             res = {"errors": [{"message": "another error"}]}
 
-        response_kwargs["body"] = json.dumps(res).encode("utf-8")
-        behaviour.act_wrapper()
-        self.mock_http_request(request_kwargs, response_kwargs)
```
Adamantios, line 823: The act which was triggering the logging was being called out of the `caplog`'s context manager.

Karrenbelt (approved):
ok I'm trying to understand this, bit more guidance would be helpful. 

- failing (flaky) case is when `is_non_indexed_res` is `False`
- call of `behaviour.act_wrapper()` happened once before and outside of the caplog context manager, AND once after and inside the caplog context manager. Is this why it was flaky instead of failing consistently?

https://github.com/valory-xyz/open-autonomy/blob/fb4897c9eac47d14d574e2144993ffb21bf68cdc/tests/test_skills/test_apy_estimation_abci/test_behaviours.py#L800-L802

https://github.com/valory-xyz/open-autonomy/blob/fb4897c9eac47d14d574e2144993ffb21bf68cdc/tests/test_skills/test_apy_estimation_abci/test_behaviours.py#L825-L830

Adamantios:
> ok I'm trying to understand this, bit more guidance would be helpful.
> 
> * failing (flaky) case is when `is_non_indexed_res` is `False`
> * call of `behaviour.act_wrapper()` happened once before and outside of the caplog context manager, AND once after and inside the caplog context manager. Is this why it was flaky instead of failing consistently?
> 
> https://github.com/valory-xyz/open-autonomy/blob/fb4897c9eac47d14d574e2144993ffb21bf68cdc/tests/test_skills/test_apy_estimation_abci/test_behaviours.py#L800-L802
> 
> https://github.com/valory-xyz/open-autonomy/blob/fb4897c9eac47d14d574e2144993ffb21bf68cdc/tests/test_skills/test_apy_estimation_abci/test_behaviours.py#L825-L830

It is hard to answer why it was not failing consistently. The important part is that the act which was triggering the logging was being called out of the `caplog`'s context manager and this is the reason why it was failing.

tests/test_skills/test_apy_estimation_abci/test_behaviours.py
Adamantios: Yes, it is intentional. It was incorrect since I initially misplaced it, thinking that this was the call that was also triggering the logging.

DavidMinarsch (approved):

PR #1278, opened 2022-09-05
Remove hardcoded values

### LINKED ISSUE #1021: Hardcoded values in contracts

Adamantios:
## Proposed changes

Replaces magic values with constants with meaningful descriptions.

## Fixes

Closes #1021.

## Types of changes

## Checklist

## Further comments

n/a


Karrenbelt (commented):
looks good, there is one more case I don't like and can be part of this same PR imo

https://github.com/valory-xyz/open-autonomy/blob/ebd5f05c62275a420f5b3184c9ad4a447c89e325/packages/valory/contracts/gnosis_safe_proxy_factory/contract.py#L141-L143

instead, we should provide `1` as a default value instead of `0`, here:
https://github.com/valory-xyz/open-autonomy/blob/ebd5f05c62275a420f5b3184c9ad4a447c89e325/packages/valory/contracts/gnosis_safe_proxy_factory/contract.py#L92

Adamantios:
> looks good, there is one more case I don't like and can be part of this same PR imo
> 
> https://github.com/valory-xyz/open-autonomy/blob/ebd5f05c62275a420f5b3184c9ad4a447c89e325/packages/valory/contracts/gnosis_safe_proxy_factory/contract.py#L141-L143
> 
> instead, we should provide `1` as a default value instead of `0`, here:
> 
> https://github.com/valory-xyz/open-autonomy/blob/ebd5f05c62275a420f5b3184c9ad4a447c89e325/packages/valory/contracts/gnosis_safe_proxy_factory/contract.py#L92

Good catch!
What if the user provides `0` as an input though? Should we do something like:
```python
# we set a value to avoid triggering the gas estimation during buildTransaction below
tx_parameters["gas"] = Wei(max(gas, 1))
```
?

Karrenbelt:
> > looks good, there is one more case I don't like and can be part of this same PR imo
> > https://github.com/valory-xyz/open-autonomy/blob/ebd5f05c62275a420f5b3184c9ad4a447c89e325/packages/valory/contracts/gnosis_safe_proxy_factory/contract.py#L141-L143
> > 
> > instead, we should provide `1` as a default value instead of `0`, here:
> > https://github.com/valory-xyz/open-autonomy/blob/ebd5f05c62275a420f5b3184c9ad4a447c89e325/packages/valory/contracts/gnosis_safe_proxy_factory/contract.py#L92
> 
> Nice catch! What if the user provides `0` as an input though? Should we do something like:
> 
> ```python
> # we set a value to avoid triggering the gas estimation during buildTransaction below
> tx_parameters["gas"] = Wei(max(gas, 1))
> ```
> 
> ?

hmm I would say no. I don't think we should overwrite to 1 if the input is explicitly provided to be zero. 

I should, however, first ask why do we want to prevent gas estimation during `buildTransaction`? If there is a solid reason to avoid this then we should. Leave a comment in the code why it's necessary or important to avoid the estimation prior to `buildTransaction` if so and call it done.

0xArdi:
@Karrenbelt when a gas estimate is not set when calling `buildTransaction`, by default gas estimation is performed. That throws an error when repricing a tx. More info on the comments in https://github.com/valory-xyz/open-autonomy/pull/1212 .

Karrenbelt:
> @Karrenbelt when a gas estimate is not set when calling `buildTransaction`, by default gas estimation is performed. That throws an error when repricing a tx. More info on the comments in #1212 .

so that's good to know, @Adamantios worth adding as comment to the code?

DavidMinarsch (approved):

DavidMinarsch (approved):

PR #1284, opened 2022-09-06
Fix constants in contracts

Adamantios:
## Proposed changes

At first, we created a `constants.py` file to store the constants in 210512d41b38546d18f3e38261e122a0058389d5, but [some tests were failing](https://github.com/valory-xyz/open-autonomy/runs/8189132511?check_suite_focus=true), because the `aea` framework was throwing an unhandled error [here](https://github.com/valory-xyz/open-aea/blob/main/aea/components/utils.py#L48):
```
KeyError: '__path__'
```

Therefore, we moved the constants to the `__init__.py` file of the contracts in ebd5f05c62275a420f5b3184c9ad4a447c89e325. However, this was now [failing at the e2e tests](https://github.com/valory-xyz/open-autonomy/runs/8190948023?check_suite_focus=true) when the `aea` framework was trying to load the packages [here](https://github.com/valory-xyz/open-aea/blob/main/aea/components/base.py#L315-L382):
```
ImportError: cannot import name 'GAS_ESTIMATE_ADJUSTMENT' from 'packages.valory.contracts' (unknown location)
```

In order to overcome both limitations, we now move the constants into the contract files, which unavoidably creates some duplication.

## Fixes

n/a

## Types of changes

## Checklist

## Further comments

There might be some room for improvements on `open-aea`, or clarifications in its documentation.


DavidMinarsch (approved):
LGTM - let's wait for e2e to finish

PR #1295, opened 2022-09-07
Test retry strategy when the node is blocking

Adamantios:
## Proposed changes

Create a test whose purpose is to check whether the retry strategy works if a node is blocking, by simulating a blocking time larger than the expected duration of the strategy.

## Fixes

n/a

## Types of changes

## Checklist

## Further comments

n/a


Adamantios:
We may investigate if we can replace the [get_transaction_receipt](https://web3py.readthedocs.io/en/stable/web3.eth.html#web3.eth.Eth.get_transaction_receipt) with the [wait_for_transaction_receipt](https://web3py.readthedocs.io/en/stable/web3.eth.html#web3.eth.Eth.wait_for_transaction_receipt) on `open-aea`, because it has a timeout and is potentially not affected by blocking node requests.

0xArdi (approved):

Adamantios:
@0xArdi and I believe that there are two potential solutions:

1. Replace the [get_transaction_receipt](https://web3py.readthedocs.io/en/stable/web3.eth.html#web3.eth.Eth.get_transaction_receipt) with the [wait_for_transaction_receipt](https://web3py.readthedocs.io/en/stable/web3.eth.html#web3.eth.Eth.wait_for_transaction_receipt) on `open-aea`, because it has a timeout and is not affected by blocking node requests. This would also allow us to replace the custom retry logic in our ledger dispatcher, but we would only be able to have a constant backoff.
2. Create our own `wait_for_transaction_receipt` method at the `open-autonomy` so that we do not lose our linear backoff.

(1.) is the cleanest solution, but has the limitation that we would need to wait for an `open-aea` release before fixing the issue.

0xArdi (approved):

DavidMinarsch:
Option 1 please.

Adamantios:
Marked as a draft until we resolve the issue on `open-aea` and the fix is included in a release.

Adamantios:
Flaky `test_attempts_get_transaction_receipt` in [[WIP] Test retry strategy when the node is blocking main_workflow #5132](https://github.com/valory-xyz/open-autonomy/runs/8270584723?check_suite_focus=true) is because of a too small value for the `retry_timeout`, which as a result makes the `get_transaction_receipt` to timeout before the mocked method gets called and therefore the count does not increase.

We can increase the timeout at the base branch on #1311 after this gets merged.

0xArdi (approved):
noice

PR #1307, opened 2022-09-08
chore: make localhost the default staging chain

0xArdi:
## Proposed changes

Replace the default staging chain with the default hardhat/ganache address. 

## Types of changes

## Checklist

### DIFF
--- a/autonomy/deploy/chain.py
+++ b/autonomy/deploy/chain.py
@@ -33,7 +33,7 @@ ABIS_DIR = "abis"
 SERVICE_REGISTRY_ABI = (
     DATA_DIR / ABIS_DIR / "service_registry" / "service_registry.json"
 )
-DEFAULT_STAGING_CHAIN = "https://chain.staging.autonolas.tech"
+DEFAULT_STAGING_CHAIN = "http://localhost:8545"
 
 CHAIN_CONFIG: Dict[str, Dict[str, Optional[str]]] = {
     "staging": {


autonomy/deploy/chain.py
```diff
@@ -33,7 +33,7 @@
 SERVICE_REGISTRY_ABI = (
     DATA_DIR / ABIS_DIR / "service_registry" / "service_registry.json"
 )
-DEFAULT_STAGING_CHAIN = "https://chain.staging.autonolas.tech"
+DEFAULT_STAGING_CHAIN = "http://localhost:8545"
```
angrybayblade, line 36: Can you also update the relevant docs on how to use local registry deployment or this won't work. The script to run the local registry deployment is `scripts/run_service_registry_locally.py`

0xArdi: Is there a way to run this script from `autonomy` ? Otherwise we make the users life way too difficult IMO. They would have to go and get the script from the open-autonomy github.

0xArdi: Alternatively, we can make the default chain to be goerli. And give an explanation on how to run the localnet.

DavidMinarsch: Let;s open an Trello ticket for making the script above part of autonomy.

DavidMinarsch (approved):

PR #1308, opened 2022-09-08
Override APY deposit endpoint at the service level

Adamantios:
## Proposed changes

Add environment variable override for the `/apy/deposit` endpoint of the backend at the APY service.

## Fixes

n/a

## Types of changes

## Checklist

## Further comments

n/a


autonomy/deploy/generators/docker_compose/templates.py
```diff
@@ -90,6 +90,7 @@
     container_name: abci{node_id}
     image: {runtime_image}
     environment:
+      - PYTHONHASHSEED=0
```
Adamantios, line 93: We need to disable hash randomization in order to have reproducible ML results among the agents.

DavidMinarsch (approved):

PR #1316, opened 2022-09-09
[WIP] Enable package tests on CI

angrybayblade:
## Proposed changes

This PR enables CI runs for ported package test

## Types of changes

## Checklist


tox.ini
```diff
@@ -75,6 +83,19 @@ deps =
     atheris==2.0.11
 setenv =
     {[testenv]setenv}
+commands = {[framework]commands}
+
+[testenv:py3.10-linux-packages]
```
DavidMinarsch, line 88: for consistency with `open-aea` please make `packages-` the prefix rather than suffix

tests/conftest.py
```diff
@@ -71,6 +71,9 @@
     DEFAULT_TENDERMINT_PORT,
     FlaskTendermintDockerImage,
 )
+from aea_test_autonomy.fixture_helpers import (  # noqa: F401  # pylint: disable=unused-import
+    tendermint,
+)
```
Adamantios, line 76: Why? I think this is not an auto-used fixture, right?

DavidMinarsch (approved):

PR #1323, opened 2022-09-09
[0.3.1] Make Tendermint parameters configurable

### LINKED ISSUE #998: Make Tendermint parameters configurable
Tendermint parameters should not be hardcoded and should be configurable from the yaml file.

https://github.com/valory-xyz/open-autonomy/pull/680#discussion_r903266641

Adamantios:
## Proposed changes

Makes the Tendermint parameters configurable.

## Fixes

Closes #998.

## Types of changes

## Checklist

## Further comments

n/a


packages/valory/skills/registration_abci/behaviours.py
```diff
@@ -332,11 +289,33 @@ def request_tendermint_info(self) -> Generator[None, None, bool]:
...
+
+        validators = []
+        for i, validator_config in enumerate(collected_agent_info.values()):
+            validator = dict(
+                address=validator_config["address"],
+                pub_key=validator_config["pub_key"],
+                power=self.params.voting_power,
+                name=f"node{i}",
+            )
+            validators.append(validator)
```
Karrenbelt, line 306: We should use a mapping from agent to node name instead of relying on being able to enumerate what is assumed here to be ordered. This may not be an issue for communication between the nodes, but I can image that running containers named node0-3 might no longer correspond which may lead to confusion on the developers side as well.

Karrenbelt (approved):
LGTM

Karrenbelt (changes requested):
meh errors `aea.exceptions.AEAEnforceError: 'genesis_config' required, but it is not set` :(

packages/valory/skills/registration_abci/behaviours.py
```diff
@@ -332,11 +289,33 @@ def request_tendermint_info(self) -> Generator[None, None, bool]:
...
+    ) -> Dict[str, Any]:
+        """Format collected agent info for genesis update"""
+
+        validators = []
+        for i, validator_config in enumerate(collected_agent_info.values()):
+            validator = dict(
+                address=validator_config["address"],
+                pub_key=validator_config["pub_key"],
+                power=self.params.voting_power,
+                name=f"node{i}",
```
angrybayblade, line 304: make this template a constant

packages/valory/skills/registration_abci/behaviours.py
```diff
@@ -332,11 +289,33 @@ def request_tendermint_info(self) -> Generator[None, None, bool]:
...
+        collected_agent_info: Dict[str, Any],
+    ) -> Dict[str, Any]:
+        """Format collected agent info for genesis update"""
+
+        validators = []
+        for i, validator_config in enumerate(collected_agent_info.values()):
+            validator = dict(
+                address=validator_config["address"],
+                pub_key=validator_config["pub_key"],
+                power=self.params.voting_power,
```
angrybayblade, line 303: Shouldn't this be the voting power of the validator node, eg something like `power=validator_config["power"]`

packages/valory/skills/abstract_round_abci/models.py
```diff
@@ -50,6 +50,8 @@ class BaseParams(Model):  # pylint: disable=too-many-instance-attributes
 
     def __init__(self, *args: Any, **kwargs: Any) -> None:
         """Initialize the parameters object."""
+        self.genesis_config = self._ensure("genesis_config", kwargs)
```
angrybayblade, line 53: Can you a comment on how should the genesis config look like

packages/valory/skills/registration_abci/behaviours.py
```diff
@@ -332,11 +289,33 @@ def request_tendermint_info(self) -> Generator[None, None, bool]:
...
+            genesis_config=self.params.genesis_config,
+        )
+        return genesis_data
+
     def request_update(self) -> Generator[None, None, bool]:
         """Make HTTP POST request to update agent's local Tendermint node"""
 
         url = self.tendermint_parameter_url
-        genesis_data = format_genesis_data(self.registered_addresses)
+        genesis_data = self.format_genesis_data(self.registered_addresses)
```
angrybayblade, line 318: Is tendermint com server up to date to work with this?

packages/valory/skills/registration_abci/behaviours.py
Adamantios: @Karrenbelt could you explain this choice?

packages/valory/skills/registration_abci/behaviours.py
Adamantios: https://github.com/valory-xyz/open-autonomy/blob/673761b24807845d04dda18e1e15dbaca8150a86/deployments/Dockerfiles/tendermint/app.py#L163-L193

packages/valory/skills/registration_abci/behaviours.py
Karrenbelt: so this was "DEFAULT_VOTING_POWER" when I wrote it, for a reason. That is because the validator config and public key are requested and shared among agents, but we cannot rely on parties reporting their own voting power honestly. The default voting power on Tendermint is `10`, hence we don't want to set it to `1` either and somehow risk those not updating to have more voting power, although technically they cannot (should not be able to) connect to the other nodes.

Either way, for out application it would be good to have voting power set to default all equal (for now) and what we could also consider is to have the agents reach consensus on the tendermint configuration. Quite frankly I don't know what happens if there are inconsistencies in e.g. voting power that different agents have assigned to other agents in their tendermint config. Anyone else?

Adamantios: Yes, that makes sense. Therefore I'd suggest that I remove it from the config files and hardcode it again.

Karrenbelt: the params we have set, are part of the IPFS hash as well right, so if you specify voting power in params or in the file as a constant should make no difference

packages/valory/skills/apy_estimation_abci/skill.yaml
```diff
@@ -168,6 +184,7 @@ models:
       tendermint_url: http://localhost:26657
       testing:
         steps_forward: 1
+      voting_power: '10'
```
Karrenbelt, line 187: why not nested under `genesis_config`?

Karrenbelt (commented):
What I would really want to see is a test that shows that we can in fact update the tendermint configuration, and restart the tendermint nodes with the altered configuration.

Opened an issue for this: #1364

packages/valory/skills/abstract_round_abci/models.py
```diff
@@ -67,12 +67,13 @@ def __init__(self, *args: Any, **kwargs: Any) -> None:
...
                 pub_key_types: List[str]
               version: dict
+            voting_power: str
 
         :param args: positional arguments
         :param kwargs: keyword arguments
         """
         self.genesis_config = self._ensure("genesis_config", kwargs)
-        self.voting_power = self._ensure("voting_power", kwargs)
+        self.voting_power = self._ensure("voting_power", self.genesis_config)
```
Karrenbelt, line 76: this should be named `genesis_config` then write, instead of `voting_power`, or what's the reason not to?

Adamantios: This is because of the way the `voting_power` is used and the `genesis_data` have been defined here:

https://github.com/valory-xyz/open-autonomy/blob/02c0af800a4f0bb84170ddd17b5216502903d130/packages/valory/skills/registration_abci/behaviours.py#L295-L315

Therefore, I chose to use separate params for these two.

Karrenbelt: yes i would prefer something else but let's not linger on this, we anyway need to address this in the near future again when we actually try to use this

angrybayblade (dismissed):

Karrenbelt (dismissed):

DavidMinarsch (changes requested):
The scaffolding tool also needs to set these new default params. Please update here: https://github.com/valory-xyz/open-autonomy/blob/main/autonomy/cli/scaffold_fsm.py#L449-L474

DavidMinarsch (approved):
For now this works

PR #1369, opened 2022-09-21
[WIP | v0.3.1] Fix `registration_abci` tests

angrybayblade:
## Proposed changes

This PR attempts to fix the failing `registration_abci` skill test

## Types of changes

## Checklist


packages/valory/skills/registration_abci/tests/test_behaviours.py
```diff
@@ -460,6 +461,7 @@ def test_request_update(
     ) -> None:
         """Test Tendermint config update"""
 
+        self.state.updated_genesis_data = {}
```
angrybayblade, line 464: If you look at the tests 

https://github.com/valory-xyz/open-autonomy/blob/70f4889c39789ace0f0da52bea4da6634e773358/packages/valory/skills/registration_abci/tests/test_behaviours.py#L457-L505

There are basically 4 tests, `test_request_update[True, False]` and `test_request_restart[True, False]`. The `[True, False]` are parametrised as `valid_response` variable. Now every time the test ran only one of these 4 tests passed and other 3 failed depending on which test ran first. The first test passed and the rest failed because of the following block of code

https://github.com/valory-xyz/open-autonomy/blob/main/packages/valory/skills/registration_abci/behaviours.py#L385-L390

Now, since `state` variable is shared among all of the tests when first test ran it populated the `updated_genesis_data`. And since the `updated_genesis_data` is populated for the rest of the tests they never reached to the point of http request and caused the test to fail

DavidMinarsch: Shouldn’t this reset happen in “setup” of the test?

angrybayblade: I was not sure if that would effect on other tests, I'll give it a check

packages/valory/skills/liquidity_rebalancing_abci/test_tools/integration.py
```diff
@@ -49,10 +49,15 @@ class AMMIntegrationBaseCase(
 ):
     """Base test class for integration tests in a Hardhat environment, with AMM interaction."""
 
+    @classmethod
+    def _setup_class(cls, **setup_class_kwargs: Any) -> None:
+        """Continue setting up the class."""
```
Adamantios, line 54: The abstract method had to be implemented.

packages/valory/skills/liquidity_rebalancing_abci/test_tools/integration.py
```diff
@@ -49,10 +49,15 @@ class AMMIntegrationBaseCase(
...
+
     @classmethod
     def setup_class(cls, **kwargs: Any) -> None:
         """Setup."""
         super().setup_class()
+        HardHatAMMBaseTest.setup_class()
```
Adamantios, line 60: Not sure why the setup class is not called above. Something wrong with the mro.

packages/valory/skills/counter/tests/test_counter.py
```diff
@@ -68,6 +68,14 @@ def setup_class(cls, **kwargs: Any) -> None:
             AbciDialogues, cls._skill.skill_context.abci_dialogues
         )
 
+    @classmethod
+    def setup(
+        self,
+    ) -> None:
+        """Setup test."""
+
+        self.abci_counter_handler.tx_count = 0
```
angrybayblade, line 77: Since before we used to load skill for every test separately, the tests did not effect the `tx_count` for other tests. Since now we load the skill once, we need to reset the `tx_count` for every test

packages/valory/skills/counter/tests/test_counter.py
```diff
@@ -68,6 +68,14 @@ def setup_class(cls, **kwargs: Any) -> None:
             AbciDialogues, cls._skill.skill_context.abci_dialogues
         )
 
+    @classmethod
```
DavidMinarsch, line 71: Remove class method; also add `super.setup()`

packages/valory/skills/liquidity_rebalancing_abci/test_tools/integration.py
DavidMinarsch: What is it overriding on the base?

packages/valory/skills/apy_estimation_abci/tests/test_behaviours.py
```diff
@@ -204,6 +204,7 @@ def setup_class(cls, **kwargs: Any) -> None:
     def setup(self, **kwargs: Any) -> None:
         """Set up the test method."""
         super().setup()
+        assert self.behaviour.current_behaviour is not None
```
Adamantios, line 207: Added this check for linters: `self.behaviour.current_behaviour` is optional, so `None` does not have `params`. We exclude that possibility with the assertion and the linters pass.

packages/valory/skills/liquidity_rebalancing_abci/test_tools/integration.py
Adamantios: This one:
https://github.com/valory-xyz/open-autonomy/blob/d96e69bdc046a0bcb0f4abfc930036aba769bf61/plugins/aea-test-autonomy/aea_test_autonomy/docker/base.py#L260-L263

packages/valory/skills/counter/tests/test_counter.py
angrybayblade: No point in calling `super().setup()` since it basically does what `setup_class` does which is setting up the skill

packages/valory/skills/liquidity_rebalancing_abci/test_tools/integration.py
```diff
@@ -44,9 +43,7 @@
 # pylint: disable=protected-access,too-many-ancestors,unbalanced-tuple-unpacking,too-many-locals,consider-using-with,unspecified-encoding,too-many-arguments,unidiomatic-typecheck
 
 
-class AMMIntegrationBaseCase(
-    _TxHelperIntegration, _HarHatHelperIntegration, HardHatAMMBaseTest
-):
+class AMMIntegrationBaseCase(_TxHelperIntegration, _HarHatHelperIntegration):
```
DavidMinarsch, line 46: Missing the fixture?

packages/valory/skills/price_estimation_abci/tests/conftest.py
```diff
@@ -36,7 +35,7 @@
 
 
 class GnosisIntegrationBaseCase(  # pylint: disable=too-many-ancestors
```
DavidMinarsch, line 37: fixture missing?

PR #1376, opened 2022-09-22
Fix URL override

Adamantios:
## Proposed changes

Adds url override at the agent level.

## Fixes

n/a

## Types of changes

## Checklist

## Further comments

n/a


packages/valory/agents/apy_estimation/aea-config.yaml
```diff
@@ -109,6 +109,9 @@ models:
       pair_ids: ${SKILL_APY_ESTIMATION_ABCI_MODELS_PARAMS_ARGS_PAIR_IDS:dict:{"uniswap_subgraph":["0xae461ca67b15dc8dc81ce7615e0320da1a9ab8d5"],"spooky_subgraph":["0x2a651563c9d3af67ae0388a5c8f89b867038089e"]}}
       sleep_time: ${SKILL_APY_ESTIMATION_ABCI_MODELS_PARAMS_ARGS_SLEEP_TIME:int:10}
       ipfs_domain_name: ${SKILL_APY_ESTIMATION_ABCI_MODELS_PARAMS_ARGS_IPFS_DOMAIN_NAME:str:/dns/localhost/tcp/5001/http}
+  server_api:
+    args:
+      url: ${SKILL_APY_ESTIMATION_ABCI_MODELS_SERVER_API_ARGS_URL:str:http://192.168.2.17:9999/apy/deposit}
```
Adamantios, line 114: I am not sure when this got removed, I remember it existing :thinking: 

This makes me think: Should we add a test to ensure that when we are overriding something on the service level, then it exists on the agent level too?

DavidMinarsch: That should be guaranteed by the framework imo.

DavidMinarsch: I removed one of the unused service specs recently. Maybe it was accidentally deleted then or present on the other one?

DavidMinarsch (approved):

8ball030 (approved):

Issue #1168, opened 2022-08-05
APY issue during test round


Adamantios:
### Subject of the issue
The APY skill failed on [this](https://github.com/valory-xyz/open-autonomy/runs/7690311560?check_suite_focus=true) CI run, because the test task returned without any data. This could be an issue either with the training (corrupted models) or with the test methods utilized during the test task.

Adamantios:
Moved to https://github.com/valory-xyz/apy-oracle/issues/14.

PR #1383, opened 2022-09-23
Fix/ TestTendermintBufferFailing on windows

0xArdi:
## Proposed changes

~~Update the skip reason.~~
Fixes `TestTendermintBufferFailing` on windows. 

## Types of changes

## Checklist

### DIFF
--- a/tests/test_deployments/conftest.py
+++ b/tests/test_deployments/conftest.py
@@ -54,6 +54,18 @@ def rpc_port() -> int:
     return __RPC_PORT
 
 
+@pytest.fixture(scope="session")
+def max_retries() -> int:
+    """Max retries to make when ensuring Tendermint is running."""
+    return 10
+
+
+@pytest.fixture(scope="session")
+def sleep_amount() -> int:
+    """Amount to sleep in seconds."""
+    return 1
+
+
 def __port_is_open(ip: str, port: int) -> bool:
     """Assess whether a port is open"""
     sock = socket.socket(socket.AF_INET, socket.SOCK_STREAM)
@@ -64,14 +76,14 @@ def __port_is_open(ip: str, port: int) -> bool:
 
 @pytest.fixture
 def wait_for_node(
-    http_: str, loopback: str, rpc_port: int
+    http_: str, loopback: str, rpc_port: int, max_retries: int, sleep_amount: int
 ) -> Generator[None, None, None]:
     """Wait for Tendermint node to run."""
-    i, max_retries = 0, 5
+    i = 0
     while not __port_is_open(loopback, rpc_port) and i < max_retries:
         logging.debug(f"waiting for node... t={i}")
         i += 1
-        time.sleep(1)
+        time.sleep(sleep_amount)
     response = requests.get(f"{http_}{loopback}:{rpc_port}/status")
     success = response.status_code == 200
     assert success, "Tendermint node not running"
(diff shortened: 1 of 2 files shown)


tests/test_deployments/test_app.py
```diff
@@ -310,7 +310,7 @@ def setup_class(cls) -> None:
 
     @pytest.mark.skipif(
         platform.system() == "Windows",
-        reason="Needs investigation. Could be: Tendermint process is not killed on teardown in Windows.",
+        reason="The pipe gets filled before the test reaches here, and the node becomes unresponsive.",
```
0xArdi, line 313: We can make this work by having a different fixture instead of `@wait_for_node_to_run`.

DavidMinarsch: Yes please make it work.

tests/test_deployments/conftest.py
```diff
@@ -76,3 +76,17 @@ def wait_for_node(
...
     assert success, "Tendermint node not running"
     yield
+
+
+@pytest.fixture
+def wait_for_occupied_rpc_port(
+    http_: str, loopback: str, rpc_port: int
+) -> Generator[None, None, None]:
+    """Wait for Tendermint to occupy rpc_port."""
+    i, max_retries = 0, 5
```
DavidMinarsch, line 86: Make max_retries configurable please. Also 5 seems very low for CI...

tests/test_deployments/conftest.py
```diff
@@ -76,3 +76,17 @@ def wait_for_node(
...
+    while not __port_is_open(loopback, rpc_port):
+        logging.debug(f"waiting for node... t={i}")
+        i += 1
+        time.sleep(1)
```
DavidMinarsch, line 90: would make sleep time also configurable ... just makes the fixture more useful?

0xArdi: copied from fixture above, which was working until now

tests/test_deployments/conftest.py
0xArdi: copied from above

Adamantios (approved):

tests/test_deployments/conftest.py
```diff
@@ -64,15 +76,29 @@ def __port_is_open(ip: str, port: int) -> bool:
...
+    http_: str, loopback: str, rpc_port: int, max_retries: int, sleep_amount: int
+) -> Generator[None, None, None]:
+    """Wait for Tendermint to occupy rpc_port."""
+    i = 0
+    while not __port_is_open(loopback, rpc_port):
+        logging.debug(f"waiting for node... t={i}")
+        i += 1
+        time.sleep(sleep_amount)
+        assert i < max_retries, "Failed to run tendermint node in time."
+    yield
```
Adamantios, line 104: I think this can be used in `wait_for_node` to remove some duplication.

Adamantios:
The APY e2e test for 4 agents fails. Looking into the logs, we noticed that:

```
2022-09-23T12:01:31.6471666Z [102816]stdout>aea.exceptions.AEAActException: An error occurred during act of behaviour valory/apy_estimation_chained_abci:0.1.0/TestBehaviour:
2022-09-23T12:01:31.6488142Z [102816]stdout>Traceback (most recent call last):
2022-09-23T12:01:31.6500558Z [102816]stdout>
2022-09-23T12:01:31.6551699Z [102819]stdout>
2022-09-23T12:01:31.6560262Z [102819]stdout>ValueError: Cannot store multiple files of type <class 'NoneType'>!Should be a dictionary of filenames mapped to their objects.
2022-09-23T12:01:31.6583857Z [102822]stdout>
2022-09-23T12:01:31.6614494Z [102822]stdout>  File "/tmp/tmpg1i_ay3m/agent_00002/vendor/valory/skills/abstract_round_abci/io_/store.py", line 75, in store
2022-09-23T12:01:31.6630147Z [102825]stdout>  File "/tmp/tmpg1i_ay3m/agent_00003/vendor/valory/skills/abstract_round_abci/io_/store.py", line 75, in store
2022-09-23T12:01:31.6655058Z [102816]stdout>  File "/tmp/tmpg1i_ay3m/agent_00000/vendor/valory/skills/abstract_round_abci/io_/store.py", line 75, in store
```

which means that it is issue #1168.

0xArdi:
I think that having a mechanism that handles exceptions by default in `async_act` can be quite useful.
In the ElCollectooorr most of the act methods are wrapped with try/except, and error payloads are sent in case of exceptions.
What we might have is:
- Define a default error/unhandled exception payload (or payload data) and event on round side
- Wrap `async_act` in try/except
- In case of exception, send the payload above
- Handle it appropriately on the round, in case of majority failure handle it appropriately, otherwise ignore.

PR #1385, opened 2022-09-23
Fix/counter client tests

0xArdi:
Fixes counter client behaviour tests.


packages/valory/skills/counter_client/tests/test_behaviours.py
```diff
@@ -63,19 +64,26 @@ def setup_class(cls, **kwargs: Any) -> None:
             HttpHandler, cls._skill.skill_context.handlers.http_handler
         )
 
+    def ticker_act_wrapper(self) -> None:
+        """Calls act wrapper after waiting for tick interval + tolerance."""
+        tolerance = 0.5
+        sleep_amount = self.behaviour.tick_interval + tolerance
+        time.sleep(sleep_amount)
+        self.behaviour.act_wrapper()
```
DavidMinarsch, line 72: Hm, doesn't this suggest the implementation of the skill should have a sleep rather than the test?

0xArdi: the skill has a sleep that is the problem. `act_wrapper` gets ignored.

0xArdi: Snippet from [TickerBehaviour.act_wrapper()](https://github.com/valory-xyz/open-aea/blob/main/aea/skills/behaviours.py#L149):
```
    def act_wrapper(self) -> None:
        """Wrap the call of the action. This method must be called only by the framework."""
        if not self.is_done() and self.is_time_to_act():
            self._last_act_time = datetime.datetime.now()
            super().act_wrapper()
```

DavidMinarsch: Ah ok, I see what's happening. Let's move on then and keep your fix.

Adamantios (commented):
LGTM, let's reset the workflow file.

DavidMinarsch (approved):

PR #1390, opened 2022-09-23
Remove skip mark

Adamantios:
## Proposed changes

Remove skip mark for `test_service_hashing` which was fixed on `open-aea` side.

## Fixes

n/a

## Types of changes

## Checklist

## Further comments

n/a

### DIFF
--- a/tests/test_autonomy/test_cli/test_hash.py
+++ b/tests/test_autonomy/test_cli/test_hash.py
@@ -21,13 +21,11 @@
 
 
 import json
-import platform
 import shutil
 from pathlib import Path
 from typing import Dict, Tuple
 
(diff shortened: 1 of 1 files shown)


0xArdi (approved):

DavidMinarsch (commented):
What do we observe? The hash of the agent folder created on Windows and other OSes is different. On open-autonomy the `counter` agent has a `tests` dir. On open-aea, where we test the hashing feature for agents we don't have a single agent with such a `tests` folder.

Next step: suggest adding such an agent on `open-aea` and then seeing if we can reproduce the failure there. Then we can fix it there.

DavidMinarsch (changes requested):
We need a fix on open-aea to make the tests pass.

PR #1394, opened 2022-09-26
Use parametrized os value in workflow

Adamantios:
## Proposed changes

Uses the parametrized os value introduced in #1389.

## Fixes

n/a

## Types of changes

## Checklist

## Further comments

n/a


Adamantios:
We had forgotten to use the parametrized os' value. Now CI fails because we are using ubuntu-specific installation commands.

.github/workflows/main_workflow.yml
```diff
@@ -24,14 +24,9 @@ jobs:
     - uses: actions/setup-python@master
       with:
         python-version: ${{ matrix.python-version }}
-    - name: Install dependencies
-      run: |
-        sudo apt-get update --fix-missing
-        sudo apt-get autoremove
-        sudo apt-get autoclean
```
Adamantios, line 31: These were not necessary for the job.

Pipfile
```diff
@@ -11,14 +11,13 @@ open-aea-cli-ipfs = "==1.20.0"
...
 asn1crypto = "==1.4.0"
-atheris = "==2.0.11"
 bandit = "==1.7.0"
 black = "==21.6b0"
 bs4 = "==0.0.1"
 certifi = "==2021.10.8"
 click = "==8.0.2"
 darglint = "==1.8.0"
-docker = "==5.0.2"
+docker = "==6.0.0"
```
Adamantios, line 20: Update `docker` to avoid conflict with `pywin32` on Windows.

Pipfile
```diff
@@ -11,14 +11,13 @@ open-aea-cli-ipfs = "==1.20.0"
 
 [dev-packages]
 asn1crypto = "==1.4.0"
-atheris = "==2.0.11"
```
Adamantios, line 14: Remove the `atheris` dependency because it is not supported on Windows.

The options would be to:
1. Add `atheris` in the `setup.py` with a condition to install only if not on Windows. This however would add an unnecessary dependency to our framework.
2. Replace `pipenv` with `poetry` which supports "environment markers" and install dev dependency only if not on Windows. We could do this if we reopen #93.
3. Remove the dependency and let the dev install it if needed. This does not cause any issues since we already have the following logic in the fuzzy tests:

```python
try:
    import atheris  # type: ignore
except (ImportError, ModuleNotFoundError):
    pytestmark = pytest.mark.skip
```
The best option, for now, would be (3.).

README.md
```diff
@@ -90,6 +90,11 @@ the dependencies:
 
       make new_env && pipenv shell
 
+  > :information_source: Note: we are using [atheris](https://github.com/google/atheris) in order to perform fuzzy testing. 
+  > The dependency is not listed in the `Pipfile` because it is not supported on Windows. 
+  > If you need to run or implement a fuzzy test, please manually install the dependency.
+  > If you are developing on Mac, please follow the extra steps described [here](https://github.com/google/atheris#mac).
```
Adamantios, line 96: Added clarification re the `atheris` dependency.

setup.py
```diff
@@ -49,25 +49,6 @@ def get_all_extras() -> Dict:
...
-        docker 5.0.3 depends on pywin32==227; sys_platform == "win32"
-        open-aea[all] 1.20.0 depends on pywin32==304
-
-    Instead, at commit docker/docker-py@3f0095a, `setup.py` has an extra that forces pywin32>=304.
-    """
-    return (
-        "docker==5.0.3"
-        if platform.system() != "Windows"
-        else "docker @ git+https://github.com/docker/docker-py.git@3f0095a7c1966c521652314e524ff362c24ff58c"
-    )
```
Adamantios, line 68: Not necessary since we updated `docker`.

Adamantios:
Failing test on macos reports:
```
Unreliable test timings! On an initial run, this test took 4940.30ms, which exceeded the deadline of 2000.00ms, but on a subsequent run it took 6.89 ms, which did not. If you expect this sort of variability in your test timings, consider turning deadlines off for this test by setting deadline=None.
```

I think this variability is expected @Karrenbelt. Let's increase the deadline for `hypothesis`.

plugins/aea-test-autonomy/setup.py
```diff
@@ -19,38 +19,15 @@
...
-        else "docker @ git+https://github.com/docker/docker-py.git@3f0095a7c1966c521652314e524ff362c24ff58c"
-    )
-
-
 base_deps = [
     "open-aea[all]>=1.20.0,<2.0.0",
     "pytest==7.0.0",
     "open-aea-ledger-ethereum==1.20.0",
-    _get_docker_dependency(),
+    "docker==6.0.0",
```
DavidMinarsch, line 30: @angrybayblade is that compatible with the rest of the plugins etc?

setup.py
DavidMinarsch: Same question as above

.github/workflows/main_workflow.yml
Adamantios: Though why would we add this in the workflow if this does not affect the CI?

packages/valory/connections/abci/tests/test_abci_fuzz.py
```diff
@@ -253,7 +253,7 @@ def list_to_tuple(node: Node) -> Node:
 def make_aea_test_method(message_key: str, strategy: Node) -> Callable:
     """Dynamically create AEA test"""
 
-    @settings(deadline=2000, suppress_health_check=[HealthCheck.too_slow])
+    @settings(deadline=5000, suppress_health_check=[HealthCheck.too_slow])
```
Adamantios, line 256: https://github.com/valory-xyz/open-autonomy/pull/1394#issuecomment-1263295057

packages/valory/connections/abci/tests/test_abci_fuzz.py
```diff
@@ -321,7 +321,7 @@ def create_tendermint_hypotheses() -> Node:
 def make_tendermint_test_method(message_key: str, strategy: Node) -> Callable:
     """Dynamically create Tendermint test"""
 
-    @settings(deadline=2000, suppress_health_check=[HealthCheck.too_slow])
+    @settings(deadline=5000, suppress_health_check=[HealthCheck.too_slow])
```
Adamantios, line 324: https://github.com/valory-xyz/open-autonomy/pull/1394#issuecomment-1263295057

plugins/aea-test-autonomy/setup.py
angrybayblade: https://github.com/valory-xyz/open-autonomy/blob/main/setup.py#L65

@Adamantios Please update this as well

angrybayblade: Also, it would be nice if we can use a range instead of a strict version specifier

Adamantios: > Please update this as well
I have:
https://github.com/valory-xyz/open-autonomy/blob/38e950524df79759915489dffc5a3349efd18623/setup.py#L61

> Also, it would be nice if we can use a range instead of a strict version specifier

We can't:
https://github.com/valory-xyz/open-autonomy/pull/1394#discussion_r984253932

DavidMinarsch (approved):

PR #1397, opened 2022-09-27
Set initial fallback gas and propagate logs

Adamantios:
## Proposed changes

Sets an initial fallback gas and propagates the logs.

## Fixes

n/a

## Types of changes

## Checklist

## Further comments

n/a


packages/valory/contracts/gnosis_safe/contract.py
```diff
@@ -462,7 +462,7 @@ def get_raw_safe_transaction(  # pylint: disable=too-many-arguments,too-many-loc
             tx_parameters.update(ledger_api.try_get_gas_pricing(old_price=old_price))
         # note, the next line makes an eth_estimateGas call iff gas is not set!
         transaction_dict = w3_tx.buildTransaction(tx_parameters)
-        if configured_gas != 1:
+        if configured_gas != MIN_GAS:
```
0xArdi, line 465: I feel like the name `MIN_GAS` is a bit confusing here. Maybe something like `NO_GAS` or something like that is more appropriate. Also, I don't see why we cannot set `configured_gas` to None and check for it as `None`, that seems much more natural to me.

0xArdi (approved):

Adamantios:
Ran this on staging DO and the `intrinsic gas too low` RPC error is gone.

DavidMinarsch (approved):

Issue #1396, opened 2022-09-27
Containers need more time to start on CI from `contracts-amm` image


Adamantios:
### Subject of the issue
Some tests are flaky because containers fail to start on time on CI. Specifically, this happened on [Merge pull request #1395 from valory-xyz/feat/autoupdate-packagelist main_workflow #5418](https://github.com/valory-xyz/open-autonomy/actions/runs/3133747181/jobs/5087596544) for some tests that use the `contracts-amm` image.

We could try increasing the max attempts, e.g., from 20 to 25, here:
https://github.com/valory-xyz/open-autonomy/blob/61c7f69b3e6290e94fdc6837cf3bbaaddccdcf22/plugins/aea-test-autonomy/aea_test_autonomy/fixture_helpers.py#L277

Example of the failure:
```
2022-09-27T08:08:33.0046808Z INFO     aea_test_autonomy.docker.base:base.py:140 Setting up image valory/contracts-amm:latest...
2022-09-27T08:08:33.0047362Z DEBUG    urllib3.connectionpool:connectionpool.py:228 Starting new HTTP connection (1): 127.0.0.1:8545
2022-09-27T08:08:33.0048124Z ERROR    root:amm_net.py:98 Exception: ConnectionError: ('Connection aborted.', ConnectionResetError(104, 'Connection reset by peer'))
2022-09-27T08:08:33.0048653Z INFO     root:amm_net.py:99 Attempt 0 failed. Retrying in 2.0 seconds...
.
.
.
2022-09-27T08:08:33.0083634Z DEBUG    urllib3.connectionpool:connectionpool.py:228 Starting new HTTP connection (1): 127.0.0.1:8545
2022-09-27T08:08:33.0084423Z ERROR    root:amm_net.py:98 Exception: ConnectionError: ('Connection aborted.', RemoteDisconnected('Remote end closed connection without response'))
2022-09-27T08:08:33.0084953Z INFO     root:amm_net.py:99 Attempt 19 failed. Retrying in 2.0 seconds...
```

PR #1400, opened 2022-09-28
Do not indirectly handle `NotFittedError`

Adamantios:
## Proposed changes

The `predict_safely` should handle a `pmdarima`'s issue without "hiding" other exceptions. However, the way it was implemented, it indirectly handled `NotFittedError` because [it inherits from `ValueError`](https://github.com/scikit-learn/scikit-learn/blob/36958fb24/sklearn/exceptions.py#L19), which we are handling:
https://github.com/valory-xyz/open-autonomy/blob/35baabb52d07bb6f45d616ab9e63d5223c77fb4a/packages/valory/skills/apy_estimation_abci/ml/forecasting.py#L344

Therefore, we catch and throw this exception earlier.

## Fixes

n/a

## Types of changes

## Checklist

## Further comments

n/a


packages/valory/skills/apy_estimation_abci/ml/forecasting.py
```diff
@@ -341,6 +342,9 @@ def predict_safely(forecaster: Pipeline, steps_forward: int) -> Any:
     """
     try:
         y_hat = forecaster.predict(steps_forward)
+    except NotFittedError:
```
DavidMinarsch, line 345: Where is it handled?

Adamantios: I chose to re-raise and not handle it since this would have been a developer's error, i.e., using a non-fitted model to predict.

Adamantios:
35baabb reveals the issue, 362fa4c fixes it.

0xArdi (approved):

Adamantios:
Ubuntu failure caused by #1396.

DavidMinarsch (approved):

PR #1402, opened 2022-09-28
[0.3.2] Replace start timestamp with number of observations

Adamantios:
## Proposed changes

Replace the start timestamp in the configuration with the number of observations we want for the timeseries.
We now have to set the number of data points we wish to download in the past (starting from the configured end, using the configured interval).
This is more intuitive than setting the exact start timestamp in UNIX, resulting in better DevX.

## Fixes

n/a

## Types of changes

## Checklist

## Further comments

n/a


.github/workflows/main_workflow.yml
```diff
@@ -164,7 +164,7 @@ jobs:
         os: [ ubuntu-latest, macos-latest, windows-latest ]
         python-version: [ "3.7", "3.8", "3.9", "3.10" ]
 
-    timeout-minutes: 65
+    timeout-minutes: 70
```
Adamantios, line 167: Increased timeout as I have noticed windows tests timing out lately and I also added two new `hypothesis` tests.

packages/valory/services/apy_estimation/service.yaml
```diff
@@ -22,7 +22,7 @@ models:
         spooky_subgraph:
         - '0x2a651563c9d3af67ae0388a5c8f89b867038089e'
         - '0x2b4c76d0dc16be1c31d4c1dc53bf9b45987fc75c'
-      history_start: ${SERVICE_APY_HISTORY_START:int:null}
+      n_observations: ${SERVICE_APY_N_OBSERVATIONS:int:120}
```
Adamantios, line 25: Set a default of 120 observations which is a good number to start experimenting with. Moreover, now the operator does not need to specify anything if they do not need to update the number of observations. Previously, they had to update the `history_start` every time they restarted the service which was cumbersome.

packages/valory/skills/apy_estimation_abci/behaviours.py
```diff
@@ -306,7 +306,7 @@ def setup(self) -> None:
 
         self._unit = sec_to_unit(self.params.interval)
         self._target_per_pool = int(
-            unit_amount_from_sec(self.params.end - self.params.start, self._unit)
+            unit_amount_from_sec(self.params.ts_length, self._unit)
```
Adamantios, line 309: We now have a property (`ts_length`) for this.

packages/valory/skills/apy_estimation_abci/behaviours.py
```diff
@@ -1402,7 +1403,6 @@ def async_act(self) -> Generator:
             self.context.logger.info(
                 "Estimates have been received:\n" f"{estimates.to_string()}"
             )
-            self.context.logger.info("Estimates have been received.")
```
Adamantios, line 1406: Removed because it was duplicate.

packages/valory/skills/apy_estimation_abci/models.py
```diff
@@ -192,9 +194,21 @@ def __init__(self, *args: Any, **kwargs: Any) -> None:
 
         self.__validate_params()
 
+    @property
+    def start(self) -> Optional[int]:
```
Adamantios, line 198: `start` is now given by a property that utilizes the `ts_length`. The formula is:
```
start = end - n_observations * interval
```
end and interval are given in UNIX, and therefore start is also in UNIX.

plugins/aea-test-autonomy/aea_test_autonomy/fixture_helpers.py
```diff
@@ -274,7 +274,7 @@ def ganache_scope_class(
 @pytest.fixture(scope="class")
 def ammnet_scope_class(
     timeout: float = 2.0,
-    max_attempts: int = 20,
+    max_attempts: int = 26,
```
Adamantios, line 277: Addresses #1396. 12 extra seconds should be enough for CI.

docs/api/plugins/aea_test_autonomy/fixture_helpers.md
```diff
@@ -257,7 +257,7 @@ Launch the Ganache image. This fixture is scoped to a class which means it will
 
 ```python
 @pytest.fixture(scope="class")
-def ammnet_scope_class(timeout: float = 2.0, max_attempts: int = 20) -> Generator
+def ammnet_scope_class(timeout: float = 2.0, max_attempts: int = 26) -> Generator
```
Karrenbelt, line 260: why 26?

plugins/aea-test-autonomy/aea_test_autonomy/fixture_helpers.py
Karrenbelt: ok so that's why. It's an odd number to pick though

packages/valory/skills/apy_estimation_abci/tests/test_models.py
```diff
@@ -127,6 +130,45 @@ def test_setup(
...
+        with mock.patch.object(
+            APYParams,
+            "ts_length",
+            new_callable=mock.PropertyMock,
+            return_value=ts_length,
+        ):
+            if expected is None:
+                assert params.start is expected
+            else:
+                assert params.start == expected
```
Karrenbelt, line 155: `assert params.start == expected`  will work for both cases

plugins/aea-test-autonomy/aea_test_autonomy/fixture_helpers.py
Adamantios: I think it's an even :sweat_smile:

packages/valory/skills/apy_estimation_abci/tests/test_behaviours.py
```diff
@@ -826,6 +835,7 @@ def test_fetch_behaviour_non_indexed_block(
         self,
         is_non_indexed_res: bool,
         block_from_timestamp_q: str,
+        timestamp_gte: str,
```
Karrenbelt, line 838: what does `gte` stand for?

Adamantios: Short for greater than or equal in GraphQL.

packages/valory/skills/apy_estimation_abci/tests/test_behaviours.py
```diff
@@ -509,32 +509,42 @@ def test_check_given_pairs(
...
         batch: bool,
-        start: int,
+        n_observations: int,
         interval: int,
         end: int,
     ) -> None:
         """Test `_reset_timestamps_iterator` method."""
+        # filter out end values that will result in negative `start`
+        assume(end >= n_observations * interval)
+
```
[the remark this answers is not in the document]
Adamantios: I think that they both filter the values and used assume as I found it easier to read.

packages/valory/services/apy_estimation/service.yaml
Karrenbelt: yes. There probably is some range that should be allowed, not up to infinity

dvilelaf: In aea-config.yaml n_observations is 10 while here it is 120. Is that right?

Adamantios: Agreed. I plan to extend this method at some point, though we might want to introduce a proper way for validation to all of our models.
https://github.com/valory-xyz/open-autonomy/blob/554269bf20fe1f4ded081fa81833b1563d59b63c/packages/valory/skills/apy_estimation_abci/models.py#L209-L220

Have you heard of [pydantic](https://pydantic-docs.helpmanual.io/)? Using it might produce some cool results.

dvilelaf (commented):
Looks good. Some failing e2e.

packages/valory/skills/apy_estimation_abci/tests/test_models.py
Adamantios: Done in e8fd98314e7d854e9e763fe1ecb944ea7ebac9a6.

dvilelaf (approved):

Adamantios:
> Looks good. Some failing e2e.

Previous run failed because of #1168.

DavidMinarsch (approved):

PR #1406, opened 2022-09-29
docs: add contract development link

dvilelaf:
## Proposed changes

Add a link in the docs to our contract dev guide.

## Fixes

n/a

## Types of changes

## Checklist

### DIFF
--- a/docs/guides/set_up.md
+++ b/docs/guides/set_up.md
@@ -31,7 +31,7 @@ pip install open-autonomy[all]
 
 4. Initialize the framework to work with the remote [IPFS](https://ipfs.io) registry. This means that when the framework will be fetching a component, it will do so from the [IPFS](https://ipfs.io):
     ```bash
-    autonomy init --reset --author john_doe --remote --ipfs --ipfs-node "/dns/registry.autonolas.tech/tcp/443/https"
+    autonomy init --remote --ipfs
     ```
 
     !!!info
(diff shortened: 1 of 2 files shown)


jmoreira-valory (approved):

docs/guides/set_up.md
```diff
@@ -31,7 +31,7 @@ pip install open-autonomy[all]
 
 4. Initialize the framework to work with the remote [IPFS](https://ipfs.io) registry. This means that when the framework will be fetching a component, it will do so from the [IPFS](https://ipfs.io):
     ```bash
-    autonomy init --remote --ipfs
+    autonomy init --reset --author john_doe --remote --ipfs --ipfs-node "/dns/registry.autonolas.tech/tcp/443/https"
```
jmoreira-valory, line 34: Does reset work if it's the first time (i.e., there is nothing to reset) ?

(Subjective comment: From a user's perspective it's a bit "ugly" adding this --reset here. I know why it is added, but it is kind of strange for a user to see that he needs to use some "reset option" to make the framework work  --- Anyways this should be a 'devx issue'.)

dvilelaf: Yes, I've just run it on my machine following the tuto from scratch.

dvilelaf: This morning the Brahma intern tried the quickstart (I also run it locally a couple times) and got:
`Error: Registry path not provided and local registry `packages` not found in current (.) and parent directory.`
So I think we still need to provide the url. Is this right @angrybayblade ?

angrybayblade: This was caused because the command used `local` repo, this can be fixed by using `--local` (Also please open an issue to update the error messages I think this issue has been dragged out really long, we can update the error message so the user knows they are using local registry and decide if they want to use the remote registry and retry)

PR #1410, opened 2022-09-30
Chore/hash table

dvilelaf:
## Proposed changes

* Adds a script to automatically generate a package table for the docs
* Aligns the IPFS fixer script with open-aea

## Fixes

n/a

## Types of changes

## Checklist


docs/guides/managing_packages.md
```diff
@@ -32,12 +32,12 @@ Now we are in the position of creating our first agent:
...
 ```
 
-This agent will have some dependencies: some packages might be already available in the remote registry and some will need to be developed. Let's say, for example, that it will need the ability to communicate using the ACN protocol. Since that package is already available on the Autonolas IPFS registry, we'll use the `add command`:
+This agent will have some dependencies: some packages might be already available in the remote registry and some will need to be developed. Let's say, for example, that it will need the ability to use the signing protocol protocol. Since that package is already available on the Autonolas IPFS registry, we'll use the `add command`:
 
 ```bash
 cd my_agent/
 # Remote flag is not needed here as we initialized the default registry to remote
-autonomy add protocol valory/acn:1.1.0:bafybeifontek6tvaecatoauiule3j3id6xoktpjubvuqi3h2jkzqg7zh7a
+autonomy add protocol open_aea/signing:1.0.0:bafybeiambqptflge33eemdhis2whik67hjplfnqwieoa6wblzlaf7vuo44
```
dvilelaf, line 40: This was changed because acn package was ported so it is easier to check and autoupdate.

scripts/check_doc_ipfs_hashes.py
```diff
@@ -271,10 +271,10 @@ def check_ipfs_hashes(  # pylint: disable=too-many-locals,too-many-statements
             hash_mismatches = True
 
             if fix:
-                new_content = content.replace(doc_full_cmd, new_command)
+                content = content.replace(doc_full_cmd, new_command)
 
                 with open(str(md_file), "w", encoding="utf-8") as qs_file:
-                    qs_file.write(new_content)
+                    qs_file.write(content)
```
dvilelaf, line 277: This was a bug: if several hashes were mismatching in the same file, only the first one would be corrected.

jmoreira-valory (approved):

DavidMinarsch (approved):

PR #1423, opened 2022-10-11
[WIP] Add support for ACN and hardhat node in deployment setup

angrybayblade:
## Proposed changes

This PR adds support for ACN and hardhat node in deployment setup using build flags

- ACN node is required for agent communication when running a local deployment
- Hardhat node works as a chain for local deployments

Using `--use-hardhat` and `--use-acn` flags you can include both nodes in the deployment setup so you don't have to run a standalone docker containers for them

## Fixes

## Types of changes

## Checklist


autonomy/deploy/generators/docker_compose/base.py
```diff
@@ -39,12 +43,17 @@
...
     ABCI_NODE_TEMPLATE,
+    ACN_NODE_TEMPLATE,
     DOCKER_COMPOSE_TEMPLATE,
+    HARDHAT_NODE_TEMPLATE,
     TENDERMINT_CONFIG_TEMPLATE,
     TENDERMINT_NODE_TEMPLATE,
 )
 
 
+N_RESERVED_IP_ADDRESSES = 4
```
angrybayblade, line 54: We will add one ACN and hardhat nodes at the top with IP being `192.167.11.2` and `192.167.11.3` so we will start from `192.167.11.4` for abci and tendermint nodes

angrybayblade:
With this feature we will be able to spin up a local deployment with an ACN and a hardhat node included. 

We just need to update several package configs to run a deployment with ACN communication enabled, I'll open a separate PR for this

docs/api/cli/deploy.md
```diff
@@ -88,11 +88,23 @@ Deploy an agent service.
     default=False,
     help="Apply environment variable when loading service config.",
 )
+@click.option(
+    "--use-hardhat",
```
DavidMinarsch, line 92: We removed this a while ago. Why are we reintroducing it? We had a good reason to remove it

Karrenbelt (commented):
please add more informative PR descriptions. Will be good for other engineers and our future selves as well. Why ACN node - can simply reference github issue / trello ticket / google doc.
Not clear to me why hardhat node is added.

docs/api/cli/deploy.md
angrybayblade: So I'm not aware on why the usage of hardhat node inside the deployment was removed, but we did not have this flag before we were including the hardhat node by default in docker compose deployments or in the dev mode in case of kubernetes deployments (https://github.com/valory-xyz/open-autonomy/pull/809#discussion_r876012304). I suppose we removed this because we wanted to use external blockchains in production. But this lead to a very annoying dev experience where we had to run a separate hardhat node when we wanted to run a local deployment. 

This flag makes the inclusion of the hardhat node optional, I think we can change this to be strictly for local deployments only but I think there is a pretty good reason to introduce this functionality

Karrenbelt (commented):
There are e2e failures due to not passing the health check. It would be useful to be able to provide more as part of the error here, such as a display of the connection setup configurations to facilitate debugging;  e.g. both issues with ACN and Tendermint can lead to this error.

```
Exit Code	Package
1       	valory/agents/register_reset
1       	valory/agents/registration_start_up
1       	valory/agents/simple_abci
1       	valory/agents/apy_estimation
1       	valory/agents/hello_world
1       	valory/agents/oracle
```

Tests for kubernetes `use_acn` and `use_hardhat` are missing


There also is a failure on win 3.10 related to hardhat, which does not appear to be directly related. 
```
ConnectionResetError: [WinError 10054] An existing connection was forcibly closed by the remote host
...
ERROR tests/test_deployments/test_app.py::TestTendermintHardResetServer::test_hard_reset[False]
```

docs/api/cli/deploy.md
DavidMinarsch: Ok. Then make it strictly for local only please.

DavidMinarsch:
> With this feature we will be able to spin up a local deployment with an ACN and a hardhat node included.
> 
> We just need to update several package configs to run a deployment with ACN communication enabled, I'll open a separate PR for this

Please open this PR so we can see it all working

PR #1432, opened 2022-10-14
[WIP] Update agent overrides to support local ACN integration

angrybayblade:
## Proposed changes

This PR updates agent and service level overrides

## Fixes

## Types of changes

## Checklist


packages/valory/agents/oracle/aea-config.yaml
```diff
@@ -138,23 +138,13 @@ public_id: valory/p2p_libp2p_client:0.1.0
 type: connection
 config:
   nodes:
-  - uri: ${P2P_URI:str:acn.staging.autonolas.tech:9005}
-    public_key: ${P2P_PUBLIC_KEY:str:02d3a830c9d6ea1ae91936951430dee11f4662f33118b02190693be835359a9d77}
-  - uri: ${P2P_URI:str:acn.staging.autonolas.tech:9006}
-    public_key: ${P2P_PUBLIC_KEY:str:02e741c62d706e1dcf6986bf37fa74b98681bc32669623ac9ee6ff72488d4f59e8}
+  - uri: localhost:11000
+    public_key: 03c74dbfbe7bbc1b42429f78778017a3cd7eaf9d59d1634c9505a3f7c1a9350e71
```
angrybayblade, line 142: We want to start with single node

angrybayblade (commented):
I tried updating the service config with overrides but it seems that current override policy on service component is a bit rigid so the service loading fails with new overrides, I think we'll have to prioritize this ticket https://trello.com/c/8yRWIuAj

angrybayblade:
I also got this error while running the ACN integration setup locally 

```
Traceback (most recent call last):

  File "/home/ubuntu/.local/share/virtualenvs/ubuntu-7Wf190Ea/lib/python3.10/site-packages/aea/skills/base.py", line 440, in act_wrapper
    raise AEAActException(

aea.exceptions.AEAActException: An error occurred during act of behaviour valory/oracle_abci:0.1.0/RegistrationStartupBehaviour:
Traceback (most recent call last):

  File "/home/ubuntu/agent/vendor/valory/skills/abstract_round_abci/base.py", line 1909, in last_timestamp
    raise ABCIAppInternalError("last timestamp is None")

packages.valory.skills.abstract_round_abci.base.ABCIAppInternalError: internal error: last timestamp is None
```

This happens after the validator config is updated and we restart tendermint node

packages/valory/services/oracle_hardhat/service.yaml
```diff
@@ -16,8 +16,9 @@ setup_args:
       safe_contract_address: []
     observation_interval: 30
     broadcast_to_server: false
-    service_registry_address: '0x0DCd1Bf9A1b36cE34237eEaFef220932846BCD82'
+    service_registry_address: ${SERVICE_REGISTRY_ADDRESS:str:0x998abeb3E57409262aE5b751f60747921B33613E}
```
DavidMinarsch, line 19: why did the default change?

packages/valory/agents/oracle/aea-config.yaml
DavidMinarsch: Why? And if so, why change the public_key and get rid of the env var?

packages/valory/agents/oracle/aea-config.yaml
```diff
@@ -138,23 +138,13 @@ public_id: valory/p2p_libp2p_client:0.1.0
 type: connection
 config:
   nodes:
-  - uri: ${P2P_URI:str:acn.staging.autonolas.tech:9005}
-    public_key: ${P2P_PUBLIC_KEY:str:02d3a830c9d6ea1ae91936951430dee11f4662f33118b02190693be835359a9d77}
-  - uri: ${P2P_URI:str:acn.staging.autonolas.tech:9006}
-    public_key: ${P2P_PUBLIC_KEY:str:02e741c62d706e1dcf6986bf37fa74b98681bc32669623ac9ee6ff72488d4f59e8}
+  - uri: localhost:11000
```
DavidMinarsch, line 141: why change to local? Default should be remote?

packages/valory/agents/oracle/aea-config.yaml
```diff
@@ -138,23 +138,13 @@ public_id: valory/p2p_libp2p_client:0.1.0
...
+    public_key: 03c74dbfbe7bbc1b42429f78778017a3cd7eaf9d59d1634c9505a3f7c1a9350e71
 cert_requests:
 - identifier: acn
   ledger_id: ethereum
   message_format: '{public_key}'
   not_after: '2023-01-01'
   not_before: '2022-01-01'
-  public_key: ${P2P_PUBLIC_KEY:str:02d3a830c9d6ea1ae91936951430dee11f4662f33118b02190693be835359a9d77}
```
DavidMinarsch, line 151: see above

DavidMinarsch (changes requested):

packages/valory/agents/oracle/aea-config.yaml
angrybayblade: The same reason as we keep URLs for chain or other specs to some local URLs, The user will mostly run these agents in a local environment and the deployments will be in once a moon so I think it's not a good idea to point these to a production or a staging node. 

These were supposed to be overridable env vars which we use when deploying the service, But because of this https://github.com/valory-xyz/open-autonomy/pull/1432#pullrequestreview-1142255156 we cannot do it.

packages/valory/services/oracle_hardhat/service.yaml
angrybayblade: We cannot use the default hardhat image anymore because it does not have the registry contracts deployed and we cannot use the staging chain anymore this address points to the service registry on staging chain. `valory/autonolas-registries` have these contracts deployed so we use it as the local hardhat deployment, and it has the service registry deployed on the updated address

packages/valory/agents/oracle/aea-config.yaml
angrybayblade: Continued from https://github.com/valory-xyz/open-autonomy/pull/1432#discussion_r996609138, as I said these are just place holders which can be overridden at the runtime using service level overrides. But for that first we need to implement the override policies and update the service component

DavidMinarsch:
Closing as this code no longer lives in this repo. It is now here: https://github.com/valory-xyz/price-oracle

packages/valory/services/oracle_hardhat/service.yaml
DavidMinarsch: Ok

packages/valory/agents/oracle/aea-config.yaml
DavidMinarsch: `${P2P_URI:str:localhost:11000}`

packages/valory/agents/oracle/aea-config.yaml
DavidMinarsch: same here: `${P2P_PUBLIC_KEY:str:03c74dbfbe7bbc1b42429f78778017a3cd7eaf9d59d1634c9505a3f7c1a9350e71}

DavidMinarsch: We should have a list somewhere in the docs with the local and staging addresses

PR #1441, opened 2022-10-18
Extract price oracle

Adamantios:
## Proposed changes

Remove the price oracle which has been extracted to https://github.com/valory-xyz/price-oracle.

## Fixes

n/a

## Types of changes

## Checklist

## Further comments

n/a


scripts/check_doc_ipfs_hashes.py
```diff
@@ -49,6 +49,9 @@
 PACKAGE_TABLE_REGEX = rf"\| {PACKAGE_TYPE_REGEX}\/{VENDOR_REGEX}\/{PACKAGE_REGEX}\/{VERSION_REGEX}(\s|\|)*(?P<hash>{IPFS_HASH_REGEX})\s*\|"
 
 ROOT_DIR = Path(__file__).parent.parent
+# We need to skip the hash for `valory/oracle_hardhat:0.1.0:bafybeie553shfmnds6v7defynjv5kmjkqf2aygj345jbbcssevtnkbodbe`
+# because the demo is not in this repo, but the docs are
+HASH_SKIPS = ("bafybeie553shfmnds6v7defynjv5kmjkqf2aygj345jbbcssevtnkbodbe",)
```
Adamantios, line 54: This is not ideal. Eventually, we will end up having an outdated hash. I suggest that we open an issue to resolve this. Reading the hash from the `price-estimation` repo and replacing it in the doc file should be enough.
@dvilelaf @jmoreira-valory

skaffold.yaml
```diff
@@ -122,13 +114,6 @@ profiles:
           <<: *docker_args
           dockerfile: deployments/Dockerfiles/documentation/Dockerfile
 
-      - image: valory/oar-oracle
```
DavidMinarsch, line 125: Make sure we have a skaffold.yaml in the apy and price repos

scripts/check_doc_ipfs_hashes.py
DavidMinarsch: No I think this needs to be a manual process for now. We can add in the release script that the hash be updated here (and in the docs)! It's like updating a library.

packages/packages.json
```diff
@@ -6,46 +6,37 @@
...
     "agent/valory/counter/0.1.0": "bafybeiewt2xx4ptqy5osnumekbu26uvv34cmzvmdirjt3wreucjt5ebtgy",
     "agent/valory/counter_client/0.1.0": "bafybeibrzheonnpbkihtov7e45yhs5azgo57k5ogxnykucpyv6sprufb7m",
     "agent/valory/hello_world/0.1.0": "bafybeideb6b5k4i6z7bm3p53eydxgknmwdefo2oshcnlxthjc6oxeox7ua",
-    "agent/valory/oracle/0.1.0": "bafybeifaowcfzsgqau7hxowybb32t3buspbhbvjlkyuhoyc5mnbqz2i7w4",
     "agent/valory/register_reset/0.1.0": "bafybeibfad2qv5gs5sbnv4drxba5fyrb4sb7moprbvvq7jkhagc5qwfoka",
     "agent/valory/registration_start_up/0.1.0": "bafybeibpnjhetlif6zesescxcypqcscedb7u3dgpmv4nfbbwabuwj42k3m",
     "agent/valory/simple_abci/0.1.0": "bafybeiez76fbbul4xyizbytr424onz3s2dd2ct3lyjip5wkjylk7opmxmi",
     "agent/valory/test_abci/0.1.0": "bafybeifllqlz32rz3uglgvug3vffr753stagotmijzu2gws24z2qmxg6ce",
     "contract/valory/multisend/0.1.0": "bafybeig45676hbh4c3p3mujrrskxgxww4cxdyyginlg5rmmav6orv4gtya",
+    "contract/valory/offchain_aggregator/0.1.0": "bafybeibp2cbuaxj4sthoaf5vdbdos6am2kqcbomrt2x6cyqsmxu34jabbu",
```
DavidMinarsch, line 35: This needs to be extracted to the oracle!

Makefile
```diff
@@ -219,17 +219,9 @@ teardown-kubernetes:
...
 	python -m autonomy.cli analyse abci generate-app-specs packages.valory.skills.safe_deployment_abci.rounds.SafeDeploymentAbciApp packages/valory/skills/safe_deployment_abci/fsm_specification.yaml || (echo "Failed to check safe_deployment_abci consistency" && exit 1)
 	python -m autonomy.cli analyse abci generate-app-specs packages.valory.skills.simple_abci.rounds.SimpleAbciApp packages/valory/skills/simple_abci/fsm_specification.yaml || (echo "Failed to check simple_abci consistency" && exit 1)
 	python -m autonomy.cli analyse abci generate-app-specs packages.valory.skills.transaction_settlement_abci.rounds.TransactionSubmissionAbciApp packages/valory/skills/transaction_settlement_abci/fsm_specification.yaml || (echo "Failed to check transaction_settlement_abci consistency" && exit 1)
 	echo "Successfully validated abcis!"
-
-
-AEA_AGENT_ORACLE:=valory/oracle:latest:$(shell cat packages/packages.json | grep "agent/valory/oracle" | cut -d "\"" -f4 )
```
DavidMinarsch, line 233: The command should stay. We just don't need the env vars anymore here - this can also be ported to price

DavidMinarsch (changes requested):

Makefile
Adamantios: I am confused. The command will be empty here since we removed the oracle.

Also, we should port it to both price and APY and we will need to duplicate some code for now. I opened a feature request for that #1442.

skaffold.yaml
Adamantios: I will open PRs for that. Also, take a look at https://github.com/valory-xyz/open-autonomy/pull/1441#discussion_r998185966 re duplication

scripts/check_doc_ipfs_hashes.py
Adamantios: Added in the process on f225f937863f9fc2821d663ecb33d90c4edc0c38.

Makefile
DavidMinarsch: Not an issue. We don't need that folder as the images will be already pushed from here. In order to push we need to run `skaffold build -p release --cache-artifacts=false && skaffold build -p release-latest`, hence why it cannot be removed

Adamantios: True, I missed that, we need to release the framework as well!

scripts/check_doc_ipfs_hashes.py
jmoreira-valory, line 54: There is an alternative about this, I suggested sometime ago: Don't write hashes in the docs, and write some macro which automatically checks the last Tagged version for the correct hash and populates it at compile time. E.g.,
autonomy fetch {% get_hash(valory/oracle) %}  etc..

jmoreira-valory: (in this case, it should be required to indicate the repo as well)

jmoreira-valory (dismissed):
As far as the docs is concerned, I don't see it should break anything (apparently), except the hash issue.

DavidMinarsch (approved):

PR #1446, opened 2022-10-19
fix: adding benchmarktool to scaffolding

dvilelaf:
## Proposed changes

Adds benchmarktool to scaffolding.

## Fixes

n/a

## Types of changes

## Checklist


autonomy/cli/scaffold_fsm_templates/components.py
```diff
@@ -276,6 +279,7 @@ def __init__(self, *args: Any, **kwargs: Any) -> None:
 
     Params = BaseParams
     Requests = BaseRequests
+    BenchmarkTool = BaseBenchmarkTool
```
DavidMinarsch, line 282: It also needs to be added in the skill config template

dvilelaf: According to @angrybayblade , since the skill.yaml scaffolding happens on open-aea side, what we will need to do is, after the FSM scaffolding is done:
- Load the skill config 
- Add the benchmark tool model definition
- Save it
I need to look into it, but as I'm still finding other small template errors, I think it's better to leave this PR open for now and I focus on the contribution service adding here the errors I find along the way. Once finished with it, then I look into this issue and fix it here.

packages/valory/skills/abstract_round_abci/test_tools/rounds.py
```diff
@@ -576,7 +576,7 @@ def _test_round(
         for behaviour_attr_getter in synchronized_data_attr_checks:
             assert behaviour_attr_getter(synchronized_data) == behaviour_attr_getter(
                 actual_next_synchronized_data
-            )
+            ), f"Mismatch in synchronized_data. Actual:\n{behaviour_attr_getter(synchronized_data)}\nExpected:\n{behaviour_attr_getter(actual_next_synchronized_data)}"
```
DavidMinarsch, line 579: Words and variable names don't match..

autonomy/cli/scaffold_fsm_templates/components.py
DavidMinarsch: We already update the skill yaml: https://github.com/valory-xyz/open-autonomy/blob/main/autonomy/cli/scaffold_fsm.py#L351

packages/valory/skills/abstract_round_abci/test_tools/rounds.py
dvilelaf: I think there's a naming problem here: `synchronized_data` is the result of `end_round()` and `actual_next_synchronized_data` the result of calling `synchronized_data_update_fn()` with `final_data`, so if I understand it correctly here actual = expected.

autonomy/cli/scaffold_fsm_templates/tests.py
```diff
@@ -62,7 +62,7 @@ class RoundTestCase:
 
         name: str
         initial_data: Dict[str, Hashable]
-        payloads: BaseTxPayload
+        payloads: Mapping[str, BaseTxPayload]
```
dvilelaf, line 65: This change is what is breaking the tests

angrybayblade: That was caused by a missing import in the template string, I've fixed it here https://github.com/valory-xyz/open-autonomy/pull/1446/commits/7a1f2a48bd0ab76b60e032cec55ec96e877116cb

dvilelaf:
Some unrelated test failures.

autonomy/cli/scaffold_fsm_templates/components.py
DavidMinarsch: @dvilelaf please address this - it's possible!

dvilelaf: Done

autonomy/cli/scaffold_fsm.py
```diff
@@ -416,6 +416,10 @@ def _update_models(  # pylint: disable=no-self-use
         config.models.create(
             "abci_dialogues", SkillComponentConfiguration("AbciDialogues")
         )
+        config.models.create(
+            "benchmark_tool",
+            SkillComponentConfiguration("BenchmarkTool", log_dir="/logs"),
```
Karrenbelt, line 421: `/` will cause issue on windows?

dvilelaf: All the skills have the `/logs` in their skill.yaml so I don't think it should be an issue.

angrybayblade (approved):
The falky test failures are being addressed here https://github.com/valory-xyz/open-autonomy/pull/1457

DavidMinarsch (approved):

PR #1448, opened 2022-10-19
Fix host parsing on tendermint handler

angrybayblade:
## Proposed changes

This PR fixes the URL check on tendermint handler and updates the test

## Fixes

## Types of changes

## Checklist


packages/valory/skills/abstract_round_abci/handlers.py
```diff
@@ -664,6 +666,12 @@ def _handle_response(
         try:  # validate message contains a valid address
             validator_config = json.loads(message.info)
             self.context.logger.error(validator_config)
+            parse_result = urlparse(validator_config["tendermint_url"])
+            if (
+                parse_result.hostname != "localhost"
+                and not parse_result.hostname.startswith("node")
+            ):
+                ipaddress.ip_network(parse_result.hostname)
```
angrybayblade, line 674: I've updated the code to allow `node{x}` hosts so we can run local deployments but this is not the way to check for valid URLs because this assumes the validator host to be an ip address whereas we can have domain name as a validator URL. I'll open a ticket to update this check

DavidMinarsch (approved):

PR #1457, opened 2022-10-19
Use subprocesses in `TestCheckSpecs` and `TestScaffoldFSMAutonomyTests`

marcofavorito:
## Proposed changes

This PR makes `TestCheckSpecs` and `TestScaffoldFSMAutonomyTests` to use subprocess for each test to better isolate the Python environment of the CLI command being executed.

## Fixes

n/a

## Types of changes

## Checklist

## Further comments

n/a

### DIFF
--- a/tests/test_autonomy/test_cli/test_analyse/test_abci/test_specs.py
+++ b/tests/test_autonomy/test_cli/test_analyse/test_abci/test_specs.py
@@ -161,30 +161,30 @@ class TestCheckSpecs(BaseCliTest):
         self,
     ) -> None:
         """Test with one class."""
-        result = self.run_cli(
+        return_code, stdout, stderr = self.run_cli_subprocess(
             ("--app-class", self.cls_name, "--infile", str(self.specification_path))
         )
 
-        assert result.exit_code == 0
-        assert result.output == f"Checking : {self.cls_name}\nCheck successful\n"
+        assert return_code == 0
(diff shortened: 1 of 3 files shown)


tests/test_autonomy/test_cli/test_analyse/test_abci/test_specs.py
```diff
@@ -229,6 +231,12 @@ def test_failures(
             "Please provide path to specification file." in result.output
         ), result.output
 
+    def teardown(
+        self,
+    ) -> None:
+        """Tear down the test."""
```
DavidMinarsch, line 237: you need to call `super().teardown()` o/w we don't have proper cleanup

marcofavorito: right

Karrenbelt (commented):
I think pytest monkeypatch might offer a solution: https://docs.pytest.org/en/7.1.x/how-to/monkeypatch.html

marcofavorito:
> I think pytest monkeypatch might offer a solution: https://docs.pytest.org/en/7.1.x/how-to/monkeypatch.html

Thank you for the suggestion! I will use it.

marcofavorito:
Finally tests are passing. Squashing all commits into one.

DavidMinarsch (approved):

PR #1459, opened 2022-10-20
[0.3.4]: feat/ Termination

0xArdi:
## Proposed changes

This PR adds a behaviour & round which runs concurrently with the rest of an agent, and is responsible for the termination of a service.

## Types of changes

## Checklist


packages/valory/skills/abstract_round_abci/behaviours.py
```diff
@@ -165,6 +168,7 @@ def __init__(self, **kwargs: Any) -> None:
         ] = self._get_round_to_behaviour_mapping(self.behaviours)
 
         self.current_behaviour: Optional[BaseBehaviour] = None
+        self.__termination_behaviour: Optional[BaseBehaviour] = None
```
DavidMinarsch, line 171: Why `__`? The whole class is a framework side class.

packages/valory/skills/abstract_round_abci/behaviours.py
```diff
@@ -236,6 +243,7 @@ def act(self) -> None:
         if self.current_behaviour is None:
             return
 
+        self.__termination_behaviour.act_wrapper()
```
DavidMinarsch, line 246: Does it make sense here or does it make more sense after 251. The latter would give the `current_behaviour` the option to finish.

packages/valory/skills/abstract_round_abci/base.py
```diff
@@ -1912,6 +1918,33 @@ def last_timestamp(self) -> datetime.datetime:
     def setup(self) -> None:
         """Set up the behaviour."""
         self._schedule_round(self.initial_round_cls)
+        self._termination_round = TerminationRound(
+            self._initial_synchronized_data,
+            self.consensus_params,
+            None,
```
DavidMinarsch, line 1924: would remove `None` as it's the default arg anyway

packages/valory/skills/abstract_round_abci/base.py
```diff
@@ -1912,6 +1918,33 @@ def last_timestamp(self) -> datetime.datetime:
...
+        )
+
+    def _get_transaction_settlement_initial_round(
```
DavidMinarsch, line 1927: This is a problem. The abstract_round_abci cannot know anything about this skill. What we need to do instead is to extend the FSM spec to include the mapping from the TerminalRound to the user defined FSM (to begin with `transaction_settlement_abci`).

packages/valory/skills/abstract_round_abci/base.py
```diff
@@ -1912,6 +1918,33 @@ def last_timestamp(self) -> datetime.datetime:
...
+        self,
+    ) -> Optional[Type[AbstractRound]]:
+        """
+        Returns the initial round of `transaction_settlement_abci`
+
+        This method works under two assumptions.
+            1. The transaction_settlement_abci is present in the running agent.
+            2. The initial round of transaction_settlement_abci is "randomness_transaction_submission".
+        Note that we cannot just reference the round directly, as this would cause a cyclic dependency,
+        abstract_round_abci -> transaction_settlement_abci -> abstract_round_abci
```
DavidMinarsch, line 1937: Sort of should have given you the hint ;)

packages/valory/skills/abstract_round_abci/base.py
```diff
@@ -2481,11 +2547,23 @@ def _update_round(self) -> None:
         Check whether the round has finished. If so, get the
         new round and set it as the current round.
         """
-        result: Optional[
+        # TODO: exit agent when termination is done ie, transaction_settlement_abci is finished after being scheduled
```
DavidMinarsch, line 2550: Relating to my comment above, I have an idea how we can define this on the spec level via a transition from transaction_settlement_abci to a custom "shutdown" behaviour. This would be preferrable as we would not create any dependency on the transaction_settlement_abci. Can discuss Monday.

packages/valory/skills/abstract_round_abci/background_round.py
```diff
@@ -0,0 +1,512 @@
...
+_SAFE_GAS = 0
+
+# hardcoded to 0 because we don't need to send any ETH when handing over multisig ownership
+_ETHER_VALUE = 0
+
+NULL_ADDRESS: str = "0x" + "0" * 40
+MAX_UINT256 = 2 ** 256 - 1
+
+# copied from transaction_settlement_abci/payload_tools.py to avoid cyclic dependency
+def hash_payload_to_hex(  # pylint: disable=too-many-arguments, too-many-locals
```
DavidMinarsch, line 53: Hm so many dependencies... we could extract this whole background setup as a `background_abci` into its own skill and introduce it via the config level - as suggested below for `transaction_settlement_abci` too. That would allow us to define arbitrary skills (with one round and one behaviour) as background skills.

DavidMinarsch (changes requested):

packages/valory/skills/termination_abci/behaviours.py
```diff
@@ -0,0 +1,522 @@
...
+
+        multisend_data = cast(str, response.raw_transaction.body["data"])[2:]
+        return multisend_data
+
+    def wait_for_termination_majority(self) -> Generator:
+        """
+        Wait until we reach majority on the termination transaction.
+        :yield: None
+        """
+        yield from self.wait_for_condition(self._is_termination_majority)
```
Karrenbelt, line 447: so what happens without timeout. We wait, if we do not get majority the RESET_TIMEOUT will kick in and the system transitions?

packages/valory/skills/termination_abci/skill.yaml
```diff
@@ -0,0 +1,101 @@
...
+    args: {}
+    class_name: Requests
+  signing_dialogues:
+    args: {}
+    class_name: SigningDialogues
+  state:
+    args: {}
+    class_name: SharedState
+dependencies: {}
+is_abstract: true
```
Karrenbelt, line 101: what is this `is_abstract` property contingent upon?

packages/valory/skills/termination_abci/behaviours.py
0xArdi: This round is immune to transitions.

packages/valory/skills/termination_abci/skill.yaml
DavidMinarsch: It ensures the skill itself is not run. We only want the composed skill to be run.

packages/valory/skills/termination_abci/rounds.py
```diff
@@ -0,0 +1,134 @@
...
+    (
+        TransactionSubmissionAbciApp,
+        PostTerminationTxAbciApp,
+    ),
+    termination_transition_function,
+)
+
+TerminationAbciApp.transition_function[BackgroundRound] = {
+    Event.TERMINATE: TransactionSubmissionAbciApp.initial_round_cls,
+}
```
DavidMinarsch, line 134: Please open an issue so this hack is not needed in the future.

packages/valory/skills/termination_abci/dialogues.py
```diff
@@ -0,0 +1,58 @@
...
+    HttpDialogues as BaseHttpDialogues,
+)
+from packages.valory.skills.abstract_round_abci.dialogues import (
+    SigningDialogue as BaseSigningDialogue,
+)
+from packages.valory.skills.abstract_round_abci.dialogues import (
+    SigningDialogues as BaseSigningDialogues,
+)
+
+
```
DavidMinarsch, line 47: Why don't we have TendermintDialogue/s?

packages/valory/skills/termination_abci/models.py
```diff
@@ -0,0 +1,54 @@
...
+
+class SharedState(BaseSharedState):
+    """Keep the current shared state of the skill."""
+
+    def __init__(self, *args: Any, **kwargs: Any) -> None:
+        """Initialize the state."""
+        super().__init__(*args, abci_app_cls=TerminationAbciApp, **kwargs)
+
+
+class TerminationParams(BaseParams):
```
DavidMinarsch, line 43: Here do `TerminationParams(BaseParams, TransactionParams)` so you don't have to do it here (and therefore in each app that uses these params) https://github.com/valory-xyz/open-autonomy/pull/1459/files#diff-484c7525547a1274d651241dba29755a16df071438a99014a967dfa2aecb6edeR53-R54.

packages/valory/skills/termination_abci/models.py
```diff
@@ -0,0 +1,54 @@
...
+    """Defines the class to hold the termination parameters."""
+
+    def __init__(self, *args: Any, **kwargs: Any) -> None:
+        """Set up the termination parameters."""
+        self.termination_sleep = self._ensure("termination_sleep", kwargs)
+        self.multisend_address = self._ensure("multisend_address", kwargs)
+        super().__init__(*args, **kwargs)
+
+
```
DavidMinarsch, line 52: Here also add randomness API: RandomnessApi = TransactionSettlementRandomness

packages/valory/skills/register_termination_abci/models.py
```diff
@@ -0,0 +1,55 @@
...
+
+class SharedState(BaseSharedState):
+    """Keep the current shared state of the skill."""
+
+    def __init__(self, *args: Any, **kwargs: Any) -> None:
+        """Initialize the state."""
+        super().__init__(*args, abci_app_cls=RegisterTerminateAbciApp, **kwargs)
+
+
+class Params(TerminationParams, TransactionParams):
```
DavidMinarsch, line 54: If you apply comments in `packages/valory/skills/termination_abci/models.py` then we don't have a direct dependency on `transaction_settlement_abci` anymore here. That's highly desirable as it makes it much easier to compose (need to know less about the internals). The whole skill then also no longer has a direct dependency on `transaction_settlement_abci`

DavidMinarsch (changes requested):
Please address my comments in a new PR on top

packages/valory/skills/termination_abci/dialogues.py
DavidMinarsch: @Adamantios

packages/valory/skills/termination_abci/models.py
DavidMinarsch: Then you can import `RandomnessApi` from termination. Although just an alias, it makes it much easier for the user (they no longer need to know about the dependency on transaction_settlement_abci internals)

packages/valory/skills/termination_abci/behaviours.py
```diff
@@ -0,0 +1,553 @@
...
+    """Behaviour responsible for terminating the agent."""
+
+    state_id: str = "termination"
+    behaviour_id = "termination_behaviour"
+    matching_round = TerminationRound
+
+    def async_act(self) -> Generator:
+        """Logs termination and terminates."""
+        self.context.logger.info("Terminating the agent.")
+        sys.exit()
```
0xArdi, line 540: A more sophisticated termination logic might be required.

packages/valory/skills/register_termination_abci/composition.py
```diff
@@ -0,0 +1,52 @@
...
+    (
+        RegistrationAbci.AgentRegistrationAbciApp,
+        ResetAndPauseAbci.ResetPauseABCIApp,
+    ),
+    abci_app_transition_mapping,
+).add_termination(
+    background_round_cls=BackgroundRound,
+    termination_event=Event.TERMINATE,
+    termination_abci_app=TerminationAbciApp,
+)
```
0xArdi, line 52: We can define a `with_termination(user_abci_app)` util function that sets the termination related attrs to a user defined abci app. I think this would be helpful for the user because the `.add_termination()` args are always the same.

packages/valory/skills/termination_abci/dialogues.py
Adamantios: This skill is never meant to be used standalone. Therefore, shouldn't the accompanied FSM define the tm dialogues since the termination skill does not utilize them?

DavidMinarsch (approved):

PR #1480, opened 2022-10-26
Feat/add termination

0xArdi:
## Proposed changes

This PR adds a way for an AbciApp to add termination support to it. 

## Types of changes

## Checklist

## Further comments

### DIFF
--- a/packages/valory/skills/abstract_round_abci/base.py
+++ b/packages/valory/skills/abstract_round_abci/base.py
@@ -1914,13 +1914,32 @@ class AbciApp(
             events.update(transitions.keys())
         return events
 
-    @classmethod
-    def get_all_round_classes(cls) -> Set[AppState]:
-        """Get all round classes."""
+    @staticmethod
+    def _get_rounds_from_transition_function(
+        transition_function: Optional[AbciAppTransitionFunction],
+    ) -> Set[AppState]:
+        """Get rounds from a transition function."""
+        if transition_function is None:
+            return set()
         result: Set[AppState] = set()
         for start, transitions in transition_function.items():
             result.add(start)
             result.update(transitions.values())
+        return result
+
+    @classmethod
+    def get_all_round_classes(
+        cls, include_termination_rounds: bool = False
+    ) -> Set[AppState]:
+        """Get all round classes."""
+        result: Set[AppState] = cls._get_rounds_from_transition_function(
+            cls.transition_function
+        )
+        if include_termination_rounds:
+            termination_rounds = cls._get_rounds_from_transition_function(
+                cls.termination_transition_function,
+            )
+            result.update(termination_rounds)
         return result
 
     @property
--- a/packages/valory/skills/abstract_round_abci/base.py
+++ b/packages/valory/skills/abstract_round_abci/base.py
@@ -1830,7 +1830,7 @@ class AbciApp(
     final_states: Set[AppState] = set()
     event_to_timeout: EventToTimeout = {}
     cross_period_persisted_keys: List[str] = []
-    termination_round_cls: Optional[AppState] = None
+    background_round_cls: Optional[AppState] = None
     termination_transition_function: Optional[AbciAppTransitionFunction] = None
     termination_event: Optional[EventType] = None
     _is_abstract: bool = True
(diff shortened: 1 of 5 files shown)


packages/valory/skills/abstract_round_abci/base.py
```diff
@@ -1901,15 +1914,34 @@ def get_all_events(cls) -> Set[EventType]:
...
+        result: Set[AppState] = cls._get_rounds_from_transition_function(
+            cls.transition_function
+        )
+        if include_termination_rounds:
+            termination_rounds = cls._get_rounds_from_transition_function(
+                cls.termination_transition_function,
+            )
+            result.update(termination_rounds)
+        return result
+
```
0xArdi, line 1944: Added so that the consistency checks won't fail.

packages/valory/skills/abstract_round_abci/behaviours.py
```diff
@@ -106,7 +106,9 @@ def _check_matching_round_consistency(
         """Check that matching rounds are: (1) unique across behaviour, and (2) covering."""
         round_to_behaviour: Dict[Type[AbstractRound], List[BehaviourType]] = {
             round_cls: []
-            for round_cls in behaviour_cls.abci_app_cls.get_all_round_classes()
+            for round_cls in behaviour_cls.abci_app_cls.get_all_round_classes(
+                behaviour_cls.is_background_behaviour_set
+            )
```
0xArdi, line 111: We take into account the termination rounds only when we have the background round set.

packages/valory/skills/termination_abci/behaviours.py
```diff
@@ -526,28 +589,72 @@ def is_round_ended(self, round_id: str) -> Callable[[], bool]:
...
+TerminationAbciApp = chain(
+    (
+        TransactionSubmissionAbciApp,
+        PostTerminationTxAbciApp,
+    ),
+    termination_transition_function,
+)
+TerminationAbciApp.transition_function[BackgroundRound] = {
+    Event.TERMINATE: TransactionSubmissionAbciApp.initial_round_cls,
+}
```
0xArdi, line 648: thats the trigger

packages/valory/skills/abstract_round_abci/base.py
```diff
@@ -1858,16 +1858,29 @@ def __init__(
...
+        background_round_cls: Type[AppState],
+        termination_event: EventType,
+        termination_abci_app: "Type[AbciApp]",
+    ) -> "Type[AbciApp]":
+        """Sets the termination related class variables."""
+        cls.background_round_cls = background_round_cls
+        cls.termination_transition_function = termination_abci_app.transition_function
+        cls.termination_event = termination_event
+        return cls
+
```
0xArdi, line 1883: The use would be:
```python
AnyAbciApp = chain(
    (
        RegistrationAbci.AgentRegistrationAbciApp,
        SafeDeploymentAbci.SafeDeploymentAbciApp,
        TransactionSubmissionAbci.TransactionSubmissionAbciApp,
        ResetAndPauseAbci.ResetPauseABCIApp,
    ),
    abci_app_transition_mapping,
).add_termination(BackgroundRound, Event.TERMINATE, TerminationAbciApp)
```
or
```python
AnyAbciApp.add_termination(BackgroundRound, Event.TERMINATE, TerminationAbciApp)
```

0xArdi: Because the arguments here will always be the same, we can have a `with_termination()` function in `termination_abci` that would automatically add these. 
The use of it could be something like:
```python
with_termination(AnyAbciApp)
```

Karrenbelt: I like that it can be used as an optional extension. Are there specific reasons / advantages to doing this over having it within the `chain` call?

DavidMinarsch: Yes the separation is desirable.

packages/valory/skills/termination_abci/behaviours.py
```diff
@@ -526,28 +589,72 @@ def is_round_ended(self, round_id: str) -> Callable[[], bool]:
...
+    state_id: str = "termination"
+    behaviour_id = "termination_behaviour"
+    matching_round = TerminationRound
+
+    def async_act(self) -> Generator:
+        """Logs termination and terminates."""
+        self.context.logger.info("Terminating the agent.")
+        ok = 0
+        quit(ok)
+        yield
```
DavidMinarsch, line 622: what does `quit()` do?

For now it's ok probably, but we need to do a bit more cleanup. Can do later.

0xArdi: Same as `exit()`, it will raise a `SystemExit` exception. I am open to alternatives.

DavidMinarsch: For now ok - but open an issue as I want to extend this later.

DavidMinarsch (approved):

PR #1481, opened 2022-10-26
[0.3.4]: chore: split round and payloads into separate files

0xArdi:
## Proposed changes

This PR moves the termination rounds and payloads into their respective files.

## Types of changes

## Checklist

## Further comments

### DIFF
--- a/packages/valory/skills/termination_abci/behaviours.py
+++ b/packages/valory/skills/termination_abci/behaviours.py
@@ -17,21 +17,9 @@
 #
 # ------------------------------------------------------------------------------
 
-"""This module contains the background behaviour and round classes."""
-from enum import Enum
+"""This module contains the termination behaviour classes."""
 from mailbox import Message
-from typing import (
-    Any,
-    Callable,
-    Dict,
-    Generator,
-    List,
-    Optional,
-    Set,
-    Tuple,
-    Type,
-    cast,
-)
+from typing import Callable, Dict, Generator, List, Optional, Set, Type, cast
 
 from hexbytes import HexBytes
 
@@ -45,29 +33,23 @@ from packages.valory.contracts.multisend.contract import (
 )
 from packages.valory.contracts.service_registry.contract import ServiceRegistryContract
 from packages.valory.protocols.contract_api import ContractApiMessage
-from packages.valory.skills.abstract_round_abci.abci_app_chain import chain
-from packages.valory.skills.abstract_round_abci.base import (
-    AbciApp,
-    AbstractRound,
-    BaseSynchronizedData,
-    BaseTxPayload,
-    CollectSameUntilThresholdRound,
-)
 from packages.valory.skills.abstract_round_abci.behaviour_utils import (
     AsyncBehaviour,
     BaseBehaviour,
 )
 from packages.valory.skills.abstract_round_abci.behaviours import AbstractRoundBehaviour
+from packages.valory.skills.termination_abci.payloads import BackgroundPayload
+from packages.valory.skills.termination_abci.rounds import (
+    BackgroundRound,
+    SynchronizedData,
+    TerminationAbciApp,
+    TerminationRound,
+)
 from packages.valory.skills.transaction_settlement_abci.behaviours import (
     TransactionSettlementRoundBehaviour,
 )
 from packages.valory.skills.transaction_settlement_abci.payload_tools import (
     hash_payload_to_hex,
-)
-from packages.valory.skills.transaction_settlement_abci.rounds import (
-    FailedRound,
-    FinishedTransactionSubmissionRound,
-    TransactionSubmissionAbciApp,
 )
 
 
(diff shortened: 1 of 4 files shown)


packages/valory/skills/termination_abci/rounds.py
```diff
@@ -0,0 +1,126 @@
...
+        return None
+
+
+class PostTerminationTxAbciApp(AbciApp):
+    """The abci app running after the multisig transaction has been settled."""
+
+    initial_round_cls = TerminationRound
+    # the following is not needed, it is added to satisfy the round check
+    # the TerminationRound when run it terminates the agent, so nothing can come after it
+    transition_function = {TerminationRound: {Event.TERMINATE: TerminationRound}}
```
DavidMinarsch, line 109: For this we have degenerate rounds & behaviours. Create one and reference it here?

packages/valory/skills/termination_abci/rounds.py
```diff
@@ -0,0 +1,126 @@
...
+    (
+        TransactionSubmissionAbciApp,
+        PostTerminationTxAbciApp,
+    ),
+    termination_transition_function,
+)
+
+TerminationAbciApp.transition_function[BackgroundRound] = {
+    Event.TERMINATE: TransactionSubmissionAbciApp.initial_round_cls,
+}
```
DavidMinarsch, line 126: Why is this being updated after ?

packages/valory/skills/termination_abci/rounds.py
```diff
@@ -0,0 +1,126 @@
...
+        """No logic required here."""
+        pass
+
+    def process_payload(self, payload: BaseTxPayload) -> None:
+        """No logic required here."""
+        pass
+
+    def end_block(self) -> Optional[Tuple[BaseSynchronizedData, Enum]]:
+        """No logic required here."""
+        return None
```
DavidMinarsch, line 100: Why does this have no implementation?

packages/valory/skills/termination_abci/behaviours.py
```diff
@@ -622,32 +510,6 @@ def async_act(self) -> Generator:
...
-        PostTerminationTxAbciApp,
-    ),
-    termination_transition_function,
-)
-TerminationAbciApp.transition_function[BackgroundRound] = {
-    Event.TERMINATE: TransactionSubmissionAbciApp.initial_round_cls,
-}
-
-
 class TerminationAbciBehaviours(AbstractRoundBehaviour):
```
DavidMinarsch, line 513: It's confusing that this is already a "chained" behaviour. Have a look at https://github.com/valory-xyz/apy-oracle/blob/main/packages/valory/skills/apy_estimation_abci/behaviours.py#L1562, we just define the standalone ABCI app, here it would be two: BackgroundAbciBehaviours & TerminationAbciBehaviours.

DavidMinarsch (commented):
A bit of conflation of concerns between the chained part and the isolate ABCI apps (and their behaviours and rounds). We have two very basic apps here with one skill each & then a composition with those two + transaction settlement. We can keep them all in one (`is_abstract`) skill, but it should be split more clearly on a code level imo.

packages/valory/skills/termination_abci/rounds.py
0xArdi: Because we don't have any logic here, the moment the agent enters this behaviour it gets terminated.

packages/valory/skills/termination_abci/rounds.py
0xArdi: To avoid having another 1 class abci app. On second thought, we can get rid of this and handle it in the AbciApp side.

packages/valory/skills/termination_abci/rounds.py
0xArdi: So this satisfies two checks:
1. That the [inital state is in the set of states](https://github.com/valory-xyz/open-autonomy/blob/v0.3.3/packages/valory/skills/abstract_round_abci/base.py#L1741).
2. That the TerminationRound has a [non timeout transition](https://github.com/valory-xyz/open-autonomy/blob/v0.3.3/packages/valory/skills/abstract_round_abci/base.py#L1811).

packages/valory/skills/termination_abci/behaviours.py
0xArdi: I would say that we don't need the `BackgroundAbciBehaviours` here. It's a single behavior only.

packages/valory/skills/termination_abci/rounds.py
DavidMinarsch: Yes that's what I'm suggesting.

packages/valory/skills/termination_abci/rounds.py
DavidMinarsch: I think it's much more transparent if we have a number of single class apps. Then we chain them, then we append them via the new `add_termination`

DavidMinarsch (approved):

PR #1483, opened 2022-10-27
fix: remove skipped doc links

dvilelaf:
## Proposed changes

Remove skipped link checks for CI testing purposes.
Investigating [this](https://github.com/valory-xyz/open-autonomy/actions/runs/3288400682/jobs/5418997739) failure that was passing locally:
```
Awaiting for results...
Found broken url in the docs:
docs/get_started/why_do_we_need_agent_services.md: ['https://encyclopedia.pub/entry/2959']
ERROR: InvocationError for command /home/runner/work/open-autonomy/open-autonomy/scripts/check_doc_links.py (exited with code 1)
```

Redirects should not be a problem as the [requests library automatically follows](https://requests.readthedocs.io/en/latest/user/quickstart/#redirection-and-history) those. Could it be that we were getting a TooManyRedirects and being rate-limited? I've added the http status code to the logging so it is easier to debug and fix next time.

## Fixes

n/a

## Types of changes

## Checklist

### DIFF
--- a/scripts/check_doc_links.py
+++ b/scripts/check_doc_links.py
@@ -43,6 +43,21 @@ MAX_WORKERS = 10
 URL_REGEX = r'(https?:\/\/(?:www\.|(?!www))[a-zA-Z0-9][a-zA-Z0-9-]+[a-zA-Z0-9]\.[^\s)"]{2,}|www\.[a-zA-Z0-9][a-zA-Z0-9-]+[a-zA-Z0-9]\.[^\s)"]{2,}|https?:\/\/(?:www\.|(?!www))[a-zA-Z0-9]+\.[^\s)"]{2,}|www\.[a-zA-Z0-9]+\.[^\s)"]{2,})'
 REQUEST_TIMEOUT = 5  # seconds
 
+# Allow some links to be HTTP because there is no HTTPS alternative
+# Remove non-url-allowed characters like ` before adding them here
+HTTP_SKIPS = [
+    "http://www.fipa.org/repository/ips.php3",
+    "http://host.docker.internal:8545",
+]
+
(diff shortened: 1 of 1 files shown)


scripts/check_doc_links.py
```diff
@@ -141,7 +145,7 @@ def main() -> None:  # pylint: disable=too-many-locals
         if broken_links:
             broken_links_str = "\n".join(
                 [
-                    f"{file_name}: {[url[1] for url in urls]}"
+                    f"{file_name}: {[(url[1], url[2]) for url in urls]}"
```
dvilelaf, line 148: Show the status code next time it fails.

Adamantios: Future devs and self would need to study the `broken_links` in order to understand what these indexes are accessing. A `dataclass` or a `dict` would have been more helpful. However, we could at least add a comment on the code with what these indexes are accessing.

Adamantios (dismissed):

scripts/check_doc_links.py
dvilelaf: Fixed

Adamantios (approved):

scripts/check_doc_links.py
```diff
@@ -90,29 +131,18 @@ def main() -> None:  # pylint: disable=too-many-locals
     """Check for broken or HTTP links"""
     all_md_files = [str(p.relative_to(".")) for p in Path("docs").rglob("*.md")]
 
-    broken_links: Dict[str, List[str]] = {}
+    broken_links: Dict[str, Dict] = {}
```
Karrenbelt, line 134: next time please specify type inside nested `Dict` as well

PR #1485, opened 2022-10-28
Chore/termination lint

0xArdi:
## Proposed changes

This PR applies linting fixes to the changes introduced by the #1459. 

## Types of changes

## Checklist


packages/valory/skills/abstract_round_abci/base.py
```diff
@@ -1954,6 +1954,7 @@ def setup(self) -> None:
         """Set up the behaviour."""
         self._schedule_round(self.initial_round_cls)
         if self.is_termination_set:
+            self.background_round_cls = cast(AppState, self.background_round_cls)
             self._background_round = self.background_round_cls(
```
Karrenbelt, line 1958: I know it's very annoying this casting for mypy, makes code messy, next time consider making it a property so you don't have to keep casting. A property that raises if not `self.is_termination_set` or so. Just a suggestion. I also kinda dislike the fact that 3 class attribute are needed for this, and you can guess why I'm sure but anyway: we use too many class attributes and they are the devil because state caries over more easily than when you're using an instance attribute.

packages/valory/skills/abstract_round_abci/base.py
```diff
@@ -2133,17 +2139,22 @@ def process_event(
         # if that's the case, we proceed with the termination transition function,
         # regardless of what the current round is
         if self.is_termination_set and event == self.termination_event:
+            self.termination_transition_function = cast(
+                AbciAppTransitionFunction, self.termination_transition_function
+            )
+            self.background_round_cls = cast(AppState, self.background_round_cls)
             next_round_cls = self.termination_transition_function[
                 self.background_round_cls
             ].get(event, None)
```
Karrenbelt, line 2148: we only enter here is event == termination_event, so what is the reason to have an optional `next_round_cls` here?

packages/valory/skills/abstract_round_abci/base.py
```diff
@@ -2582,15 +2593,13 @@ def _update_round(self) -> None:
...
-                Tuple[BaseSynchronizedData, Any]
-            ] = self.abci_app.background_round.end_block()
+            background_result = self.abci_app.background_round.end_block()
         if background_result is not None and not self._termination_called:
             # when the background round returns, it takes priority over normal rounds
             # because the BackgroundRound never ends, we should only take into account
             # its response only once
             self._termination_called = True
-            result = background_result
+            result: Optional[Tuple[BaseSynchronizedData, Any]] = background_result
```
Karrenbelt, line 2602: this `Any` should be an `Enum`, representing some sort of `Event`, correct?

packages/valory/skills/abstract_round_abci/base.py
0xArdi: That was the definition before this change. I didn't see a reason to change it.

packages/valory/skills/abstract_round_abci/base.py
0xArdi: Correct, again as above the definition of result was like that, and I didn't see a reason to change it.

packages/valory/skills/termination_abci/behaviours.py
```diff
@@ -61,8 +63,8 @@
 # hardcoded to 0 because we don't need to send any ETH when handing over multisig ownership
 _ETHER_VALUE = 0
 
-NULL_ADDRESS: str = "0x" + "0" * 40
-MAX_UINT256 = 2 ** 256 - 1
+# payload to represent a non-existing event
+_NO_EVENT_FOUND: Dict = {}
```
Karrenbelt, line 67: use a sentinel: `_NO_EVENT_FOUND = object()`

packages/valory/skills/termination_abci/behaviours.py
```diff
@@ -156,7 +179,7 @@ def check_for_signal(self) -> Generator[None, None, bool]:
     def _get_latest_removed_owner_event(self) -> Generator[None, None, Optional[Dict]]:
         """Returns the latest event in which the service owner was removed from the set of owners of the safe."""
         response = yield from self.get_contract_api_response(
-            performative=ContractApiMessage.Performative.GET_STATE,
+            performative=ContractApiMessage.Performative.GET_STATE,  # type: ignore
```
Karrenbelt, line 182: any idea why mypy always complains about this? :shrug: 
`assert isinstance(ContractApiMessage.Performative.GET_STATE, ContractApiMessage.Performative)`

0xArdi: I gave up after a while. It passes on another (private) repo.

Karrenbelt (approved):

DavidMinarsch (approved):

PR #1486, opened 2022-10-28
[DO NOT MERGE] Docs/guide tests

dvilelaf:
## Proposed changes

Implement guide test base class and the first test for the quickstart guide.

Code blocks in the documentation have been moved to external .txt files so they are easier to organize and parse without conflating different guides or codeblocks in the same file that should be run separately. Optionally, this could be achieved by parsing the markdown files but hardcoding block separation when needed.

These tests collect all the commands in the txt files, run those commands in order and check that they finish succesfully.

Pending implementation:
- [ ] Command output check
- [ ] Simultaneous commands (for example, docker logs check)
- [ ] Skip comments in txt files

## Fixes

n/a

## Types of changes

## Checklist


tests/test_docs/test_guides.py
```diff
@@ -0,0 +1,194 @@
...
+    def teardown(self) -> None:
+        """Teardown"""
+        # Remove execution directory
+        remove_test_directory(str(self.execution_path))
+
+        # Run teardown commands
+        self.run_commands(self.teardown_commands)
+
+
+@pytest.mark.e2e
```
dvilelaf, line 149: I think the test is not running. I've marked it e2e but I don't see it in the logs. Maybe the e2e jobs only runs tests from a specific folder, need to investigate. It passes locally.

.github/workflows/main_workflow.yml
```diff
@@ -387,12 +387,16 @@ jobs:
...
           docker pull valory/safe-contract-net:latest
 
-      - name: Unit tests
+      - name: Package e2e tests
         run: |
           tox --develop -e packages-py${{ matrix.python-version }}-linux -- -m 'e2e'
+
+      - name: Guide tests
+        run: |
+          tox --develop -e py${{ matrix.python-version }}-linux -- -m 'e2e'
```
dvilelaf, line 402: Adding the guide tests as a separate job.

jmoreira-valory:
I think the am of this PR is important, as it is true that documentation comands can get outdated easily. However, I feel the current approach of the PR will make somewhat harder to maintain documentation. Here is my proposal on how to tackle this issue, which, of course, can be discarded/overriden by @DavidMinarsch.

Instead of writing the commands in separate .txt files, wirte a "magic comment" that will be ignored by the Markdown parser, but that our script will interpret and "virtually generate" these .txt file contents on the fly. For example:

```
Please follow this instructions to fetch and deploy the Hello world service:

<--- test_this_code_magic_comment -->
'''bash
autonomy fetch bafibey123451....
cd hello_world
cp ../keys.json .
autonomy deploy
''' 
```

The script will
1. search for occurrences of the magic comment <--- test_this_code_magic_comment --> (change approppriately)
2. collect the commands between the first opening ''' and the immediate ''' following (I think it's called greedy match),
3. possibly remove the language 'bash' and do any other required sanitizations.
4. do whatever checks it is doing now in the separate .txt files. 

This has the advantage of not "dismembering" the page, enhancing manteinability, with little effort (just adding one comment on top of each code that needs to be tested).

It should be decided: (1) what is the best name for this magic comment, and (2) what is the best way to use comments in markdown. The one presented works, but **it leaves the comment in the output HTML file** (i.e., an user can see the comment if he views the source code of the page). See https://stackoverflow.com/questions/4823468/comments-in-markdown  for how to avoid this situation.

jmoreira-valory:
Alternatively, it can be enforced that every code block comment in the documentation has either a magic comment to ignore it OR a magic comment to verify the block. For example:

With this comment, the linter should parse and verify the code block:
```
<!-- LINTER_VERIFY -->
'''bash
cd <directory>
ls
autonomy fetch ...
'''
```

With this comment, the linter should ignore the code block:
```
<!-- LINTER_IGNORE -->
'''bash
cd <directory>
ls
autonomy fetch ...
'''
```


If no magic comment on a code block --> linter should fail (instead of ignore). This way we ensure that all ignored code blocks are indeed intended to be ignored.
```
'''bash
cd <directory>
ls
autonomy fetch ...
'''
```

DavidMinarsch (changes requested):
This snippets approach is a bit messy imo

DavidMinarsch:
We will take a different approach as suggested by @jmoreira-valory

PR #1487, opened 2022-10-28
[0.3.4]: Add tests for tools rounds

Adamantios:
## Proposed changes

Adds tests for `packages/valory/skills/abstract_round_abci/test_tools/rounds.py`.

## Fixes

n/a

## Types of changes

## Checklist

## Further comments

n/a


packages/valory/skills/abstract_round_abci/tests/test_tools/__init__.py
```diff
@@ -0,0 +1,20 @@
+# -*- coding: utf-8 -*-
+# ------------------------------------------------------------------------------
+#
+#   Copyright 2021-2022 Valory AG
```
Karrenbelt, line 4: copyright incorrect

packages/valory/skills/abstract_round_abci/tests/test_tools/test_rounds.py
```diff
@@ -0,0 +1,689 @@
+# -*- coding: utf-8 -*-
+# ------------------------------------------------------------------------------
+#
+#   Copyright 2021-2022 Valory AG
```
Karrenbelt, line 4: same

packages/valory/skills/abstract_round_abci/tests/test_tools/test_rounds.py
```diff
@@ -0,0 +1,689 @@
...
+        )
+
+    def __eq__(self, other: Any) -> bool:
+        """The method that will be used for the assertion comparisons."""
+        return (
+            self.expected._round_count == other._round_count
+            and self.expected.sender == other.sender
+            and self.expected._value == other._value
+            and self.expected._vote == other._vote
+        )
```
Karrenbelt, line 101: first instinct: shouldn't this method be present on the actual payload? Turns out `BaseTxPayload` compares the assigned `uuid4` so for sure no payload should even be considered the same, might as well use `is` instead of `==`. Good practice is to implement a type check and return (not raise) `NotImplemented` if not the same to delegate comparison to `other`.

https://github.com/valory-xyz/open-autonomy/blob/20551f16131c8de0144729f2422bcca91e2210b6/packages/valory/skills/abstract_round_abci/base.py#L263-L272

Do you think vote could also be called `value`? That would simplify matters. I know it's outside the scope of this PR, but added it as one of the notes on the [[WD] refactoring open-autonomy core packages](https://docs.google.com/document/d/1kEC-GctRTU7rOuJ7q18-t13OQQcAwN2QtLhd7-WUlTw). If you have any more ideas / considerations, please add them as well

packages/valory/skills/abstract_round_abci/tests/test_tools/test_rounds.py
```diff
@@ -0,0 +1,689 @@
...
+        for agent in sorted(participants)
+    ]
+
+    actual = get_dummy_tx_payloads(participants, value, vote, is_value_none)
+
+    assert len(actual) == len(expected) == len(participants)
+    assert all(
+        actual_payload == expected_payload
+        for actual_payload, expected_payload in zip(actual, expected)
+    )
```
Karrenbelt, line 134: these are lists, so can just use `==` on the list directly (and should also work with set, tuple, frozenset)

packages/valory/skills/abstract_round_abci/tests/test_tools/test_rounds.py
```diff
@@ -0,0 +1,689 @@
...
+    @staticmethod
+    @given(st.text(max_size=200), st.binary(max_size=500), st.booleans(), st.integers())
+    def test_properties(
+        sender: str, value: bytes, vote: bool, round_count: int
+    ) -> None:
+        """Test all the properties."""
+        dummy_tx_payload = DummyTxPayload(sender, value, vote, round_count)
+        assert dummy_tx_payload.value == value
+        assert dummy_tx_payload.vote == vote
+        assert dummy_tx_payload.data == {"value": value}
```
Karrenbelt, line 149: yeah so you see the issue here? If we have a voting payload with a vote instead of value, it won't be in the data. Which it actually should be. If we use "value" instead of "vote" for voting payloads that would be resolved. It's not like "value" is in use already on voting payloads. 

Note that this vote attribute is also not compared in the `__eq__` operator of the BaseTxPayload referenced above.

packages/valory/skills/abstract_round_abci/tests/test_tools/test_rounds.py
```diff
@@ -0,0 +1,689 @@
...
+
+
+class TestDummySynchronizedData:  # pylint: disable=too-few-public-methods
+    """Test class for `DummySynchronizedData`."""
+
+    @staticmethod
+    @given(st.lists(st.text(max_size=200), max_size=100))
+    def test_most_voted_keeper_address(
```
Karrenbelt, line 157: a class with a single method that is also a staticmethod; a stateless class. Should be a function. Using a minimal amount of classes and the bare minimum of state will make you a happier man.

Karrenbelt, line 157: a class with a single method that is also a staticmethod; a stateless class. Should be a function. Using a minimal amount of classes and the bare minimum of state; you will be a happier man.

packages/valory/skills/abstract_round_abci/tests/test_tools/test_rounds.py
```diff
@@ -0,0 +1,689 @@
...
+            return
+
+        assert (
+            dummy_synchronized_data.most_voted_keeper_address
+            == most_voted_keeper_address_data[-1]
+        )
+
+
+class TestBaseRoundTestClass:
+    """Test `BaseRoundTestClass`."""
```
Karrenbelt, line 189: a class with two static method. Lets do two functions.

packages/valory/skills/abstract_round_abci/tests/test_tools/test_rounds.py
```diff
@@ -0,0 +1,689 @@
...
+            DummyEvent
+        )
+        self.base_round_test.consensus_params._max_participants = (  # pylint: disable=protected-access
+            MAX_PARTICIPANTS
+        )
+
+
+class DummyCollectDifferentUntilAllRoundWithEndBlock(
+    DummyCollectDifferentUntilAllRound
+):
```
Karrenbelt, line 255: just a note: this should be moved into a directory with dummy data. I need it as well. We have such on open-aea, let's discuss.

packages/valory/skills/abstract_round_abci/tests/test_tools/test_rounds.py
```diff
@@ -0,0 +1,689 @@
...
+            cast(List[BaseTxPayload], round_payloads),
+            lambda synchronized_data, _: synchronized_data.update(
+                most_voted_keeper_address=[
+                    f"agent_{i}" for i in range(MAX_PARTICIPANTS)
+                ]
+            ),
+            synchronized_data_attr_checks,
+            exit_event,
+        )
+        for _ in range(4):
```
Karrenbelt, line 318: `MAX_PARTICIPANTS`

packages/valory/skills/abstract_round_abci/tests/test_tools/test_rounds.py
```diff
@@ -0,0 +1,689 @@
...
+        self.dummy_exit_event = dummy_exit_event
+
+    def end_block(self) -> Optional[Tuple[BaseSynchronizedData, Enum]]:
+        """A dummy `end_block` implementation."""
+        if self.collection_threshold_reached and self.dummy_exit_event is not None:
+            return (
+                cast(
+                    DummySynchronizedData,
+                    self.synchronized_data.update(
+                        most_voted_keeper_address=list(self.collection.keys())
```
Karrenbelt, line 270: probably doesn't matter, but this should be the value instead of the key.
`most_voted_keeper_address=set(self.collection.values()).pop()`

packages/valory/skills/abstract_round_abci/tests/test_tools/test_rounds.py
```diff
@@ -0,0 +1,689 @@
...
+        return None
+
+
+class TestBaseCollectDifferentUntilAllRoundTest(BaseTestBase):
+    """Test `BaseCollectDifferentUntilAllRoundTest`."""
+
+    base_round_test: BaseCollectDifferentUntilAllRoundTest
+    base_round_test_cls: Type[
+        BaseCollectDifferentUntilAllRoundTest
+    ] = BaseCollectDifferentUntilAllRoundTest
```
Karrenbelt, line 284: lets remove the type annotation here :smiling_face_with_tear:

packages/valory/skills/abstract_round_abci/tests/test_tools/test_rounds.py
```diff
@@ -0,0 +1,689 @@
...
+    def end_block(self) -> Optional[Tuple[BaseSynchronizedData, Enum]]:
+        """A dummy `end_block` implementation."""
+        if self.collection_threshold_reached:
+            return (
+                cast(
+                    DummySynchronizedData,
+                    self.synchronized_data.update(
+                        most_voted_keeper_address=self.common_payload
+                    ),
+                ),
```
Karrenbelt, line 340: can we introduce a helper function (with `**kwags` signature) for this on the base class for tests at least? it's a repeated all over the place (6x) and will make the code much more readable.

packages/valory/skills/abstract_round_abci/tests/test_tools/test_rounds.py
```diff
@@ -0,0 +1,689 @@
...
+            lambda synchronized_data, _: synchronized_data.update(
+                most_voted_keeper_address=most_voted_payload
+            ),
+            synchronized_data_attr_checks,
+            most_voted_payload,
+            exit_event,
+        )
+        for _ in range(4):
+            next(gen)
+        last_iteration(gen)
```
Karrenbelt, line 470: this is also repeated 5 times. Put into a function and place it on the base class

Karrenbelt (changes requested):
there is a lot of repeated code. Looks okay otherwise, but really needs to be cleaned up. I also did not repeat comments when it concerned code that will not exist once the implementation is sufficiently DRY. Please don't copy paste or if you do and you do then take the time to clean up after.

packages/valory/skills/abstract_round_abci/tests/test_tools/test_rounds.py
Adamantios: Refactored in 2502ecd562d7bb198078e9edb96444bdb505266e.

packages/valory/skills/abstract_round_abci/__init__.py
```diff
@@ -20,6 +20,11 @@
...
 
 from aea.configurations.base import PublicId
+from hypothesis import settings
 
 
 PUBLIC_ID = PublicId.from_str("valory/abstract_round_abci:0.1.0")
+CI = "CI"
+
+
+settings.register_profile(CI, deadline=5000)
```
Adamantios, line 30: We need a larger deadline on CI so it does not report flaky tests.

packages/valory/skills/abstract_round_abci/tests/test_tools/test_rounds.py
```diff
@@ -61,8 +63,9 @@
 # this is how many times we need to iterate before reaching the last iteration for a base test.
 BASE_TEST_GEN_ITERATIONS = 4
 
-
-settings(deadline=5000)
+running_on_ci = os.getenv(CI)
+if running_on_ci:
+    settings.load_profile(CI)
```
Adamantios, line 68: Using CI profile only on CI.

packages/valory/skills/abstract_round_abci/__init__.py
Karrenbelt: how does this recognize it's running on CI? on open-aea I introduced an env var for this in the [workflow.yaml](https://github.com/valory-xyz/open-aea/blob/ca2140f3bd21bc1cdbf20dad8f33d727e982bfff/.github/workflows/workflow.yml#L10-L11)

packages/valory/skills/abstract_round_abci/tests/test_tools/test_rounds.py
Karrenbelt: aah here it is. So where is the env var set?

packages/valory/skills/abstract_round_abci/tests/test_tools/test_rounds.py
Karrenbelt: exactly why it's added to the WD

packages/valory/skills/abstract_round_abci/tests/test_tools/test_rounds.py
Karrenbelt: I consider this non-pythonic; a class in not the equivalent of a namespace, even pylint complains. It is needless bloat, but in the end it's a matter of style; I hate code and classes, you don't, lets leave it at that.

packages/valory/skills/abstract_round_abci/tests/test_tools/test_rounds.py
```diff
@@ -0,0 +1,678 @@
...
+        self.base_round_test.consensus_params._max_participants = (  # pylint: disable=protected-access
+            MAX_PARTICIPANTS
+        )
+        test_round = DummyCollectDifferentUntilAllRoundWithEndBlock(
+            exit_event,
+            self.base_round_test.synchronized_data,
+            self.base_round_test.consensus_params,
+        )
+        round_payloads = [
+            (DummyTxPayload(f"agent_{i}", i)) for i in range(MAX_PARTICIPANTS)
```
Karrenbelt, line 319: needless extra packing
`[DummyTxPayload(f"agent_{i}", i) for i in range(MAX_PARTICIPANTS)]`

Karrenbelt (approved):

packages/valory/skills/abstract_round_abci/__init__.py
Adamantios: GH [sets it automatically](https://docs.github.com/en/actions/learn-github-actions/environment-variables#default-environment-variables). This should do the job, no?

packages/valory/skills/abstract_round_abci/tests/test_tools/test_rounds.py
Adamantios: See above.

packages/valory/skills/abstract_round_abci/__init__.py
Karrenbelt: very nice, I didn't know

Karrenbelt (approved):

Karrenbelt (approved):

Adamantios:
Framework test `FAILED tests/test_deployments/test_app.py::TestTendermintBufferWorking::test_tendermint_buffer`, which is not related to this PR.

DavidMinarsch (approved):
Test setup should not happen in base init. Please fix in a future PR.

PR #1488, opened 2022-10-28
[0.3.4]: Test/termination abci

0xArdi:
## Proposed changes

This PR adds the tests for `termination_abci`. 

## Fixes

## Types of changes

## Checklist

### DIFF
--- a/packages/valory/skills/termination_abci/behaviours.py
+++ b/packages/valory/skills/termination_abci/behaviours.py
@@ -171,7 +171,7 @@ class BackgroundBehaviour(BaseBehaviour):
         # a `RemovedOwner` event is thrown, which is what `_get_latest_removed_owner_event()` captures.
 
         termination_signal_occurrence = int(termination_signal["block_number"])
-        service_owner_removal_occurrence = int(termination_signal["block_number"])
+        service_owner_removal_occurrence = int(service_owner_removal["block_number"])
 
         # if the termination signal has occurred after the owner has been removed the service should terminate,
         # otherwise it's a signal that has already been handled previously
--- a/packages/valory/skills/termination_abci/behaviours.py
+++ b/packages/valory/skills/termination_abci/behaviours.py
@@ -403,10 +403,10 @@ class BackgroundBehaviour(BaseBehaviour):
         # we remove all but one safe owner
         for owner in safe_owners[1:]:
             # we generate a tx to remove the current owner
-            remove_tx = yield from self._get_remove_owner_tx(owner, threshold)
-            if remove_tx is None:
+            remove_tx_str = yield from self._get_remove_owner_tx(owner, threshold)
+            if remove_tx_str is None:
                 return None
-
+            remove_tx = bytes.fromhex(remove_tx_str)
             # we append it to the list of the multisend txs
             transactions.append(
                 {
--- a/packages/valory/skills/termination_abci/behaviours.py
+++ b/packages/valory/skills/termination_abci/behaviours.py
@@ -88,7 +88,6 @@ class BackgroundBehaviour(BaseBehaviour):
         if self._is_termination_majority():
             # if termination majority has already been reached
             # there is no need to run the rest of this method
-            yield
             return
 
         signal_present = yield from self.check_for_signal()
@@ -111,6 +110,8 @@ class BackgroundBehaviour(BaseBehaviour):
                 "Couldn't prepare multisend transaction for termination."
             )
             return
+
+        self.context.logger.info("Successfully prepared termination multisend tx.")
         termination_payload = BackgroundPayload(
             self.context.agent_address, background_data=background_data
         )
@@ -140,7 +141,7 @@ class BackgroundBehaviour(BaseBehaviour):
         :returns: True if the termination signal is found, false otherwise
         """
         if self._service_owner_address is None:
-            self._service_owner_address = yield from self.get_service_owner()
+            self._service_owner_address = yield from self._get_service_owner()
 
         termination_signal = yield from self._get_latest_termination_signal()
         if termination_signal is None:
@@ -236,7 +237,7 @@ class BackgroundBehaviour(BaseBehaviour):
 
         return latest_zero_transfer_event
 
-    def get_service_owner(self) -> Generator[None, None, Optional[str]]:
+    def _get_service_owner(self) -> Generator[None, None, Optional[str]]:
         """Method that returns the service owner."""
         response = yield from self.get_contract_api_response(
             performative=ContractApiMessage.Performative.GET_STATE,  # type: ignore
--- a/packages/valory/skills/termination_abci/behaviours.py
+++ b/packages/valory/skills/termination_abci/behaviours.py
@@ -244,7 +244,7 @@ class BackgroundBehaviour(BaseBehaviour):
             contract_id=str(ServiceRegistryContract.contract_id),
             contract_callable="get_service_owner",
             contract_address=self.params.service_registry_address,
-            service_id=self.params.service_id,
+            service_id=self.params.on_chain_service_id,
         )
 
         if response.performative != ContractApiMessage.Performative.STATE:
--- a/packages/valory/skills/termination_abci/behaviours.py
+++ b/packages/valory/skills/termination_abci/behaviours.py
@@ -441,10 +441,10 @@ class BackgroundBehaviour(BaseBehaviour):
             contract_callable="get_tx_data",
             multi_send_txs=transactions,
         )
-        if response.performative != ContractApiMessage.Performative.RAW_MESSAGE:
+        if response.performative != ContractApiMessage.Performative.RAW_TRANSACTION:
             self.context.logger.error(
                 f"Couldn't compile the multisend tx. "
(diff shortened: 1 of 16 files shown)


0xArdi:
> Is this finished for review?

@DavidMinarsch behaviour tests pending, will push in a bit.

packages/valory/skills/termination_abci/tests/test_behaviours.py
```diff
@@ -0,0 +1,547 @@
...
+    initial_data: Dict[str, Hashable]
+    ok_reqs: List[Callable]
+    err_reqs: List[Callable]
+    expected_logs: List[str]
+
+
+class BaseTerminationTest(FSMBehaviourBaseCase):
+    """Base test case."""
+
+    path_to_skill = Path(__file__).parent.parent
```
Adamantios, line 72:
```suggestion
    path_to_skill = Path(__file__).parents(1)
```

Adamantios (approved):
LGTM, though why skip `pylint` for entier test files (dialogues, handlers, models)?

packages/valory/skills/termination_abci/tests/test_behaviours.py
0xArdi: The pylint skip was actually copied over from another open-autonomy package, its probably a bit too much, and we can pin down the exact error we have to ignore.

packages/valory/skills/termination_abci/tests/test_behaviours.py
```diff
@@ -0,0 +1,547 @@
...
+                name="agent fails to get the safe owners",
+                initial_data=dict(safe_contract_address=SAFE_ADDRESS),
+                ok_reqs=[
+                    _mock_get_service_owner_request,
+                    _mock_get_zero_transfer_events_request,
+                    _mock_get_removed_owner_events_request,
+                ],
+                err_reqs=[_mock_get_owners_request],
+                expected_logs=[_SAFE_OWNERS_ERR_LOG],
+            ),
```
Karrenbelt, line 408: I like this a lot, context managers on the text case, it's exactly what I wanted :+1:

packages/valory/skills/termination_abci/tests/test_rounds.py
```diff
@@ -0,0 +1,122 @@
...
+    TerminationRound,
+)
+
+
+MAX_PARTICIPANTS: int = 4
+
+
+def get_participants() -> FrozenSet[str]:
+    """Participants"""
+    return frozenset([f"agent_{i}" for i in range(MAX_PARTICIPANTS)])
```
Karrenbelt, line 40: why is this a separate function?

packages/valory/skills/termination_abci/tests/test_rounds.py
```diff
@@ -0,0 +1,122 @@
...
+
+
+class BaseRoundTestClass:  # pylint: disable=too-few-public-methods
+    """Base test class for Rounds."""
+
+    synchronized_data: SynchronizedData
+    consensus_params: ConsensusParams
+    participants: FrozenSet[str]
+
+    @classmethod
```
Karrenbelt, line 50: should not be a class method

packages/valory/skills/termination_abci/tests/test_behaviours.py
Karrenbelt: you have some pytest skip in other files as well that seems unnecessary

packages/valory/skills/termination_abci/tests/test_behaviours.py
```diff
@@ -0,0 +1,547 @@
...
+                contract_address=SERVICE_REGISTRY_ADDRESS,
+            ),
+            response_kwargs=dict(
+                performative=response_performative,
+                state=State(
+                    ledger_id="ethereum",
+                    body=response_body,
+                ),
+            ),
+        )
```
Karrenbelt, line 187: these mock contract api calls are the only thing that is difficult to read and understand. Do a CTRL + F on `mock_contract_api_request` and skim over them hitting enter, you'll find 8 nearly identical code blocks. I suggest to introduce a helper function that only takes the variable arguments.

Karrenbelt, line 187: these mock contract api calls are the only thing that is difficult to read and understand. Do a CTRL + F to look for "`mock_contract_api_request`" and skim over them hitting enter, you'll find 8 nearly identical code blocks. I suggest to introduce a helper function that only takes the variable arguments.

packages/valory/skills/termination_abci/tests/test_rounds.py
```diff
@@ -0,0 +1,122 @@
...
+class TestTerminationRound(BaseRoundTestClass):
+    """Tests for TerminationRound."""
+
+    def test_run(
+        self,
+    ) -> None:
+        """Run tests."""
+
+        test_round = TerminationRound(
+            synchronized_data=deepcopy(self.synchronized_data),
```
Karrenbelt, line 118: same question regarding the use of deepcopy here.

packages/valory/skills/termination_abci/tests/test_behaviours.py
```diff
@@ -0,0 +1,547 @@
...
+            err_req(self, error=True)
+
+        # check that the expected logs appear
+        for expected_log in test_case.expected_logs:
+            assert expected_log in caplog.text
+
+        if len(test_case.err_reqs) == 0:
+            # no mocked requests fail,
+            # the behaviour should complete
+            self.complete()
```
Karrenbelt, line 501: is there a reason not to run test on failure with a pytest.raises? Can't you treat them just the same as the other context managers and put these on the test case?

packages/valory/skills/termination_abci/behaviours.py
```diff
@@ -402,10 +403,10 @@ def _get_multisend_tx(self) -> Generator[None, None, Optional[str]]:
         # we remove all but one safe owner
         for owner in safe_owners[1:]:
```
Karrenbelt, line 404: I did not study the logic here. Is the service owner always the first of the safe owners? Or can there be a situation where the first is not, and the first needs to transfer control to the owner. This would mean there is a moment that a single agent - operated by someone else than the service owner - is in control of the vault.

packages/valory/skills/termination_abci/tests/test_rounds.py
```diff
@@ -0,0 +1,122 @@
...
+class TestBackgroundRound(BaseRoundTestClass):
+    """Tests for BackgroundRound."""
+
+    def test_run(
+        self,
+    ) -> None:
+        """Run tests."""
+
+        test_round = BackgroundRound(
+            synchronized_data=deepcopy(self.synchronized_data),
```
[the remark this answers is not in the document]
0xArdi: So this test checks whether `synchronized_data` is written to correctly. If I don't copy here, `synchronized_data` will be changed by the round, and the check below will become meaningless because `expected_state` and `actual_state` point to the same instance of `synchronized_data`.
```
        assert (
            cast(SynchronizedData, actual_state).termination_majority_reached
            == cast(SynchronizedData, expected_state).termination_majority_reached
        )
```

packages/valory/skills/termination_abci/tests/test_behaviours.py
0xArdi: I understand, but this already is a util function, and I don't think it has a lot of obsolete params.

packages/valory/skills/termination_abci/tests/test_behaviours.py
0xArdi: We don't raise an exception here, we just log that there's an error and continue.

packages/valory/skills/termination_abci/behaviours.py
0xArdi: No, so what happens is we remove all owners but one, and then the last owner that is left we swap with the service owner.

packages/valory/skills/termination_abci/tests/test_rounds.py
0xArdi: It's a pattern we have all over the round tests. I assume its like this because it doesnt have any logic related to the test class, ie no `self` or `cls` use.

packages/valory/skills/termination_abci/tests/test_rounds.py
```diff
@@ -0,0 +1,122 @@
+# -*- coding: utf-8 -*-
+# ------------------------------------------------------------------------------
+#
+#   Copyright 2021-2022 Valory AG
```
Karrenbelt, line 4: 2022

packages/valory/skills/termination_abci/tests/test_rounds.py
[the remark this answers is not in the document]
Karrenbelt: that is not good and I expect this is the case in other tests then as well. There are 9 such occurrences in transaction settlement test_rounds.py alone, but given the convoluted implementation using generators I am not certain this is in fact an issue. Nonetheless seems worthy of investigation to me, what do you think @Adamantios ?

https://github.com/valory-xyz/open-autonomy/blob/20551f16131c8de0144729f2422bcca91e2210b6/packages/valory/skills/transaction_settlement_abci/tests/test_rounds.py#L244

DavidMinarsch (approved):

PR #1490, opened 2022-11-01
[ WIP]  Autonomy refactor 1 - analyse module

angrybayblade:
## Proposed changes

Refactors 

- Spec check
  - Merges `autonomy analyse abci check-app-specs` and `autonomy analyse generate-app-specs` into `autonomy analyse fsm-specs`
  - Extracts helper utils out of the top level command definition
  - Deprecated the `--infile` flag
  - Updates the app class name format, older format was `packages.author.skills.skill_name.rounds.SomeAbciApp` which should be just `SomeAbciApp` now
  - Adds schema validation

- Handler check
  - Moves command from `autonomy analyse abci check-handlers` to `autonomy analyse handlers`
  - Removes `--packages-dr` flag and utilises `--registry-path` for defining packages directory
  - Redefines the common handlers and ignore skills options to use multiple value click argument
    - Old format - `--common abci,http,contract_api,ledger_api,signing --skip abstract_abci,counter,counter_client,hello_world_abci`
    - New format `-h abci -h http -h contract_api -h ledger_api -h signing -i abstract_abci -i counter -i counter_client -i hello_world_abci`
  - Moves skip check out of the core logic to command logic
  
 Docstring check
 - Moves command from `autonomy analyse abci docstrings` to `autonomy analyse docstrings`
 - Extract core logic to core modules and separates implementations by concern

Log parser
- Moves the log parser from `autonomy analyse abci logs` to `autonomy analyse logs`

With this refactor we also deprecate the abci command group

## Fixes

https://github.com/valory-xyz/open-autonomy/issues/1450

## Types of changes

## Checklist


autonomy/cli/analyse.py
```diff
@@ -165,37 +168,47 @@ def parse_logs(file: Path) -> None:
...
+    help="Specify which skills to skip. Eg. -i skill_0 -i skill_1 -i skill_2",
+    multiple=True,
 )
-def run_handler_check(packages_dir: Path, skip: str, common: str) -> None:
+def run_handler_check(
+    ctx: Context, ignore: List[str], common_handlers: List[str]
+) -> None:
     """Check handler definitions."""
 
-    # TODO: use more appropriate input type and apply validation on it
```
angrybayblade, line 189: Addressed with the refactor

autonomy/cli/analyse.py
```diff
@@ -165,37 +168,47 @@ def parse_logs(file: Path) -> None:
...
+    "--common-handlers",
+    "-h",
     type=str,
-    default="abstract_abci",
-    help="Specify which skills to skip. Eg. skill_0,skill_1,skill_2",
+    default=[
+        "abci",
+    ],
+    help="Specify which handlers to check. Eg. -h handler_a -h handler_b -h handler_c",
+    multiple=True,
```
angrybayblade, line 181: Utilise multiple values for a click option to take multiple values for common handlers. So

```
autonomy analyse abci check-handlers --common abci,http,contract_api,ledger_api,signing --skip abstract_abci,counter,counter_client,hello_world_abci
```
becomes

```
autonomy analyse handlers -h abci -h http -h contract_api -h ledger_api -h signing -i abstract_abci -i counter -i counter_client -i hello_world_abci
```

autonomy/cli/analyse.py
```diff
@@ -165,37 +168,47 @@ def parse_logs(file: Path) -> None:
...
+    "--ignore",
+    "-i",
     type=str,
-    default="abci",
-    help="Specify which handlers to check. Eg. handler_a,handler_b,handler_c",
+    default=[
+        "abstract_abci",
+    ],
+    help="Specify which skills to skip. Eg. -i skill_0 -i skill_1 -i skill_2",
+    multiple=True,
```
angrybayblade, line 191: Same as above

autonomy/cli/analyse.py
```diff
@@ -165,37 +168,47 @@ def parse_logs(file: Path) -> None:
...
-    common_handlers = common.split(",")
-    packages_dir = Path(packages_dir)
-
     try:
-        for yaml_file in sorted(packages_dir.glob(f"**/{DEFAULT_SKILL_CONFIG_FILE}")):
+        for yaml_file in sorted(
+            Path(ctx.registry_path).glob(f"*/*/*/{DEFAULT_SKILL_CONFIG_FILE}")
+        ):
+            if yaml_file.parent.name in ignore:
+                click.echo(f"Skipping {yaml_file.parent}")
```
angrybayblade, line 203: Moved out the skip logic from core logic to here

tox.ini
```diff
@@ -405,7 +405,7 @@ skip_install = True
 deps =
     tomte[safety]==0.1.5
 commands =
-    safety check -i 37524 -i 38038 -i 37776 -i 38039 -i 39621 -i 40291 -i 39706 -i 41002
+    safety check -i 37524 -i 38038 -i 37776 -i 38039 -i 39621 -i 40291 -i 39706 -i 41002 -i 51499 -i 51358
```
angrybayblade, line 408: @DavidMinarsch safety check is failing with 

```
+============================+===========+==========================+==========+
| safety                     | 1.10.3    | <2.2.0                   | 51358    |
| wheel                      | 0.37.1    | <0.38.0                  | 51499    |
+==============================================================================+
```

I think we'll have to bump safety in the next tomte release

autonomy/cli/analyse.py
Karrenbelt: what is the advantage of this?

autonomy/cli/analyse.py
Karrenbelt: you reckon we should be using [`.resolve` instead of `.absolute`](https://github.com/python/cpython/blob/0e15c31c7e9907fdbe38a3f419b669fed5bb3b33/Lib/pathlib.py#L915-L950) as a default?

autonomy/cli/analyse.py
```diff
@@ -165,37 +168,47 @@ def parse_logs(file: Path) -> None:
...
+                continue
+
             click.echo(f"Checking {yaml_file.parent}")
-            check_handlers(yaml_file.resolve(), common_handlers, skip_skills)
-    except Exception as e:  # pylint: disable=broad-except
+            check_handlers(
+                yaml_file.resolve(),
+                common_handlers=common_handlers,
+            )
+    except (FileNotFoundError, ValueError, ImportError) as e:
```
Karrenbelt, line 211: you reckon we should be using [`.resolve` instead of `.absolute`](https://github.com/python/cpython/blob/0e15c31c7e9907fdbe38a3f419b669fed5bb3b33/Lib/pathlib.py#L915-L950) as a default?

Karrenbelt (changes requested):
failing test on macos needs to be addressed. Looks like an issue with path
```
FAILED tests/test_autonomy/test_cli/test_analyse/test_check_handlers.py::TestCheckHandlers::test_check_handlers
```

if such keep happening, we've seen it more than once already, may be worth implementing a small utility function to use that prevents this.

autonomy/cli/analyse.py
angrybayblade: The only advantage is this is more systematic and reduces chances of the user providing badly formatted input strings

autonomy/cli/analyse.py
```diff
@@ -165,37 +168,47 @@ def parse_logs(file: Path) -> None:
...
+@pass_ctx
 @click.option(
-    "--skip",
+    "--common-handlers",
+    "-h",
     type=str,
-    default="abstract_abci",
-    help="Specify which skills to skip. Eg. skill_0,skill_1,skill_2",
+    default=[
+        "abci",
```
DavidMinarsch, line 178: Why is this default?

angrybayblade: On an assumption that ABCI handler is going to be present in every skill, should we remove the defaults?

Karrenbelt: looks like a mutable default argument to me :grimacing:

autonomy/cli/analyse.py
```diff
@@ -165,37 +168,47 @@ def parse_logs(file: Path) -> None:
...
 @click.option(
-    "--common",
+    "--ignore",
+    "-i",
     type=str,
-    default="abci",
-    help="Specify which handlers to check. Eg. handler_a,handler_b,handler_c",
+    default=[
+        "abstract_abci",
+    ],
```
Karrenbelt, line 189: like I'm not well acquainted with click. But mutable default arguments give me the creeps

autonomy/cli/analyse.py
```diff
@@ -165,13 +160,13 @@ def docstrings(ctx: Context, update: bool) -> None:
             raise click.ClickException(str(e)) from e
 
 
-@abci_group.command(name="logs")
+@analyse_group.command(name="logs")
```
angrybayblade, line 163: Since we've moved all other subcommands out of the `abci` there is no point in just keeping one last command.

dvilelaf (dismissed):
I see there are several PRs merged into this one that have been already reviewed, so due to the PR size I've skipped tests. It looks good to me although I'm not familiar with some of this code. Some conflicts must be solved ATM.

jmoreira-valory (dismissed):
Given that the PR comes from a reviewed branches,I have just checked the cli command changes and seem OK (althought not totally convinced from the change in analyse :) Prob will suggest some changes to be considered in some next PR/release.

DavidMinarsch (approved):

PR #1495, opened 2022-11-01
Fix/termination

0xArdi:
## Fixes

This PR fixes various issues with the termination logic discovered while performing e2e tests.

## Types of changes

## Checklist

## Further comments


packages/valory/skills/termination_abci/behaviours.py
```diff
@@ -398,10 +403,13 @@ def _get_multisend_tx(self) -> Generator[None, None, Optional[str]]:
...
             return None
-
         owner_to_be_swapped = safe_owners[0]
         # we remove all but one safe owner
-        for owner in safe_owners[1:]:
+        # reverse the list to avoid errors when removing owners
+        # this is because the owners are stored as a linked list
+        # in the safe, hence the order is important
+        safe_owners = list(reversed(safe_owners[1:]))
+        for owner in safe_owners:
```
Karrenbelt, line 412: `collections.deque` is a pretty versatile data structure. It can serve as a stack, queue and as a linked list.

Adamantios: Since you are getting an `iterator` and immediately converting it to a `list`, it could also have been:

```python
safe_owners = safe_owners[:0:-1]
```

0xArdi: On a future release we will get rid of this reversal completely.

Adamantios (approved):

packages/valory/skills/termination_abci/behaviours.py
Karrenbelt: you mean you're going to convert the linked list into a regular array / stack right? that is a good idea, you might want to open an issue on this or a trello ticket so you do not forget - unless you intend to do it straight away of course.

0xArdi: We don't need to have a linked list, this reversal is a result of how we currently set [`prevOwner`](https://github.com/safe-global/safe-contracts/blob/v1.3.0/contracts/base/OwnerManager.sol#L71). When I did this fix it came across my mind, but I went with this because it was simpler.

DavidMinarsch (approved):

PR #1521, opened 2022-11-05
fix: improve consistency of check_abci_app_specs

### LINKED ISSUE #1451: Fix FSM check implementation
https://github.com/valory-xyz/open-autonomy/pull/1445/files#r998718292

marcofavorito:
## Proposed changes

After committing ec3068474ba256cd7ca9015cedba7b9ecc241062, The command `check-app-specs` also accepts a value to --packages-dir, which is not necessarily a subdirectory of the current working directory. Without the change, the command would have failed as the call to `importlib.import_module("packages...")` could not find the ${packages_dir} folder in the Python system path.

Another requirement is that the packages directory is named 'packages'. Again, otherwise, imports of modules whose root module is 'packages' would have failed. This limitation could be overcome by artificially populating `sys.modules` with key "packages" and value "{packages_dir}/__init__.py".

[8595fe1](https://github.com/valory-xyz/open-autonomy/pull/1521/commits/8595fe1aa926483e1b19851b9c2ef64463dcd617): Since FSM specification files can only be stored in packages skills, the glob pattern can be made more strict by only checking the spec files in skill packages, i.e. `*/skills/*/fsm_specification.yaml` instead of `**/fsm_specification.yaml`.

## Fixes

Fix #1451

## Types of changes

- [X] Breaking fix (breaking change which fixes an issue)

## Checklist

## Further comments

n/a


autonomy/cli/analyse.py
```diff
@@ -97,8 +97,23 @@ def check_abci_app_specs(
 ) -> None:
     """Check ABCI app specs."""
 
+    # The command expects 'packages_dir' to be named 'packages', so to make import statements to be compatible with
+    # the standard Python import machinery. The directory can be outside the working directory from which the command
+    # is executed.
```
marcofavorito, line 102: I'd have put those comments in the function docstring such that Click could show them in the command help message. However, Darglint complains if the argument documentation is not present; on the other hand, if we added the explanation of the documentation, this would appear in the command helper message, which does not make much sense.

autonomy/analyse/abci/app_spec.py
```diff
@@ -430,19 +429,22 @@ def check_one(informat: str, infile: str, classfqn: str) -> bool:
...
 
         did_not_match = []
         fsm_specifications = sorted(
             [
-                *packages_dir.glob(f"**/{DEFAULT_FSM_SPEC_YAML}"),
+                *packages_dir.glob(f"*/skills/*/{DEFAULT_FSM_SPEC_YAML}"),
             ]
         )
-        # TODO: fix this implementation:
-        # - above file names should be constant and explained in helper text
```
marcofavorito, line 442: - "file names should be constant" this is done even before this PR, right?

- "explained in helper text", done in 9ce34dfd2411f45ea201f16aa20e72ab9ce1438a (not in the helper text of the *CLI command* though)

autonomy/analyse/abci/app_spec.py
```diff
@@ -430,19 +429,22 @@ def check_one(informat: str, infile: str, classfqn: str) -> bool:
...
-        # - investigate: either a) we need to do this in the order of the inverse dependency tree, so that
-        # dependencies are already loaded. Otherwise dependencies might not be present. Or b)
-        # we need to ensure relative modules are loaded too (https://docs.python.org/3/library/importlib.html#importlib.import_module)?
```
marcofavorito, line 445: (a) does not hold because we are assuming (1) the package directory is named `packages` and (2) we add its parent directory to the `sys.path` at position `0`, hence giving the highest priority to the modules in `packages` when importing modules.

(b) should hold as a consequence of the above, since we rely on the standard Python import machinery.

autonomy/cli/analyse.py
```diff
@@ -97,8 +97,23 @@ def check_abci_app_specs(
...
+    # add parent directory to Python system path
+    # so to give priority to the import of packages modules
+    sys.path.insert(0, str(packages_dir.parent))
+
+    # make sure the package_dir is named "packages", otherwise
+    # the import of packages.* modules won't work
+    if packages_dir.name != PACKAGES:
+        raise click.ClickException(
+            f"packages directory {packages_dir} is not named '{PACKAGES}'"
+        )
```
marcofavorito, line 114: as I said in the main PR comment, this is a bit restrictive, but it makes things more manageable. It is easy to address by loading `packages_dir/__init__.py` and then `sys.modules["packages"] = <module packages_dir/__init__.py>`

autonomy/cli/analyse.py
```diff
@@ -97,8 +97,23 @@ def check_abci_app_specs(
...
     """Check ABCI app specs."""
 
+    # The command expects 'packages_dir' to be named 'packages', so to make import statements to be compatible with
+    # the standard Python import machinery. The directory can be outside the working directory from which the command
+    # is executed.
+    packages_dir = Path(packages_dir).absolute()
+
+    # add parent directory to Python system path
+    # so to give priority to the import of packages modules
+    sys.path.insert(0, str(packages_dir.parent))
```
Karrenbelt, line 107: This looks distasteful and I highly recommend against it. This will presumably affect imports in other modules that do not anticipate such wickedness. Lasting state changes should be avoided at all cost. Use a context manager. What's more is that I don't think it should be here; how is this a concern of the `check_abci_app_spec` functionality? It should not be.

marcofavorito: this is a Click command that is run in a separate Python interpreter, and all the execution paths die with `sys.exit` in any possible execution, so there are no dangerous side effects. But I will add the context manager here so as to be cleaner (I would not add it in `SpecCheck` methods as they should, and do, implicitly assume that the AEA modules in packages are importable using standard Python import machinery).

Karrenbelt: that is very useful to know, more so than the current comment in the code, I suggest exchanging current code comment with the first sentence of your reply.

DavidMinarsch (approved):

Issue #1363, opened 2022-09-19
Set minimum assertions on timeouts in the base skill


Adamantios:
We need to guard against too short timeout values. This was for example the reason for a failure addressed in #1361.

PR #1529, opened 2022-11-07
feat: set lower bound for `validate_timeout` in `transaction_settlement_abci`

0xArdi:
## Proposed changes
Addresses #1363 by adding a lower bound for the `validate_timeout` parameter. 

## Fixes

## Types of changes

## Checklist


packages/valory/skills/transaction_settlement_abci/models.py
```diff
@@ -37,6 +37,7 @@
 )
 
 
+_MINIMUM_VALIDATE_TIMEOUT = 300  # 5 minutes
```
DavidMinarsch, line 40: Hm that’s a bit aggressive? What about tests and test chains and faster chains?

0xArdi: That's just the timeout, if it finishes earlier it should be fine. But I see your point, as there are cases where it is not necessary to wait for this long to assume that something has gone wrong. Based on the default value we set `1205`, I don't this is too big.

packages/valory/skills/transaction_settlement_abci/tests/test_models.py
```diff
@@ -0,0 +1,66 @@
...
+        """Read the default config only once."""
+        skill_yaml = PACKAGE_DIR / "skill.yaml"
+        with open(skill_yaml, "r", encoding="utf-8") as skill_file:
+            skill = yaml.safe_load(skill_file)
+        self.default_config = skill["models"]["params"]["args"]
+
+    def test_ensure_validate_timeout(  # pylint: disable=no-self-use
+        self,
+    ) -> None:
+        """Test that validate applies increases `validate_timeout` when it's lower than the allowed minimum."""
```
Adamantios, line 52: The docstring is a bit confusing. How about something like:

```suggestion
        """Test that `ensure_validate_timeout` raises when `validate_timeout` is lower than the allowed minimum."""
```

Adamantios (dismissed):

Adamantios (approved):

DavidMinarsch (approved):

PR #1545, opened 2022-11-11
fix/ check tendermint communication

0xArdi:
## Fixes

This PR fixes the bug on showcased on #1554. 

## Types of changes

## Checklist

## Further comments


packages/valory/skills/abstract_round_abci/behaviours.py
```diff
@@ -247,8 +249,15 @@ def teardown(self) -> None:
...
-        self._process_current_round()
+        if not self.tm_manager.is_tm_unhealthy or self.tm_manager.is_acting:
+            # tendermint is not healthy, or we are already applying a fix.
+            # try_fix() internally uses generators, that's why it's relevant
+            # to know whether a fix is already being applied.
+            # It might happen that tendermint is healthy, but the fix is not yet finished.
+            self.tm_manager.try_fix()
+            return
 
+        self._process_current_round()
```
0xArdi, line 260: the check now is concurrently with the operation of the behaviour.

0xArdi: This check ensures that the behaviours are run when tendermint is healthy. I couldn't think of a reason why we would want to have it otherwise.

packages/valory/skills/abstract_round_abci/behaviour_utils.py
```diff
@@ -1795,6 +1714,130 @@ def reset_tendermint_with_wait(
         return True
 
 
+class TmManager(BaseBehaviour, ABC):
```
0xArdi, line 1717: TmManager inherits from BaseBehaviour because it was easier to do so, not because it semantically makes sense. We could get rid of the inheritances given that we can access `BaseBehaviour.reset_tendermint_with_wait()` in a different way.

0xArdi: Also more importantly, the rest of the utils related to making requests etc.

DavidMinarsch (approved):
LGTM - would like to hear what @marcofavorito and @Adamantios think

Adamantios (approved):
LGTM. Do you think we could make this a background behaviour in the future, like the termination one?

0xArdi:
@Adamantios so this is like the termination round, in the sense that it runs concurrently with the behaviours, that is unless a fix (hard reset) is being performed.

PR #1551, opened 2022-11-14
Add error logging at the `ApiSpecs`

Adamantios:
## Proposed changes

Adds error logging at the ApiSpecs for when we fail to parse the response.

## Fixes

n/a

## Types of changes

## Checklist

## Further comments

n/a


packages/valory/skills/abstract_round_abci/models.py
```diff
@@ -277,8 +283,20 @@ def _get_response_data(
...
+            self.context.logger.error(
+                f"Could not access response using the given key(s) ({self.response_key})!"
+            )
+        except IndexError:
+            self.context.logger.error(
+                f"Could not access response using the given key(s) ({self.response_key}) "
+                f"and index ({self.response_index})!"
+            )
+
+        self._log_response(decoded_response)
```
0xArdi, line 298: This will log even when the response is not a valid JSON, was that intended?

Adamantios: Yes, it was, so that the users have all the information they need to fix their configuration or understand why the response could not be accessed.

DavidMinarsch (approved):

Issue #540, opened 2022-02-17
Replace loop in `CheckTxHistoryBehaviour`


Adamantios:
Remove loop in `CheckTxHistoryBehaviour`. Behaviors are already in a loop.
Implement the loop in a similar way as `SynchronizeLateMessagesBehaviour` does: 7d536e83ba38f2a4ff71447e51e2df064f41e28b

PR #1562, opened 2022-11-16
fix: add `yield` statement to loop in `CheckTransactionHistoryBehaviour`

0xArdi:
## Fixes

Addresses #540 by adding a yield statement in the `_check_tx_history` loop.  More details below.

## Types of changes

## Checklist

## Further comments

Our (FSM) behaviours should support concurrency, which is why we have `async_act` be a generator. 
`async_act` should never block, as it would starve the rest of the `AbstractRoundBehaviour` it belongs to. In CheckTransactionHistoryBehaviour, we have a loop which could potentially have a lot of elements in it, which can lead to blocking. To avoid this, we add a `yield` statement on every iteration of the loop, this ensures that the loop will be executed in a concurrent-friendly way. 
Note that this mechanism is already part of BaseBehaviour, ex.: [AsyncBehaviour.wait_for_condition](https://github.com/valory-xyz/open-autonomy/blob/v0.3.5/packages/valory/skills/abstract_round_abci/behaviour_utils.py#L197):
```python
    @classmethod
    def wait_for_condition(
        cls, condition: Callable[[], bool], timeout: Optional[float] = None
    ) -> Generator[None, None, None]:
        """Wait for a condition to happen.
        This is a local method that does not depend on the global clock,
        so the usage of datetime.now() is acceptable here.
        :param condition: the condition to wait for
        :param timeout: the maximum amount of time to wait
        :yield: None
        """
        if timeout is not None:
            deadline = datetime.datetime.now() + datetime.timedelta(0, timeout)
        else:
            deadline = datetime.datetime.max

        while not condition():
            if timeout is not None and datetime.datetime.now() > deadline:
                raise TimeoutException()
            yield
```


0xArdi:
The e2e failure is caused by 
```
2022-11-16T12:48:19.4253003Z [85495]stdout>packages.valory.skills.abstract_round_abci.base.ABCIAppInternalError: internal error: last timestamp is None
```
on
```
tests/test_registration.py::TestRegistrationStartUpFourAgentsCatchUp::test_run[4]
```
That agent doesnt depend on the changed package here, so the failure is unrelated.

Adamantios (approved):

DavidMinarsch (approved):

PR #1563, opened 2022-11-17
Bump open-aea to v1.24.0

angrybayblade:
## Proposed changes

## Fixes

## Types of changes

## Checklist


tests/test_docs/test_commands.py
```diff
@@ -49,16 +49,17 @@ def get_cmd_data(cmd: Union[Command, Group]) -> Dict:
...
-def get_group_tree(cmd: click.Group) -> Dict:
+def get_group_tree(cmd: Union[click.Group, click.Command]) -> Dict:
     """Returns a tree containing the command data."""
     tree = get_cmd_data(cmd)
+    ctx = click.Context(command=cmd)
 
     if isinstance(cmd, click.Group):
 
-        for sub_cmd_name in cmd.list_commands(click.Context):
+        for sub_cmd_name in cast(click.Group, cmd).list_commands(ctx=ctx):
```
angrybayblade, line 59: This was causing the mypy to fail after recreating the env.

tests/test_autonomy/test_cli/test_push_all.py
```diff
@@ -55,8 +56,9 @@ def test_run(
         """Test run."""
 
         # packages/<author>/<component>/<name>/<config>.yaml
-        path = f"{PACKAGES}/*/*/*/*.yaml"
-        packages = {file.parent for file in self.t.glob(path)}
+        available_packages = len(
+            PackageManagerV1.from_dir(ROOT_DIR / PACKAGES).dev_packages
+        )
         published_packages = []
```
Karrenbelt, line 62: why not `PackageManagerV1.from_dir(self.packages_dir).dev_packages`? We checked the temporary directory before, now we check the original packages directory instead. I don't think it makes a difference.

packages/packages.json
```diff
@@ -2,38 +2,38 @@
...
-        "service/valory/hello_world/0.1.0": "bafybeica6d3y6mb7lgv7wvr6dj2yidflmevd7opu3rjc3hgdq4ni7nqf2i",
-        "service/valory/register_reset/0.1.0": "bafybeib7gdxx76dv2hszzclmyamff2qjjgr3smdyp6islsvnlpwtb6wogi",
-        "service/valory/simple_abci/0.1.0": "bafybeihrbix7zrbv4yvjjdcpmzsupdov5ffgoq7a2ld3uj3gpb5nj6hl24",
-        "skill/valory/safe_deployment_abci/0.1.0": "bafybeiedso32a73rq37y24jaaiezvrzic4dmt4zghfy5pqzdwkhaw2rzzu"
+        "service/valory/hello_world/0.1.0": "bafybeie5dicy4addou6fhiiag7scueieem55dux62vdjrybo2b5nosmuze",
+        "service/valory/register_reset/0.1.0": "bafybeifgm3blejm257kx2kjarjridhgnrolo7skxakfuihcxw5bjy54xri",
+        "service/valory/simple_abci/0.1.0": "bafybeianli7tnsxsui67tzubger5bbwk24i5kuvg7arxs4fdqtyghhy3l4",
+        "skill/valory/safe_deployment_abci/0.1.0": "bafybeigsjva3vdywnzbwnwvbbvuzx5che4xagtkldkvux2t4bzr6mjjyma"
     },
     "third_party": {
```
DavidMinarsch, line 38: Did none of them change during the last release?

tests/test_autonomy/test_cli/test_push_all.py
angrybayblade: Honestly this won't make a difference because we're dealing with `packages.json` which is going to be same in both root dir and temp dir, but I'll keep this in mind for future

packages/packages.json
angrybayblade: They did https://github.com/valory-xyz/open-autonomy/pull/1534/files#diff-e459cec4c3e84b7f458724acf6256afb5eb4d16d0fe4f9ff97db9a57e95e8cbbR2

Adamantios (dismissed):

Karrenbelt (dismissed):

DavidMinarsch (approved):

PR #1566, opened 2022-11-17
Extend service component

angrybayblade:
## Proposed changes

This PR adds tests for the service config loader to show it's not very well generalised for various types of overrides

Breaking changes

- After the merge of this PR, the service overrides will require an update to the latest override format

## Types of changes

## Checklist

## Further comments


tests/data/dummy_service_config_files/service_2.yaml
```diff
@@ -0,0 +1,46 @@
...
+public_id: valory/dummy_skill:0.1.0
+type: skill
+0:
+  models:
+    params:
+      args:
+        message: ${str:"Hello, World!"}
+    benchmark_tool:
+      args:
+        log_dir: /logs
```
angrybayblade, line 22: The old format does not make sense at all

```yaml
benchmark_persistence_params:
  args: &id001
    log_dir: /benchmarks
public_id: valory/hello_world_abci:0.1.0
type: skill
models:
  0:
  - params:
      args:
        hello_world_message: ${SERVICE_HELLO_WORLD_HELLO_WORLD_STRING:str:HELLO_WORLD!}
  - benchmark_tool:
      args: *id001
```

So this is the new proposed format for multiple overrides

```yaml
public_id: {public_id}
type: {component_type}
{agent_index}:
    {overrides}
```

tests/data/dummy_service_config_files/service_3.yaml
```diff
@@ -0,0 +1,20 @@
...
+number_of_agents: 4
+---
+public_id: valory/dummy_connection:0.1.0
+type: connection
+config:
+  nodes:
+  - uri: some_uri
+    public_key: some_key
+  - uri: some_uri
+    public_key: some_key
```
angrybayblade, line 20: The current implementation also fails when there are `list` overrides, it cannot generate env vars for such cases

tests/data/dummy_service_config_files/service_4.yaml
```diff
@@ -0,0 +1,23 @@
...
+  README.md: bafybeiapubcoersqnsnh3acia5hd7otzt7kjxekr6gkbrlumv6tkajl6jm
+fingerprint_ignore_patterns: []
+agent: valory/hello_world:0.1.0:bafybeiaotnukv7oq2sknot73a4zssrrnjezh6nd2fwptrznxtnovy2rusm
+number_of_agents: 4
+---
+benchmark_persistence_params:
+  args: &id001
+    log_dir: /benchmarks
+public_id: valory/hello_world_abci:0.1.0
+type: skill
```
angrybayblade, line 17: This should fail the validation because `benchmark_persistence_params` is extra parameter, but it does not.

tests/data/dummy_service_config_files/service_2.yaml
Karrenbelt: on a side note: agent-specific overwrite are necessary but carry additional risk. We may consider adding a check that ensures all agent overwrite the same set of items.

tests/test_autonomy/test_configurations.py
```diff
@@ -139,3 +144,27 @@ def teardown(
...
+def test_load_service(service_file: Path) -> None:
+    """Test loading and processing service component."""
+
+    with open_file(service_file, "r", encoding="utf-8") as fp:
+        data = yaml_load_all(fp)
+
+    data = cast(List[Dict], apply_env_variables(data, env_variables=os.environ))
+    service_config, *overrides = data
+    Service.validate_config_data(service_config)
+    service_config["license_"] = service_config.pop("license")
```
Karrenbelt, line 167: this is somehow needed because it's a reserved variable name? :confused:

tests/test_autonomy/test_configurations.py
```diff
@@ -139,3 +144,27 @@ def teardown(
...
+        "dummy_service_config_files",
+    ).iterdir(),
+)
+def test_load_service(service_file: Path) -> None:
+    """Test loading and processing service component."""
+
+    with open_file(service_file, "r", encoding="utf-8") as fp:
+        data = yaml_load_all(fp)
+
+    data = cast(List[Dict], apply_env_variables(data, env_variables=os.environ))
```
Karrenbelt, line 164: seeing this makes me wonder: how do we ensure we clean up environmental variables after a test?

tests/test_autonomy/test_configurations.py
angrybayblade: Yes

tests/test_autonomy/test_configurations.py
angrybayblade: We don't need to, `apply_env_variables ` does not change anything in the `os.environ` neither we do for any kind of change

Adamantios (dismissed):

tests/data/dummy_service_config_files/service_2.yaml
DavidMinarsch: Imo that's too restrictive.

8ball030: We still require the templates for the overrides to be defined in the agent spec. 
The proposed changes are for the parsing of the service spec & its format

0xArdi (dismissed):

docs/guides/service_level_overrides.md
```diff
@@ -0,0 +1,119 @@
...
+---
+public_id: valory/hello_world_abci:0.1.0
+type: skill
+public_id: valory/hello_world_abci:0.1.0
+type: skill
+0:
+  models:
+    params:
+      args:
+        hello_world_message: Hello, from agent 0
```
Karrenbelt, line 58: message is purposefully different from here from the `service.yaml` I suppose?
https://github.com/valory-xyz/open-autonomy/blob/8282f1542f3c2a972de4c050d1bd2d312a71a624/packages/valory/services/hello_world/service.yaml#L23

Karrenbelt (approved):
LGTM

docs/guides/service_level_overrides.md
jmoreira-valory: Should it also be defined with an environment variable per agent? e.g.:
${SERVICE_HELLO_WORLD_HELLO_WORLD_STRING**_1**:str:HELLO_WORLD**_1**!}

packages/valory/services/hello_world/service.yaml
```diff
@@ -7,15 +7,43 @@ license: Apache-2.0
...
+  benchmark_persistence_params:
+    args: &id001
+      log_dir: /benchmarks
 public_id: valory/hello_world_abci:0.1.0
 type: skill
-models:
-  params:
-    args:
-      hello_world_message: ${SERVICE_HELLO_WORLD_HELLO_WORLD_STRING:str:HELLO_WORLD!}
+0:
```
jmoreira-valory, line 19: Why not use the "compact" form here if all the agents have the same parameters?

docs/guides/service_configuration_file.md
```diff
@@ -0,0 +1,304 @@
...
+    args: &id001
+      log_dir: /benchmarks
+public_id: valory/hello_world_abci:0.1.0
+type: skill
+models:
+  params:
+    args:
+      list_with_mappings:
+        - key: value_0
+        - key: value_1
```
jmoreira-valory, line 281: Can we have the same key twice??

docs/guides/service_configuration_file.md
```diff
@@ -0,0 +1,304 @@
...
+      args:
+        hello_world_message: Hello, from agent 3
+    benchmark_tool:
+      args: *id001
+```
+
+
+### Overrides with complicated data structures
+
+If you have nested list the environment export rules will differ as per the inner data structures. If the list is a strict list, meaning the inner values are either primitive data types like `int`, `str`, `float` or `boolean`, the list will be exported as a `json` string.
```
jmoreira-valory, line 231: First time export is mentioned in the documentation. It should be mentioned what the "export" means before explaining the difference between exporting one data type or another.

docs/guides/service_configuration_file.md
angrybayblade: It's a list of keys, interpret it as `"list_with_mappings: [{"key": "value_0"}, {"key": "value_1"}]"`

docs/guides/service_configuration_file.md
angrybayblade: `export ` is a pretty common word, I don't think we need to define what environment exports are

Karrenbelt: I think it is clear what export to environment means yes, the fact that this is happening in itself is not obvious though and is mentioned nowhere else in the documentation.

docs/guides/service_configuration_file.md
jmoreira-valory: So this means that a dictionary {"key": "value_0"} will be passed to Agent_0, a dictionary  {"key": "value_1"} will be passed to Agent_1... and so on?

jmoreira-valory (approved):
There are a couple of minor comments. Since it is a big PR, it is sensible not to delay merging it. @angrybayblade you can open a ticket or an issue to address the comments.

docs/guides/service_configuration_file.md
angrybayblade: No, they will be exported for the same agent but since they are a part of the same list they'll be exported with different index

docs/guides/service_configuration_file.md
angrybayblade: @jmoreira-valory I suggest you open a ticket to add section on how env var export and parsing works, we can have a quick call and I'll explain the process, then you can add the section

DavidMinarsch (approved):

PR #1568, opened 2022-11-18
[0.4.1] Test: `SynchronizedData` type during `AbciApp` instantiation

Karrenbelt:
## The issue

Test showcasing that the first time we instantiate the AbciApp we do not use the custom `SynchronizedData` class. To illustrate, in skills where we have the following situation:
https://github.com/valory-xyz/open-autonomy/blob/6229547d8d613d6688b37d707870c05d3e4759ab/packages/valory/skills/hello_world_abci/rounds.py#L70-L76
https://github.com/valory-xyz/open-autonomy/blob/6229547d8d613d6688b37d707870c05d3e4759ab/packages/valory/skills/simple_abci/rounds.py#L88-L94

During the first round of the `AbciApp` when we access the `.synchronized_data` property we get a `BaseSynchronizedData` returned and hence the type casting is misleading. Indeed the Synchronized data is only updated after the round is done, however we should cast it to the correct `SynchronizedData` type when we initialize the `AbciApp`.

## Proposed solution

This class can be provided as a class attribute on the `BaseRound` of a skill as `synchronized_data_class`. For example here:

https://github.com/valory-xyz/open-autonomy/blob/6229547d8d613d6688b37d707870c05d3e4759ab/packages/valory/skills/transaction_settlement_abci/rounds.py#L213
https://github.com/valory-xyz/open-autonomy/blob/6229547d8d613d6688b37d707870c05d3e4759ab/packages/valory/skills/safe_deployment_abci/rounds.py#L89

This can be used to optionally cast the `synchronized_data` during `AbciApp` initialization. In case `synchronized_data_class` is not present we will not cast the type and log a warning message.

NOTE: We should consider chained AbciApp's as well - which will be addressed in a subsequent PR.

## Types of changes

What types of changes does your code introduce? (A **breaking change** is a fix or feature that would cause existing functionality and APIs to not work as expected.)
_Put an `x` in the box that applies_

- [ ] Non-breaking fix (non-breaking change which fixes an issue)
- [ ] Breaking fix (breaking change which fixes an issue)
- [ ] Non-breaking feature (non-breaking change which adds functionality)
- [ ] Breaking feature (breaking change which adds functionality)
- [ ] Refactor (non-breaking change which changes implementation)
- [ ] Messy (mixture of the above - requires explanation!)

## Checklist

_Put an `x` in the boxes that apply._

- [ ] I have read the [CONTRIBUTING](../blob/main/CONTRIBUTING.md) doc
[...]

Issue #1564, opened 2022-11-17
Only `BaseSynchronizedData` can be used during the first round of an FSM


Adamantios:
More info about the issue: https://github.com/valory-xyz/apy-oracle/pull/15#discussion_r1024266125

PR #1569, opened 2022-11-18
[0.4.1] Fix: `SynchronizedData` type during `AbciApp` instantiation

Karrenbelt:
## Proposed changes

addressing #1564 

Minimally invasive solution to the problem where the first round of each AbciApp accesses a synchronized_data instance that is not of the type they may have specified in that skill - see base PR #1568 

## Fixes

## Types of changes

## Checklist

## Further comments


packages/valory/skills/abstract_round_abci/base.py
```diff
@@ -1871,6 +1871,13 @@ def __init__(
         logger: logging.Logger,
     ):
         """Initialize the AbciApp."""
+
+        if not hasattr(self.initial_round_cls, "synchronized_data_class"):
+            logger.warning(f"No `synchronized_data_class` set on {self}")
+        else:
+            synchronized_data_class = self.initial_round_cls.synchronized_data_class  # type: ignore
+            synchronized_data = synchronized_data_class(db=synchronized_data.db)
+
```
Karrenbelt, line 1880: This is I believe the minimally invasive solution that allows casting to the correct type on initialization of the `AbciApp`. This does not yet deal with chained applications, which requires a separate test and solution only worth implementing if this solution is approved.

packages/valory/skills/abstract_round_abci/tests/test_base.py
```diff
@@ -2230,6 +2231,12 @@ def dummy_attr(self) -> object:
     # this is how it's setup in SharedState.setup, using BaseSynchronizedData
     synchronized_data = BaseSynchronizedData(db=AbciAppDB(setup_data={}))
 
+    with caplog.at_level(logging.WARNING):
+        abci_app = AbciAppTest(synchronized_data, MagicMock(), logging)  # type: ignore
+        expected = f"No `synchronized_data_class` set on {abci_app}"
+        assert expected in caplog.text
+        assert not isinstance(abci_app.synchronized_data, SynchronizedData)
+
```
Karrenbelt, line 2239: some additional test for the logging

packages/valory/skills/abstract_round_abci/base.py
```diff
@@ -1871,6 +1871,13 @@ def __init__(
         logger: logging.Logger,
     ):
         """Initialize the AbciApp."""
+
+        if not hasattr(self.initial_round_cls, "synchronized_data_class"):
+            logger.warning(f"No `synchronized_data_class` set on {self}")
```
DavidMinarsch, line 1876: why are we not raising?

packages/valory/skills/abstract_round_abci/base.py
DavidMinarsch: Why do you think it needs a different implementation? 

It certainly needs a test.

packages/valory/skills/abstract_round_abci/base.py
Karrenbelt: We could but that would break things. Many skills don't use this (e.g. `registration_abci`, `reset_pause_abci`). Also this can be abstracted away entirely when we get to the refactoring, it has been added to that list on @Adamantios 's request.

Adamantios (approved):
LGTM, let's extend this solution to work for the chained apps as well.

DavidMinarsch (approved):

PR #1575, opened 2022-11-22
[WIP] Tendermint node._stop_tm_process

Karrenbelt:
## Proposed changes

This addresses the failure on windows (seen e.g. [here](https://github.com/valory-xyz/open-autonomy/actions/runs/3517814475/jobs/5896197811)) w.r.t.
`tests/test_deployments/test_app.py::TestTendermintHardResetServer::test_hard_reset[True]`

## Fixes

## Types of changes

## Checklist

## Further comments


deployments/Dockerfiles/tendermint/tendermint.py
```diff
@@ -194,22 +197,25 @@ def _start_monitoring_thread(self) -> None:
...
+        send_control_c(process)
+        try:
+            process.wait(timeout=TERMINATION_TIMEOUT)
+        except subprocess.TimeoutExpired:
+            if platform.system() == "Windows":
+                process.send_signal(signal.CTRL_BREAK_EVENT)  # type: ignore  # pylint: disable=no-member
+                process.wait(timeout=TERMINATION_TIMEOUT)
+            else:
+                process.terminate()
+                process.wait(timeout=TERMINATION_TIMEOUT)
```
Karrenbelt, line 218: specifically the wait after sending the `signal.CTRL_BREAK_EVENT` should address the issue with the hard reset on Windows

Karrenbelt:
The good news is that `test_hard_reset` in no longer flaky on Windows
```
FAILED tests/test_deployments/test_app.py::TestTendermintHardResetServer::test_hard_reset[False]
FAILED tests/test_deployments/test_app.py::TestTendermintBufferFailing::test_tendermint_buffer
ERROR tests/test_deployments/test_app.py::TestTendermintHardResetServer::test_hard_reset[False]
```
we may increase timeout, or opt for a kill signal. The question we need to address first though is how this may affect the performance of out applications.

Karrenbelt:
I have temporarily reverted the change to ascertain that this is in fact the issue - e414abd05fcb9225884a028514601e6d556919d5

WIP

deployments/Dockerfiles/tendermint/tendermint.py
```diff
@@ -194,22 +197,24 @@ def _start_monitoring_thread(self) -> None:
...
-        else:
-            self._process.send_signal(signal.SIGTERM)
-            self._process.wait(timeout=5)
-            poll = self._process.poll()
-            if poll is None:  # pragma: nocover
-                self._process.terminate()
-                self._process.wait(3)
+        process.poll()
+        if process.returncode is not None:
+            return
```
angrybayblade, line 207: Why polling before even calling a function to kill the subprocess?

deployments/Dockerfiles/tendermint/tendermint.py
```diff
@@ -29,11 +29,14 @@
...
 from typing import Any, Dict, List, Optional
 
+from aea.helpers.base import send_control_c
+
 
 ENCODING = "utf-8"
 DEFAULT_P2P_LISTEN_ADDRESS = "tcp://0.0.0.0:26656"
 DEFAULT_RPC_LISTEN_ADDRESS = "tcp://0.0.0.0:26657"
 DEFAULT_TENDERMINT_LOG_FILE = "tendermint.log"
+TERMINATION_TIMEOUT = 5
```
angrybayblade, line 39: This should a default value and timeout should be configurable

deployments/Dockerfiles/tendermint/tendermint.py
Karrenbelt: good practice, this function could be called more than once (not in current setup though I believe), and it makes it facilitates readability

Karrenbelt:
My suggestion is to use `terminate` also on Windows and go for insta-kill. We do not care about the Tendermint data anymore at this point as we perform a hard reset.

Karrenbelt:
- the issue with e2e tests is that `aea` module is not available on docker image.
- I suggest we close this PR since we have no idea what we're doing. There are the following options:
    1. we do nothing
    2. we add increased timeout, but we have no idea what that will do, certainly not in e2e scenario on windows
    3. we use .terminate() on Windows as well. Again we have no idea what that will do on Windows

Adamantios:
> * The issue with e2e tests is that `aea` module is not available on docker image.
> * I suggest we close this PR since we have no idea what we're doing. There are the following options:
>   
>   1. we do nothing
>   2. we add increased timeout, but we have no idea what that will do, certainly not in e2e scenario on windows
>   3. we use .terminate() on Windows as well. Again we have no idea what that will do on Windows

I'd say let's not risk breaking things at this moment.

PR #1576, opened 2022-11-23
[WIP] Fix service component and propose new override format

angrybayblade:
## Proposed changes

This PR introduces more flexible approach to defining the service overrides

## Types of changes

## Checklist

## Further comments


autonomy/configurations/base.py
```diff
@@ -207,179 +199,146 @@ def validate_config_data(
...
+        return configuration, component_id, overide_type
 
-    def process_component_section(
+    def process_component_overrides(
         self,
-        component_index: int,
+        agent_idx: int,
         component_configuration_json: Dict,
-    ) -> Tuple[ComponentId, Dict]:
+    ) -> Dict:
```
angrybayblade, line 258: This should only return processed metadata

autonomy/configurations/base.py
```diff
@@ -207,179 +199,146 @@ def validate_config_data(
...
+        if override_type == OverrideType.MULTIPLE:
+            configuration = configuration.get(agent_idx, {})
+            if configuration == {}:
+                raise ValueError(
+                    f"Overrides not provided for agent {agent_idx}; component={component_id}"
                 )
-        return component_id, overrides
+        configuration = apply_env_variables(
+            data=configuration, env_variables=os.environ.copy()
+        )
```
angrybayblade, line 279: Apply env vars by default

autonomy/configurations/base.py
```diff
@@ -207,179 +199,146 @@ def validate_config_data(
...
+                    cls._generate_env_vars_recursively(
+                        data=value,
+                        export_path=_export_path,
                     )
-        return overrides
+                )
+            elif isinstance(value, list):
+                env_var_dict[
+                    export_path_to_string(export_path=export_path)
+                ] = json.dumps(value)
```
angrybayblade, line 307: https://github.com/valory-xyz/open-autonomy/pull/1566#discussion_r1028932199 , export the list overrides explicitly

autonomy/configurations/base.py
```diff
@@ -55,25 +57,16 @@
 }
 
 
-def recurse(obj: Dict[str, Any]) -> Dict[str, Any]:
```
angrybayblade, line 58: Merged into service component.

autonomy/configurations/base.py
```diff
@@ -55,25 +57,16 @@
...
-            for k2, v2 in v.items():
-                new_obj["_".join([str(k), str(k2)])] = v2
-        else:
-            new_obj[k] = v
-    return recurse(new_obj)
 
+class OverrideType(Enum):
+    """Override types."""
 
-def _parse_nested_override(env_var_name: str, nested_override_value: Dict) -> Dict:
```
angrybayblade, line 73: Same as above

autonomy/configurations/base.py
```diff
@@ -207,179 +199,153 @@ def validate_config_data(
...
-            component_configurations[component_id] = component_configuration_json
+            if override_type == OverrideType.MULTIPLE:
+                for idx in range(self.number_of_agents):
+                    try:
+                        _configuration = cast(Dict, configuration[idx])
+                    except KeyError as e:
+                        raise AEAValidationError(
+                            f"Not enough overrides for component {component_id};"
+                            f" Number of agents: {self.number_of_agents}"
+                        ) from e
```
Karrenbelt, line 224: can we not simply check `len(configuration) == self.number_of_agents`? Otherwise might still end up specifying more config, e.g. for a 5th agent, plus it will be cleaner as a bonus.

Karrenbelt:
```
            if override_type == OverrideType.MULTIPLE:
                enforce(len(configuration) == self.number_of_agents)
            else:
                configuration = (configuration,)
            base_validator.validate_component_configuration(
                component_id=component_id,
                configuration=configuration,
                env_vars_friendly=env_vars_friendly,
            )
            processed.append(component_id)
```

autonomy/configurations/base.py
```diff
@@ -207,179 +199,153 @@ def validate_config_data(
...
+            configuration = configuration.get(agent_idx, {})
+            if configuration == {}:
+                raise ValueError(
+                    f"Overrides not provided for agent {agent_idx}; component={component_id}"
                 )
-        return component_id, overrides
+        configuration = apply_env_variables(
+            data=configuration, env_variables=os.environ.copy()
+        )
+        env_var_dict = self.generate_envrionment_variables(
```
Karrenbelt, line 287: typo `envrionment` -> `environment`

autonomy/configurations/base.py
Karrenbelt: - this should be a function not a class method.
- It's too permissive in the else statement since you can cast (almost) anything to a string in python.
``` 
if isinstance(value, (int, float, str, bytes, ...)): 
    ... 
elif isinstance(value, typing.Mapping): 
    ... 
elif isinstance(value, typing.Iterable):
    ... 
else: 
    raise
```

autonomy/configurations/base.py
angrybayblade: If we go with `len(configuration) == self.number_of_agents`, this might lead to an issue where if a user defines an agent index wrong it will raise an error. For example let's say `number_of_agents` is 4 and an user defined indexes as `(0, 1, 2, 3, 5)` the check will still pass and will raise error later

tests/test_autonomy/test_configurations.py
```diff
@@ -161,7 +133,6 @@ def test_load_service(service_file: Path) -> None:
     with open_file(service_file, "r", encoding="utf-8") as fp:
         data = yaml_load_all(fp)
 
-    data = cast(List[Dict], apply_env_variables(data, env_variables=os.environ))
```
DavidMinarsch, line 164: why removed?

tests/data/dummy_service_config_files/service_4.yaml
```diff
@@ -10,10 +10,11 @@ fingerprint_ignore_patterns: []
 agent: valory/hello_world:0.1.0:bafybeiaotnukv7oq2sknot73a4zssrrnjezh6nd2fwptrznxtnovy2rusm
 number_of_agents: 4
 ---
-benchmark_persistence_params:
-  args: &id001
-    log_dir: /benchmarks
-public_id: valory/hello_world_abci:0.1.0
+extra:
+  benchmark_persistence_params:
```
DavidMinarsch, line 14: what's this line 14 for? below we only reference `id001`...

tests/data/dummy_service_config_files/service_2.yaml
```diff
@@ -10,6 +10,7 @@ fingerprint_ignore_patterns: []
 agent: valory/hello_world:0.1.0:bafybeiaotnukv7oq2sknot73a4zssrrnjezh6nd2fwptrznxtnovy2rusm
 number_of_agents: 4
 ---
+override_type: multiple
```
DavidMinarsch, line 13: ? what does this mean

autonomy/configurations/loader.py
```diff
@@ -52,16 +53,24 @@ def load_service_config(
 ) -> Service:
     """Load service config from the path."""
 
+    if substitute_env_vars:
+        warn(
+            "`substitute_env_vars` argument is deprecated and will be removed in v1.0.0, "
+            "usage of envrionment varibales is default now.",
```
DavidMinarsch, line 59:
```suggestion
            "usage of environment variables is default now.",
```

autonomy/configurations/base.py
angrybayblade: The function serves no purpose outside the class and the only data that this method is going to process will be coming from the yaml loader which means the mappings will either be `dict` or `OrderedDict` both of which can be checked with `dict` and the iterables are going to be of type `list` so there's no point in complicating this

tests/test_autonomy/test_configurations.py
angrybayblade: https://github.com/valory-xyz/open-autonomy/pull/1576#discussion_r1030081867

autonomy/configurations/base.py
```diff
@@ -55,25 +57,16 @@
 }
 
 
-def recurse(obj: Dict[str, Any]) -> Dict[str, Any]:
-    """Recursively explore a json object until no dictionaries remain."""
-    if not any([isinstance(i, dict) for i in obj.values()]):
-        return obj
+def export_path_to_string(export_path: List[str]) -> str:
```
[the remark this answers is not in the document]
angrybayblade: This was implemented before the post release. I'll import this.

autonomy/configurations/base.py
```diff
@@ -207,179 +199,146 @@ def validate_config_data(
...
+                cast(
+                    str,
+                    configuration.pop("public_id"),
+                ),
+            ),
+        )
+        _ = configuration.pop("extra", {})
+        overide_type = OverrideType(
+            configuration.pop("override_type", OverrideType.SINGULAR.value)
+        )
```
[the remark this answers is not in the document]
angrybayblade: If we don't do this explicitly we this will make the process of validating the overrides much more complicated. I'll try to make the detection for multiple overrides automatic

autonomy/configurations/base.py
```diff
@@ -207,179 +199,146 @@ def validate_config_data(
...
+        component_id = ComponentId(
+            component_type=cast(str, configuration.pop("type")),
+            public_id=PublicId.from_str(
+                cast(
+                    str,
+                    configuration.pop("public_id"),
+                ),
+            ),
+        )
+        _ = configuration.pop("extra", {})
```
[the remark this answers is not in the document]
angrybayblade: This is a copy of the override object for validation, when dumping we will use the original as it is

autonomy/configurations/base.py
```diff
@@ -248,10 +241,10 @@ def process_metadata(
             ),
         )
         _ = configuration.pop("extra", {})
-        overide_type = OverrideType(
-            configuration.pop("override_type", OverrideType.SINGULAR.value)
+        has_multiple_overrides = all(
+            map(lambda x: isinstance(x, int), configuration.keys())
```
angrybayblade, line 245: Assuming all of the keys will be integers eg (0, 1, ...) we will use this check to decide if the override is singular or multiple type

autonomy/configurations/base.py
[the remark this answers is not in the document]
angrybayblade: [dc23ba2](https://github.com/valory-xyz/open-autonomy/pull/1576/commits/dc23ba2027fea7592ca886de0db748942e07d9f3)

autonomy/configurations/base.py
Karrenbelt: - We can disagree on the use of class methods. This is simply not what they are meant for and classes should not be used as a namespace. 
- Regarding my suggestion, it is generalized, it doesn't need be. What I care about is that there is no explicit check for the `else` where the string casting happens. We expect something specific here, primitives presumably, check for expected instances, else raise.

autonomy/configurations/base.py
```diff
@@ -207,179 +186,154 @@ def validate_config_data(
...
+                        export_path=_export_path,
                     )
-        return overrides
-
-    def try_to_process_nested_fields(  # pylint: disable=too-many-locals
+                )
+            elif isinstance(value, list):
+                env_var_dict[
+                    export_path_to_env_var_string(export_path=export_path)
+                ] = json.dumps(value)
```
Karrenbelt, line 302: we know this is a list of primitives? What if it's a list of list or list of dict? There is no check for this.

angrybayblade: There is no point in checking whatever content is inside of the list because it's going to be exported as a json string because whatever inside the list, it's not going to be an actual parameter so we should not care about it

Karrenbelt (approved):
LGTM - opened [an issue on open-aea](https://github.com/valory-xyz/open-aea/issues/452) related to yaml data types and environmental variable parsing

DavidMinarsch (approved):

PR #1584, opened 2022-11-24
feat/ selector-groups Update templates.py

8ball030:
Please consider adding in this selector. I will then be able to use it to route traffic to agents in a round robin fashion

## Proposed changes

selector labels are used to kube services to determine what pods to route traffic to.

ATM we can only route traffic to specific agents, this allows us to route traffic to a *group* of agents

## Fixes

## Types of changes

## Checklist

## Further comments

### DIFF
--- a/autonomy/deploy/generators/kubernetes/templates.py
+++ b/autonomy/deploy/generators/kubernetes/templates.py
@@ -188,7 +188,6 @@ spec:
     matchLabels:
       app: agent-node-{validator_ix}
       valory: agent-svc
-      
   template:
     metadata:
       labels:
--- a/autonomy/deploy/generators/kubernetes/templates.py
+++ b/autonomy/deploy/generators/kubernetes/templates.py
@@ -187,10 +187,13 @@ spec:
   selector:
     matchLabels:
       app: agent-node-{validator_ix}
+      valory: agent-svc
+      
   template:
     metadata:
       labels:
         app: agent-node-{validator_ix}
+        valory: agent-svc
     spec:
       imagePullSecrets:
       - name: regcred
--- a/autonomy/deploy/generators/kubernetes/templates.py
+++ b/autonomy/deploy/generators/kubernetes/templates.py
@@ -176,7 +176,8 @@ spec:
     protocol: TCP
     name: tcp2
   selector:
-    app: agent-node-{validator_ix}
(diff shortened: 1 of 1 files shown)


autonomy/deploy/generators/kubernetes/templates.py
```diff
@@ -187,10 +187,12 @@
   selector:
     matchLabels:
       app: agent-node-{validator_ix}
+      valory: agent-svc
```
DavidMinarsch, line 190: Open Autonomy is not only used by Valory. This is a non-generic addition to the template. We should think of a cleaner way to do this.

autonomy/deploy/generators/kubernetes/templates.py
```diff
@@ -176,7 +176,8 @@
     protocol: TCP
     name: tcp2
   selector:
-    app: agent-node-{validator_ix}
+    app: agent-svc
```
8ball030, line 179: @Karrenbelt 
How about this one?

autonomy/deploy/generators/kubernetes/templates.py
8ball030: I agree 100%
I had approached it from the other direction with this pr;

https://github.com/valory-xyz/internal-infrastructure/pull/636/files#

Which allows us to deploy prior to the next release of the open-autonomy framework.

autonomy/deploy/generators/kubernetes/templates.py
Karrenbelt: I don't think it can get more generic than that

Karrenbelt (approved):

autonomy/deploy/generators/kubernetes/templates.py
DavidMinarsch: How does this solve the problem where you have two autonomous services in the same cluster? They now are indistinguishable again?

8ball030: They are segmented by name space

8ball030: I.e. i have seperate name spaces for all of my agent services.

0xArdi (approved):

angrybayblade (approved):

PR #1608, opened 2022-12-02
Fix/hard reset recovery

0xArdi:
## Fixes

This PR fixes 2 issues with hard reset when it is used as a recovery mechanism for broken agent <-> tendermint communication. Namely:

1. The initial height and genesis time should not be changed when resetting.
2. We should not wait for 1/2 the observation interval before and after reset when recovering. 

## Types of changes

## Checklist

## Further comments


packages/valory/skills/abstract_round_abci/behaviour_utils.py
```diff
@@ -1722,7 +1742,10 @@ def reset_tendermint_with_wait(
                     for handler_name in self.context.handlers.__dict__.keys():
                         dialogues = getattr(self.context, f"{handler_name}_dialogues")
                         dialogues.cleanup()
-
+                    # in case of successful reset we store the reset params in the shared state,
+                    # so that in the future if the communication with tendermint breaks, and we need to
+                    # perform a hard reset to restore it, we can use these as the right ones
+                    self.context.shared_state["reset_params"] = reset_params
```
DavidMinarsch, line 1748: `shared_state` is not conceptually the right place. There we store information which is relevant cross skills. I'd save it in a model on the skill.

DavidMinarsch: @angrybayblade could rename `shared_state` for v2 and to call it `skills_stared_state` or similar. It gets often misinterpreted.

packages/valory/skills/abstract_round_abci/behaviour_utils.py
```diff
@@ -1607,6 +1607,28 @@ def __parse_rpc_error(error: str) -> RPCResponseStatus:
...
+    @property
+    def sleep_after_hard_reset(self) -> float:
+        """
+        Amount of time to sleep after performing a hard reset.
+
+        We sleep for half the observation interval as there are no immediate transactions on either side of the reset.
+
+        :returns: the amount of time to sleep in seconds
+        """
+        return self.params.observation_interval / 2
```
angrybayblade, line 1630: Why not just make this as a one variable since they both return the same value

packages/valory/skills/abstract_round_abci/behaviour_utils.py
angrybayblade: https://github.com/valory-xyz/open-aea/issues/470

angrybayblade (approved):

DavidMinarsch (approved):

PR #1612, opened 2022-12-05
Fix agent instance validation

angrybayblade:
## Proposed changes

This PR fixes the agent instance validation. 

Before this we used to raise when the key file had missing addresses from the agent instances list, which is not really a good idea because the key file might only contain one key for an operator and the agent instances might contain n amount of addresses.

## Fixes

## Types of changes

## Checklist

## Further comments

### DIFF
--- a/autonomy/deploy/base.py
+++ b/autonomy/deploy/base.py
@@ -186,6 +186,17 @@ class ServiceBuilder:
                 f"Key file contains keys which are not registered as instances; invalid keys={key_not_in_instances}"
             )
 
+        instances_not_in_keys = instances.difference(addresses)
+        if instances_not_in_keys:
+            logging.warning(
+                f"Key file does not contain key pair for following instances {instances_not_in_keys}"
+            )
+
+        keys_found_with_instances = instances.intersection(addresses)
+        logging.info(
+            f"Found following keys with registered instances {keys_found_with_instances}"
+        )
+
     def read_keys(self, keys_file: Path) -> None:
         """Read in keys from a file on disk."""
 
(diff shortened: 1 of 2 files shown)


autonomy/deploy/base.py
```diff
@@ -186,12 +186,6 @@ def verify_agent_instances(
                 f"Key file contains keys which are not registered as instances; invalid keys={key_not_in_instances}"
             )
 
-        instances_not_in_keys = instances.difference(addresses)
-        if instances_not_in_keys:
-            raise NotValidKeysFile(
-                f"Key file does not contain key pair for following instances {instances_not_in_keys}"
-            )
```
angrybayblade, line 193: We should not raise if the keys file does not contain key pair for every single registered agent instance

Karrenbelt: it was here for a reason though. Should we log the addresses for which we do and do not have keys?

Karrenbelt (approved):

autonomy/deploy/base.py
DavidMinarsch: We should indeed log something. Otherwise it's hard to use this magical tool. How about logging the detected addresses and their matching instance identifiers? This would be helpful in the logs

angrybayblade: [d296dc9](https://github.com/valory-xyz/open-autonomy/pull/1612/commits/d296dc9f56d6f83cb8806997ac98ff426b092424)

Karrenbelt (approved):

DavidMinarsch (approved):

PR #1616, opened 2022-12-06
Release v0.5.0

angrybayblade:
## Release summary

Version number: v0.5.0

## Release details

Autonomy:
- Introduces more flexible approach to defining the overrides for multiple agents in a service component
- Deprecates the usage of `autonomy hash all` command, `autonomy packages lock` can be used to perform package dependency hash updates and checks
- Updates the override serialisation mechanism to be consistent with the environment variable parser
- Extends the `from-token` command to provide password for encrypted private keys at the runtime
- Updates the process of overriding the safe contract address at the runtime to be more generalised
- Fixes a bug that allowed the image build to continue even after a command run failed when building the image on `build-image` command
 
Packages:
- Fixes the `AbciApp` initialization to ensure that synchronized data is retrieved as an instance of the `synchronized_data_class` specified on the round.
- Updates the `IPFSInteract` tool to
  - Catch the broken connection exceptions
  - Remove the correct path before downloading
- Fixes the tendermint reset mechanism to 
  - Avoid race conditions when performing a hard reset
  - Not updated the initial height and genesis time when resetting
  - Update the waiting interval

Tests:
- Tests for the service config loader
- Adds test coverage for `autonomy` framework

Docs:
- Adds documentation on the usage of the service level overrides
- Makes sure that naming convention for autonomous services is consistent throughout the documentation

## Checklist

- [x] I have read the [CONTRIBUTING](../main/CONTRIBUTING.md) doc
- [x] I am making a pull request against the `main` branch (left side), from `develop`
- [x] I've updated the dependencies versions to the latest, wherever is possible.
- [x] Lint and unit tests pass locally (please run tests also manually, not only with `tox`)
- [x] I built the documentation and updated it with the latest changes
- [x] I've added an item in `HISTORY.md` for this release
- [x] I bumped the version number in the `__init__.py` file.
- [ ] I published the latest version on TestPyPI and checked that the following command work:
       ```pip install project-name==<version-number> --index-url https://test.pypi.org/simple --force --no-cache-dir --no-deps```
- [ ] After merging the PR, I'll publish the build also on PyPI. Then, I'll make sure the following
      command will work:
      ```pip install project-name-template==<version_number> --force --no-cache-dir --no-deps```  
- [ ] After merging the PR, I'll tag the repo with `v${VERSION_NUMVER}` (e.g. `v0.1.2`)


docs/upgrading.md
```diff
@@ -5,6 +5,14 @@ Below we describe the additional manual steps required to upgrade between differ
 
 # Open Autonomy
 
+## `v0.4.0` to `v0.5.0`
+
+
+One backwards incompatible changes
+## Service component
+
+- This release introduces a new format for defining multiple overrides for an agent on a service configuration. Please follow this [guide](https://docs.autonolas.network/guides/service_configuration_file/) to update your service configurations accordingly.
```
angrybayblade, line 14: This link will work after the new docs release

angrybayblade: Turns out this was not really a good idea because this lead to the doc link check failure, so I updated the reference to the markdown file on the git repo [c533f22](https://github.com/valory-xyz/open-autonomy/pull/1616/commits/c533f22c19b29d5b46a190d1b36f2c232e58e4db)

HISTORY.md
```diff
@@ -1,6 +1,35 @@
...
+- Fixes a bug that allowed the image build to continue even after a command run failed when building the image on `build-image` command
+ 
+Packages:
+- Fixes the `AbciApp` initialization to ensure that synchronized data is retrieved as an instance of the `synchronized_data_class` specified on the round.
+- Updates the `IPFSInteract` tool to
+  - Catch the broken connection exceptions
+  - Remove the correct path before downloading
+- Fixes the tendermint reset mechanism to 
+  - Avoid race conditions when performing a hard reset
+  - Not updated the initial height and genesis time when resetting
```
0xArdi, line 21: Not update the initial height and genesis time when resetting for recovering agent <-> tendermint communication

Karrenbelt: "when it is used as a recovery mechanism for broken agent <-> tendermint communication" - #1608
when there is no failure then these are updated.

Karrenbelt (approved):

0xArdi (approved):

docs/upgrading.md
```diff
@@ -5,6 +5,14 @@ Below we describe the additional manual steps required to upgrade between differ
 
 # Open Autonomy
 
+## `v0.4.0` to `v0.5.0`
+
+
+One backwards incompatible changes
```
DavidMinarsch, line 11: Not proper English, please fix

DavidMinarsch (approved):

PR #1634, opened 2022-12-08
[WIP] experimental: make explicit dependencies between abci components

DavidMinarsch:
## Proposed changes

This PR explicitly introduces the links between the rounds, behaviours and transaction ids as well as keys used in the db.

The purpose of this PR is not to simplify the code base. It is a first step in the refactor of the Payload class. By making links explicit the refactor should be much simpler.

## Fixes

-

## Types of changes

## Checklist

## Further comments

-

### DIFF
--- a/packages/valory/skills/abstract_round_abci/base.py
+++ b/packages/valory/skills/abstract_round_abci/base.py
@@ -714,6 +714,19 @@ class AbciAppDB:
         return {k: [v] for k, v in data.items()}
 
 
+class BaseDBKeys(Enum):
+    """Database keys for the base synchronized data db."""
+
+    PARTICIPANTS = "participants"
+    ALL_PARTICIPANTS = "all_participants"
+    MOST_VOTED_RANDOMNESS = "most_voted_randomness"
+    MOST_VOTED_KEEPER_ADDRESS = "most_voted_keeper_address"
+    BLACKLISTED_KEEPERS = "blacklisted_keepers"
+    PARTICIPANT_TO_SELECTION = "participant_to_selection"
+    PARTICIPANT_TO_RANDOMNESS = "participant_to_randomness"
+    PARTICIPANT_TO_VOTES = "participant_to_votes"
+
+
 class BaseSynchronizedData:
     """
     Class to represent the synchronized data.
@@ -754,7 +767,7 @@ class BaseSynchronizedData:
     @property
     def participants(self) -> FrozenSet[str]:
         """Get the currently active participants."""
-        participants = self.db.get_strict("participants")
+        participants = self.db.get_strict(BaseDBKeys.PARTICIPANTS.value)
         if len(participants) == 0:
             raise ValueError("List participants cannot be empty.")
         return cast(FrozenSet[str], participants)
@@ -762,7 +775,7 @@ class BaseSynchronizedData:
     @property
     def all_participants(self) -> FrozenSet[str]:
         """Get all registered participants."""
-        all_participants = self.db.get_strict("all_participants")
+        all_participants = self.db.get_strict(BaseDBKeys.ALL_PARTICIPANTS.value)
         if len(all_participants) == 0:
             raise ValueError("List participants cannot be empty.")
         return cast(FrozenSet[str], all_participants)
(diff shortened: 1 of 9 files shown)


packages/valory/skills/abstract_round_abci/base.py
```diff
@@ -714,6 +714,19 @@ def data_to_lists(data: Dict[str, Any]) -> Dict[str, List[Any]]:
         return {k: [v] for k, v in data.items()}
 
 
+class BaseDBKeys(Enum):
```
DavidMinarsch, line 717: By introducing an enum we make it more transparent that these are reserved keys in the DB. We can then also apply checks to all the enums across all chained skills to ensure no duplication of keys (which could lead to unintended overwriting).

packages/valory/skills/safe_deployment_abci/__init__.py
```diff
@@ -19,7 +19,22 @@
...
+    """Enumeration of states."""
+
+    DEPLOY_SAFE = "deploy_safe"
+    VALIDATE_SAFE = "validate_safe"
+    RANDOMNESS_SAFE = "randomness_safe"
+    SELECT_KEEPER_SAFE = "select_keeper_safe"
+
+    def __str__(self) -> str:
+        """Get the string value of the state."""
+        return self.value
```
DavidMinarsch, line 40: Could be dropped for consistency

packages/valory/skills/safe_deployment_abci/rounds.py
```diff
@@ -64,58 +72,58 @@ class SynchronizedData(BaseSynchronizedData):
...
 
 
 class RandomnessSafeRound(CollectSameUntilThresholdRound):
     """A round for generating randomness"""
 
-    round_id = "randomness_safe"
+    round_id = States.RANDOMNESS_SAFE.value
     allowed_tx_type = RandomnessPayload.transaction_type
-    payload_attribute = "randomness"
+    payload_attribute = RandomnessPayload.randomness.fget.__name__  # type: ignore
```
DavidMinarsch, line 83: It's annoying that we have to add a type ignore as the whole point was to have the static check. But mypy has a bug: https://github.com/python/mypy/issues/6185

In any case, once we refactor this as part of payload refactor this can be improved upon.

At least now the link is explicit.

packages/valory/skills/abstract_round_abci/base.py
Karrenbelt: So what about introducing. An extra library? :dancing_women: :fireworks: 

[StrEnum](https://docs.python.org/3/library/enum.html#enum.StrEnum) has been introduced in python 3.11. However, it's already available for python 3.7 onward here https://github.com/irgeek/StrEnum. It is a pure python library without any dependencies of its own. We can use it in combination with `auto()` to ensure the key: value pairs are never out of wack.

DavidMinarsch: Hm seems to be offering only marginal utility here. Also keep in mind that this PR is mostly a prep for the actual refactor. It's backward compatible and only intends to make the links more explicit so we have an easier time refactoring.

Karrenbelt: I was hoping it could resolve this issue where mypy always complains that Performative.<SOME_SPECIFIC_PERFORMATIVE> is of type "str" while "Performative" is expected. We have `type: ignore` all throughout our code base for this

Karrenbelt (commented):
generally looks fine, but I think it's good to start with the endgame in mind. To the best of my understanding there is not yet a draft design for what things should look like when we perform the refactor. There are many things to consider here to it's best to ensure we do so. This relates to the metaclass registry of classes, potential DB key conflicts among apps, and more.

packages/valory/skills/safe_deployment_abci/__init__.py
Karrenbelt: this States enumeration looks unnecessary to me. We have rounds, behaviours and payloads, they map one to one. We do not need `TransactionType` or `States` I think, but I would need to see a design on paper, perhaps it does make sense in a scenario I did not envision.

DavidMinarsch: Somewhere one has to define the identifiers. It's better to do this in one place than in two

packages/valory/skills/abstract_round_abci/base.py
DavidMinarsch: Good news on this!

One reason for these
https://github.com/valory-xyz/open-autonomy/blob/main/packages/valory/skills/safe_deployment_abci/behaviours.py#L220-L226
is this issue in mypy https://github.com/python/mypy/issues/4717 which can now be circumvented https://github.com/python/mypy/issues/4717#issuecomment-1264388209 . So when we update tomte next we can finally remove the ignores in this place.

I haven't found the exact issue on mypy that causes the failure you describe. But I am 99% sure it is a mypy issue. The types can be manually checked to be correctly declared.

packages/valory/skills/safe_deployment_abci/behaviours.py
```diff
@@ -59,23 +60,23 @@ def synchronized_data(self) -> SynchronizedData:
 class RandomnessSafeBehaviour(RandomnessBehaviour):
     """Retrieve randomness for oracle deployment."""
 
-    behaviour_id = "randomness_safe"
+    behaviour_id = States.RANDOMNESS_SAFE.value
```
DavidMinarsch, line 63: Alternatively, we can just autogenerate these framework side. Might be simpler still and reduce LOC.

packages/valory/skills/safe_deployment_abci/rounds.py
```diff
@@ -64,58 +72,58 @@ class SynchronizedData(BaseSynchronizedData):
...
     fail_event = Event.FAILED
-    payload_key = "safe_contract_address"
+    payload_key = DBKeys.SAFE_CONTRACT_ADDRESS.value
 
 
 class ValidateSafeRound(VotingRound):
     """A round in a which the safe address is validated"""
 
-    round_id = "validate_safe"
+    round_id = States.VALIDATE_SAFE.value
```
DavidMinarsch, line 119: Same as with behaviours, we can autogenerate this instead.

DavidMinarsch:
Closed in favour of `ref/exp2`

PR #1640, opened 2022-12-09
refactor: adds support for auto-generated behaviour ids

DavidMinarsch:
## Proposed changes

This PR refactors `BaseBehaviour` such that `behaviour_id` is no longer required to be set by the developer on the child class. The developer can still do so, as backwards compatibility is maintained in that regard.

## Fixes

-

## Types of changes

## Checklist

## Further comments

-


packages/valory/skills/abstract_round_abci/behaviour_utils.py
```diff
@@ -488,6 +489,26 @@ def __init__(self, **kwargs: Any):  # pylint: disable=super-init-not-called
...
+        """
+        Get behaviour id automatically.
+
+        This method returns the auto generated id from the class name if the
+        class variable behaviour_id is not set on the child class.
+        Otherwise, it returns the class variable behaviour_id.
+        """
+        return (
+            cls.behaviour_id
+            if type(cls.behaviour_id) == str and cls.behaviour_id != ""
```
DavidMinarsch, line 503: On this class, the base class, the type of `cls.behaviour_id` is property. On a child class, if the class var is set a string, hence the check.

packages/valory/skills/abstract_round_abci/test_tools/base.py
```diff
@@ -130,9 +130,10 @@ def setup(self, **kwargs: Any) -> None:
 
         self.benchmark_dir = TemporaryDirectory()
         self._skill.skill_context.benchmark_tool.log_dir = Path(self.benchmark_dir.name)
+        # import pdb; pdb.set_trace()  # noqa: E402
```
DavidMinarsch, line 133: clean up

packages/valory/skills/safe_deployment_abci/behaviours.py
```diff
@@ -59,23 +59,20 @@ def synchronized_data(self) -> SynchronizedData:
 class RandomnessSafeBehaviour(RandomnessBehaviour):
     """Retrieve randomness for oracle deployment."""
 
-    behaviour_id = "randomness_safe"
```
DavidMinarsch, line 62: This shows the benefit. We can drop these - optionally.

packages/valory/skills/safe_deployment_abci/tests/test_behaviours.py
```diff
@@ -89,7 +89,7 @@ def test_validate_behaviour(self) -> None:
         """Run test."""
         self.fast_forward_to_behaviour(
             self.behaviour,
-            self.behaviour_class.behaviour_id,
+            self.behaviour_class.auto_behaviour_id(),
```
DavidMinarsch, line 92: When dropping, this is a required change. As now the `behaviour_id` at the class level is a property object, not a string.

packages/valory/skills/abstract_round_abci/behaviour_utils.py
```diff
@@ -488,6 +489,26 @@ def __init__(self, **kwargs: Any):  # pylint: disable=super-init-not-called
...
+        class variable behaviour_id is not set on the child class.
+        Otherwise, it returns the class variable behaviour_id.
+        """
+        return (
+            cls.behaviour_id
+            if isinstance(cls.behaviour_id, str)
+            else cls.__pattern.sub("_", cls.__name__).lower()
+        )
+
+    @property  # type: ignore
```
DavidMinarsch, line 507: Type ignore because we keep the above class variable for backwards compatibility.

packages/valory/skills/abstract_round_abci/behaviour_utils.py
```diff
@@ -488,6 +489,26 @@ def __init__(self, **kwargs: Any):  # pylint: disable=super-init-not-called
...
+
+        This method returns the auto generated id from the class name if the
+        class variable behaviour_id is not set on the child class.
+        Otherwise, it returns the class variable behaviour_id.
+        """
+        return (
+            cls.behaviour_id
+            if isinstance(cls.behaviour_id, str)
+            else cls.__pattern.sub("_", cls.__name__).lower()
+        )
```
0xArdi, line 505: Would it make sense to include the module name as well? Otherwise we can have conflicts that are not obvious.

packages/valory/skills/abstract_round_abci/test_tools/base.py
```diff
@@ -132,7 +132,7 @@ def setup(self, **kwargs: Any) -> None:
         self._skill.skill_context.benchmark_tool.log_dir = Path(self.benchmark_dir.name)
         assert (  # nosec
             cast(BaseBehaviour, self.behaviour.current_behaviour).behaviour_id
-            == self.behaviour.initial_behaviour_cls.behaviour_id
+            == self.behaviour.initial_behaviour_cls.auto_behaviour_id()
```
0xArdi, line 135: We will need to update all our tests, if we decide to remove the `behaviour_id` from them, ie see the benefit. We can obviously avoid that by leaving the `behaviour_id` where it already is.

packages/valory/skills/safe_deployment_abci/behaviours.py
0xArdi: That would mean that we will need to update the tests accordingly.

packages/valory/skills/abstract_round_abci/test_tools/base.py
DavidMinarsch: Yes if you want to upgrade you need to fix tests, o/w no change needed.

Adamantios (approved):
LGTM. Maybe a tiny test for `auto_behaviour_id` would be helpful since it uses the `__pattern`.

Karrenbelt (commented):
I am not a fan of this approach for the following reasons: it introduced an additional class attribute, a class method and a property that is also used as a method. None of these are needed.

1. for internal usage, e.g. metaclass registry, we can use the cls.__name__ directly; there is no need to mutate it in any way.
2. I would prefer keeping the class name as the `id` as well for the instance, alternatively however, one can use a private module, akin to `pytest` and `_pytest`. This can contain a simple helper function that transforms a class name into an instance attribute.

```
from _deprecation_helpers import _auto_identifier

class MySpecialBehaviour:

    def __init__(self):
        if not hasattr(self, "behaviour_id"):
            self.behaviour_id = _auto_identifier(self)
```

this way we do not expose any additional interface to developers either (e.g. `auto_behaviour_id`). Plus it can be reused by the round classes as well.

Adamantios:
> I am not a fan of this approach for the following reasons: it introduced an additional class attribute, a class method and a property that is also used as a method. None of these are needed.
> 
> 1. for internal usage, e.g. metaclass registry, we can use the cls.**name** directly; there is no need to mutate it in any way.
> 2. I would prefer keeping the class name as the `id` as well for the instance, alternatively however, one can use a private module, akin to `pytest` and `_pytest`. This can contain a simple helper function that transforms a class name into an instance attribute.
> 
> ```
> from _deprecation_helpers import _auto_identifier
> 
> class MySpecialBehaviour:
> 
>     def __init__(self):
>         if not hasattr(self, "behaviour_id"):
>             self.behaviour_id = _auto_identifier(self)
> ```
> 
> this way we do not expose any additional interface to developers either (e.g. `auto_behaviour_id`). Plus it can be reused by the round classes as well.

This approach would require an instance and would not work for example in cases like:

https://github.com/valory-xyz/open-autonomy/blob/95b71d5dcd59137a87aa600af31ad64fe4fe3250/packages/valory/agents/hello_world/tests/test_hello_world.py#L54-L58

Karrenbelt:
> > I am not a fan of this approach for the following reasons: it introduced an additional class attribute, a class method and a property that is also used as a method. None of these are needed.
> > 
> > 1. for internal usage, e.g. metaclass registry, we can use the cls.**name** directly; there is no need to mutate it in any way.
> > 2. I would prefer keeping the class name as the `id` as well for the instance, alternatively however, one can use a private module, akin to `pytest` and `_pytest`. This can contain a simple helper function that transforms a class name into an instance attribute.
> > 
> > ```
> > from _deprecation_helpers import _auto_identifier
> > 
> > class MySpecialBehaviour:
> > 
> >     def __init__(self):
> >         if not hasattr(self, "behaviour_id"):
> >             self.behaviour_id = _auto_identifier(self)
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > this way we do not expose any additional interface to developers either (e.g. `auto_behaviour_id`). Plus it can be reused by the round classes as well.
> 
> This approach would require an instance and would not work for example in cases like:
> 
> https://github.com/valory-xyz/open-autonomy/blob/95b71d5dcd59137a87aa600af31ad64fe4fe3250/packages/valory/agents/hello_world/tests/test_hello_world.py#L54-L58

yes you would need to use:  `RoundChecks(RegistrationRound.auto_round_id())`  -> `RoundChecks(_auto_identified(RegistrationRound))`. Or you don't mutate the class name and just use it directly CamelCase style :shrug:

DavidMinarsch:
> > > I am not a fan of this approach for the following reasons: it introduced an additional class attribute, a class method and a property that is also used as a method. None of these are needed.
> > > 
> > > 1. for internal usage, e.g. metaclass registry, we can use the cls.**name** directly; there is no need to mutate it in any way.
> > > 2. I would prefer keeping the class name as the `id` as well for the instance, alternatively however, one can use a private module, akin to `pytest` and `_pytest`. This can contain a simple helper function that transforms a class name into an instance attribute.
> > > 
> > > ```
> > > from _deprecation_helpers import _auto_identifier
> > > 
> > > class MySpecialBehaviour:
> > > 
> > >     def __init__(self):
> > >         if not hasattr(self, "behaviour_id"):
> > >             self.behaviour_id = _auto_identifier(self)
> > > ```
> > > 
> > > 
> > >     
> > >       
> > >     
> > > 
> > >       
> > >     
> > > 
> > >     
> > >   
> > > this way we do not expose any additional interface to developers either (e.g. `auto_behaviour_id`). Plus it can be reused by the round classes as well.
> > 
> > 
> > This approach would require an instance and would not work for example in cases like:
> > https://github.com/valory-xyz/open-autonomy/blob/95b71d5dcd59137a87aa600af31ad64fe4fe3250/packages/valory/agents/hello_world/tests/test_hello_world.py#L54-L58
> 
> yes you would need to use: `RoundChecks(RegistrationRound.auto_round_id())` -> `RoundChecks(_auto_identified(RegistrationRound))`. Or you don't mutate the class name and just use it directly CamelCase style 🤷

We never care about instances of an `AbstractRound`. They are ephemeral and all the relevant data is either present in code (on the class) or in the models. All consistency checks happen at the class level and the chosen approach is minimally invasive and fully backward compatible.

packages/valory/skills/abstract_round_abci/abci_app_chain.py
```diff
@@ -172,6 +172,7 @@ def chain(  # pylint: disable=too-many-locals
     new_cross_period_persisted_keys = []
     for app in abci_apps:
         new_cross_period_persisted_keys.extend(app.cross_period_persisted_keys)
+    new_cross_period_persisted_keys = list(set(new_cross_period_persisted_keys))
```
DavidMinarsch, line 175: There was an issue here with potential duplicate keys.

DavidMinarsch: (One day this attribute should be changed to set type)

packages/valory/skills/abstract_round_abci/base.py
```diff
@@ -1959,6 +1959,12 @@ def add_termination(
         cls.background_round_cls = background_round_cls
         cls.termination_transition_function = termination_abci_app.transition_function
         cls.termination_event = termination_event
+        new_cross_period_persisted_keys = copy(cls.cross_period_persisted_keys)
+        new_cross_period_persisted_keys.extend(
+            termination_abci_app.cross_period_persisted_keys
+        )
+        new_cross_period_persisted_keys = list(set(new_cross_period_persisted_keys))
+        cls.cross_period_persisted_keys = new_cross_period_persisted_keys
```
DavidMinarsch, line 1967: This was missing leading to flakiness in the e2e tests of register_termination_abci

packages/valory/skills/abstract_round_abci/base.py
```diff
@@ -78,6 +79,15 @@
 TransactionType = TypeVar("TransactionType")
 
 
+def get_name(prop: Any) -> str:
```
DavidMinarsch, line 82: https://github.com/valory-xyz/open-autonomy/pull/1642#discussion_r1045711425
-> property typing doesn't work

packages/valory/skills/abstract_round_abci/tests/test_base.py
```diff
@@ -1031,7 +1033,25 @@ def setup(self) -> None:
         )
         self.round = ConcreteRoundA(self.base_synchronized_data, self.params)
 
-    def test_must_set_round_id(self) -> None:
+    def test_auto_round_id(self) -> None:
```
DavidMinarsch, line 1036: https://github.com/valory-xyz/open-autonomy/pull/1640#pullrequestreview-1213320197

scripts/check_doc_links.py
```diff
@@ -56,6 +56,7 @@
     "https://gateway.autonolas.tech/ipfs/<hash>,",  # non link (400)
     "https://github.com/valory-xyz/open-autonomy/trunk/infrastructure",  # svn link (404)
     "http://host.docker.internal:8545",  # internal (ERR_NAME_NOT_RESOLVED)
+    "https://notes.ethereum.org/@vbuterin/eip-1559-faq",  # temporarily down
```
Adamantios, line 59: cab3399662a2b2210c66c007c04499757ff632c3 can be reverted before merging on `main`.

Adamantios (approved):

packages/valory/skills/abstract_round_abci/base.py
DavidMinarsch: (As in it does work but it doesn't yield the desired effect. Mypy detects a different type than `property` and so it all becomes very cumbersome)

PR #1641, opened 2022-12-09
refactor: adds support for auto-generated round ids

DavidMinarsch:
## Proposed changes

This PR refactors AbstractRound such that round_id is no longer required to be set by the developer on the child class. The developer can still do so, as backwards compatibility is maintained in that regard.

## Fixes

-

## Types of changes

## Checklist

## Further comments

-


packages/valory/skills/abstract_round_abci/base.py
```diff
@@ -906,6 +908,26 @@ def _check_class_attributes(self) -> None:
         if not hasattr(self, "synchronized_data_class"):
             logging.warning(f"No `synchronized_data_class` set on {self}")
 
+    @classmethod
```
DavidMinarsch, line 911: symmetric to behaviour side

packages/valory/skills/safe_deployment_abci/rounds.py
```diff
@@ -70,7 +70,6 @@ def safe_contract_address(self) -> str:
 class RandomnessSafeRound(CollectSameUntilThresholdRound):
     """A round for generating randomness"""
 
-    round_id = "randomness_safe"
```
DavidMinarsch, line 73: the benefit

packages/valory/skills/safe_deployment_abci/tests/test_behaviours.py
```diff
@@ -303,7 +303,7 @@ class TestValidateSafeBehaviour(BaseValidateBehaviourTest, SafeDeploymentAbciBas
     """Test ValidateSafeBehaviour."""
 
     behaviour_class = ValidateSafeBehaviour
-    next_behaviour_class = make_degenerate_behaviour(FinishedSafeRound.round_id)
+    next_behaviour_class = make_degenerate_behaviour(FinishedSafeRound.auto_round_id())
```
DavidMinarsch, line 306: When no longer specifying behaviour ids we need to use the class method to retrieve the auto generated id

packages/valory/skills/abstract_round_abci/base.py
```diff
@@ -906,6 +908,26 @@ def _check_class_attributes(self) -> None:
...
+
+        This method returns the auto generated id from the class name if the
+        class variable behaviour_id is not set on the child class.
+        Otherwise, it returns the class variable behaviour_id.
+        """
+        return (
+            cls.round_id
+            if isinstance(cls.round_id, str)
+            else cls.__pattern.sub("_", cls.__name__).lower()
+        )
```
0xArdi, line 924: Same concern about conflicts as on the behaviour side. Does it make sense to include the module name?

DavidMinarsch: Same response

Adamantios (approved):
Similar to the behaviour refactoring, therefore I would like to comment on [the same](https://github.com/valory-xyz/open-autonomy/pull/1640#pullrequestreview-1213320197), but for `auto_round_id`.

~~`test_must_set_round_id` needs fixing.~~ I see it's been fixed on #1643.

PR #1642, opened 2022-12-10
refactor: removes duplication and improves type checks

DavidMinarsch:
## Proposed changes

This PR adds the `get_name` method to retrieve the name of a property dynamically. This allows us to benefit from the type checks on the property.

## Fixes

-

## Types of changes

## Checklist

## Further comments

-


packages/valory/skills/abstract_round_abci/behaviour_utils.py
```diff
@@ -1803,7 +1803,6 @@ def reset_tendermint_with_wait(  # pylint: disable= too-many-locals
 class TmManager(BaseBehaviour, ABC):
     """Util class to be used for managing the tendermint node."""
 
-    behaviour_id = "tm_manager"
     _active_generator: Optional[Generator] = None
```
DavidMinarsch, line 1806: Now autogenerated

packages/valory/skills/safe_deployment_abci/payloads.py
```diff
@@ -46,7 +46,10 @@ class RandomnessPayload(BaseTxPayload):
     def __init__(
         self, sender: str, round_id: int, randomness: str, **kwargs: Any
     ) -> None:
-        """Initialize an 'select_keeper' transaction payload.
+        """Initialize an 'randomness' transaction payload.
```
DavidMinarsch, line 49: Added for clarification as it's otherwise hard to understand why we have this second property

packages/valory/skills/safe_deployment_abci/rounds.py
```diff
@@ -71,47 +72,47 @@ class RandomnessSafeRound(CollectSameUntilThresholdRound):
...
 
     allowed_tx_type = RandomnessPayload.transaction_type
-    payload_attribute = "randomness"
+    payload_attribute = get_name(RandomnessPayload.randomness)
     synchronized_data_class = SynchronizedData
     done_event = Event.DONE
     no_majority_event = Event.NO_MAJORITY
-    collection_key = "participant_to_randomness"
-    selection_key = "most_voted_randomness"
+    collection_key = get_name(SynchronizedData.participant_to_randomness)
```
DavidMinarsch, line 79: Benefit of RHS over LHS. It's harder to make an error as `mypy` will complain during static checks if the property does not exist. Since we have a 1:1 mapping between properties on `SynchronizedData` and the db keys this then gives us more static checks and it's more easy to follow how things connect.

packages/valory/skills/safe_deployment_abci/rounds.py
```diff
@@ -71,47 +72,47 @@ class RandomnessSafeRound(CollectSameUntilThresholdRound):
     """A round for generating randomness"""
 
     allowed_tx_type = RandomnessPayload.transaction_type
-    payload_attribute = "randomness"
+    payload_attribute = get_name(RandomnessPayload.randomness)
```
DavidMinarsch, line 75: Here benefit should be obvious. Impossible to pass static checks with misdefined properties.

packages/valory/skills/safe_deployment_abci/rounds.py
```diff
@@ -71,47 +72,47 @@ class RandomnessSafeRound(CollectSameUntilThresholdRound):
...
     synchronized_data_class = SynchronizedData
     done_event = Event.DONE
     no_majority_event = Event.NO_MAJORITY
-    collection_key = "participant_to_randomness"
-    selection_key = "most_voted_randomness"
+    collection_key = get_name(SynchronizedData.participant_to_randomness)
+    selection_key = get_name(SynchronizedData.most_voted_randomness)
```
DavidMinarsch, line 81: A further benefit is that we can now more easily refactor the rounds and remove some of these class variables as they can be inferred automatically (e.g. `payload_attribute` could be inferred if we made the assumption that every transaction has exactly one payload attribute - which currently holds apart from one edge case that can be altered).

packages/valory/skills/abstract_round_abci/base.py
```diff
@@ -79,6 +79,15 @@
 TransactionType = TypeVar("TransactionType")
 
 
+def get_name(prop: Any) -> str:
```
Adamantios, line 82: Would `mypy` warn about this? 

```suggestion
def get_name(prop: property) -> str:
```

packages/valory/skills/abstract_round_abci/base.py
```diff
@@ -79,6 +79,15 @@
...
+    """Get the name of a property."""
+    if not (isinstance(prop, property) and hasattr(prop, "fget")):
+        raise ValueError(f"{prop} is not a property")
+    if prop.fget is None:
+        raise ValueError(f"{prop.fget} is None")
+    return prop.fget.__name__
```
Adamantios, line 88: We can also set the `payload_attribute` automatically in the `AbstractRound` - only if not set so we keep the backward compatibility.

DavidMinarsch: Correct. But only if we assume that the payload only has a single property. This assumption is not uniformly true yet (see eg RandomnessRound), but I think we should force that as part of `0.6.0`.

PR #1643, opened 2022-12-10
refactor: applies new features changes consistently throughout

DavidMinarsch:
## Proposed changes

This PR brings the auto id feature to all skills and applies the other refactoring from the base branches.

## Fixes

-

## Types of changes

## Checklist

## Further comments

-


packages/valory/skills/abstract_round_abci/base.py
```diff
@@ -208,7 +208,12 @@ def __init__(
         """
         self.id_ = uuid.uuid4().hex if id_ is None else id_
         self._round_count = round_count
-        self.sender = sender
+        self._sender = sender
+
+    @property
```
DavidMinarsch, line 213: for consistency with the other properties

packages/valory/skills/registration_abci/behaviours.py
```diff
@@ -83,8 +83,6 @@ class RegistrationStartupBehaviour(RegistrationBaseBehaviour):
     """Agent registration to the FSM App."""
 
     ENCODING: str = "utf-8"
-    state_id = "registration_startup"
```
DavidMinarsch, line 86: No idea what `state_id` was meant for...

packages/valory/skills/termination_abci/behaviours.py
```diff
@@ -530,8 +529,6 @@ def is_round_ended(self, round_id: str) -> Callable[[], bool]:
 class TerminationBehaviour(BaseBehaviour):
     """Behaviour responsible for terminating the agent."""
 
-    state_id: str = "termination"
```
DavidMinarsch, line 533: Another mysterious `state_id`

packages/valory/skills/termination_abci/rounds.py
```diff
@@ -65,19 +66,12 @@ def termination_majority_reached(self) -> bool:
         """Get termination_majority_reached."""
         return cast(bool, self.db.get("termination_majority_reached", False))
 
-    @property
```
DavidMinarsch, line 68: Already exists at the base

packages/valory/skills/termination_abci/tests/test_behaviours.py
```diff
@@ -514,7 +514,9 @@ def test_run(self, test_case: BehaviourTestCase, caplog: LogCaptureFixture) -> N
 
     def test_termination_majority_already_reached(self) -> None:
         """Tests the background behaviour when the termination is already reached."""
-        self.fast_forward(data=dict(termination_majority_reached=True))
+        self.fast_forward(
+            data=dict(termination_majority_reached=True, participants=["a"])
```
DavidMinarsch, line 518: Added one more pair to account for using base version of `nb_participants`

packages/valory/skills/transaction_settlement_abci/behaviours.py
```diff
@@ -533,23 +530,17 @@ def _check_tx_history(
...
     ) -> Generator[None, None, Tuple[VerificationStatus, Optional[str]]]:
         """Check the transaction history."""
-        history = (
-            self.synchronized_data.tx_hashes_history
-            if self.behaviour_id == CHECK_TX_HISTORY
-            else self.synchronized_data.late_arriving_tx_hashes
-        )
-
-        if not history:
+        if not self.history:
```
DavidMinarsch, line 533: Refactor to avoid conditioning on id (which would require unnecessary hardcoding)

.github/workflows/main_workflow.yml
```diff
@@ -406,6 +406,7 @@ jobs:
           docker pull valory/autonolas-registries:latest
           docker pull valory/contracts-amm:latest
           docker pull valory/safe-contract-net:latest
+          docker pull tendermint/tendermint:v0.34.19
```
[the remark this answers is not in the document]
0xArdi: It should speed up at least 1 test I suppose. Its a small image compared to the others ~110MB compared to >1GB for those above it.

packages/valory/skills/termination_abci/behaviours.py
0xArdi: The scaffold has it by default I believe, here we use it all over the place https://github.com/valory-xyz/autonomous-fund/blob/main/packages/balancer/skills/fear_and_greed_oracle_abci/behaviours.py

packages/valory/skills/termination_abci/rounds.py
0xArdi: I don't think this removal is right. The base one uses `self.participants` which uses `get_strict` instead of `get`. So this method would raise a `ValueError` when the list of participants is not set. Which can happen since the BackgroundRound doesn't depend on any of the other rounds finishing.

DavidMinarsch: This has been changed on the base now. You are right but the base implementation was not ideal. Your override is the more appropriate solution as we should also be able to use the property with 0 participants.

Adamantios (approved):

packages/valory/skills/registration_abci/rounds.py
```diff
@@ -44,6 +44,19 @@ class Event(Enum):
     FAST_FORWARD = "fast_forward"
 
 
+class SynchronizedData(BaseSynchronizedData):
```
DavidMinarsch, line 47: I added this here as we often use `safe_contract_address` in setup and hence need to insist it being propagated through periods

packages/valory/skills/termination_abci/rounds.py
```diff
@@ -173,6 +173,7 @@ class PostTerminationTxAbciApp(AbciApp[Event]):
     # the following is not needed, it is added to satisfy the round check
     # the TerminationRound when run it terminates the agent, so nothing can come after it
     transition_function = {TerminationRound: {Event.TERMINATE: TerminationRound}}
+    # TODO: ensure this gets considered in chaining too
```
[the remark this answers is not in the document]
0xArdi: Yes we have this in the backlog. What we could do is ensure that the `cross_period_persisted_keys` defined in the `TerminationAbciApp` get also merged into the `cross_period_persisted_keys` of the "normal" AbciApp being executed. This would require small modification to the `AbciApp.add_termination()`. 
This would help us make sure that the `cross_period_persisted_keys` of the TerminationAbciApp get persisted properly, as there is no need to make the assumption that the `cross_period_persisted_keys` of the TerminationAbciApp are necessary anywhere else.

Issue #1430, opened 2022-10-13
Registration start up - `ABCIAppInternalError("last timestamp is None")`


Karrenbelt:
In failing registration e2e test ([CI run here](https://github.com/valory-xyz/open-autonomy/actions/runs/3219922582/jobs/5266113780)) we observe the following:
```
2022-10-10T15:00:48.6841187Z packages.valory.skills.abstract_round_abci.base.ABCIAppInternalError: internal error: last timestamp is None
2022-10-10T15:00:48.6843648Z     raise ABCIAppInternalError("last timestamp is None")
```
which is caused by the first Tendermint reset being performed before after Tendermint config sharing and before any transactions have been recorded. 

Currently, we start Tendermint before sharing configurations. If Tendermint config sharing is enabled, Tendermint should not start automatically and instead only after configurations have been exchanged. Then we won't need to restart Tendermint, but start Tendermint for the first time, where no last timestamp should be checked and the issue hence not exist.

Karrenbelt:
Logs from locally production pertaining to the issue
```
2022-10-13 16:37:50 [    INFO] Waiting for string My address:  in last agent output (agents.py:444)
2022-10-13 16:37:51 [    INFO] Terminated agent_00000 (agents.py:464)
2022-10-13 16:37:51 [    INFO] Waiting 10 seconds before restarting the agent (agents.py:467)
2022-10-13 16:38:01 [    INFO] Launching agent agent_00000... (agents.py:250)
2022-10-13 16:38:01 [    INFO] Restarted agent_00000 (agents.py:476)
2022-10-13 16:38:15 [    INFO] Performing checks for the round strings. (agents.py:312)
[180030]stdout>    raise ABCIAppInternalError("last timestamp is None")
[180030]stdout>packages.valory.skills.abstract_round_abci.base.ABCIAppInternalError: internal error: last timestamp is None
…
[180030]stdout>[2022-10-13 16:39:15,473] [INFO] [agent_00000] Sharing Tendermint config on start-up?: True
[180030]stdout>[2022-10-13 16:39:15,631] [INFO] [agent_00000] My address: 0xf39Fd6e51aad88F6F4ce6aB8827279cffFb92266
[180030]stdout>[2022-10-13 16:39:15,632] [ERROR] [agent_00000] <An error occurred during act of behaviour valory/registration_abci:0.1.0/AgentRegistrationRoundBehaviour:
```

Karrenbelt:
[...]

PR #1651, opened 2022-12-12
Fix/last timestamp error

0xArdi:
## Fixes

Fixes https://github.com/valory-xyz/open-autonomy/issues/1430, by:
- Removing sleep on `reset_tendermint_with_wait` on startup.
- Waiting for block production to begin before sending an a2a transaction via tendermint.  

## Types of changes

## Checklist

## Further comments


packages/valory/skills/registration_abci/behaviours.py
```diff
@@ -333,7 +337,36 @@ def request_update(self) -> Generator[None, None, bool]:
...
+        # every agent will finish with the reset at a different time
+        # hence the following will be different for all agents
+        start_time = datetime.datetime.now()
+
+        def received_block() -> bool:
+            """Check whether we have received a block after "start_time"."""
+            try:
+                shared_state = cast(SharedState, self.context.state)
+                last_timestamp = shared_state.round_sequence.last_timestamp
+                if last_timestamp > start_time:
```
Adamantios, line 351: Would that make sense? We might have already received a block.

```suggestion
        start_time =  shared_state.round_sequence.last_timestamp or datetime.datetime.now()

        def received_block() -> bool:
            """Check whether we have received a block after "start_time"."""
            try:
                shared_state = cast(SharedState, self.context.state)
                last_timestamp = shared_state.round_sequence.last_timestamp
                if last_timestamp:
```

0xArdi: We need to know that its a block received after the reset, the block *CAN* be before the reset as well.

plugins/aea-test-autonomy/aea_test_autonomy/docker/tendermint.py
```diff
@@ -269,33 +268,25 @@ def _fix_persistent_peers(self) -> None:  # pylint: disable=too-many-locals
...
-                'persistent_peers = "' + ",".join(updated_peers) + '"\n'
-            )
-            updated_config = re.sub(
-                'persistent_peers = ".*\n', persistent_peers_string, config_text
-            )
-
-            config_file.write_text(updated_config, encoding="utf-8")
+        replace_cmd = ""
+        for node_num in range(self.nb_nodes):
+            preconfigured_endpoint = f"node{node_num}:26656"
```
Adamantios, line 273:
```suggestion
            preconfigured_endpoint = f"node{node_num}:{self.p2p_port}"
```

packages/valory/skills/registration_abci/behaviours.py
Adamantios: Ok so we need the comparison! How about the initialization? To avoid the `datetime.datetime.now()` if possible. Even better if we could wait for the first block before starting so that we do not use local time at all.

plugins/aea-test-autonomy/aea_test_autonomy/docker/tendermint.py
0xArdi: Why do we take it as a param at all?

Adamantios: Yes, that works too. Also, no need to have this as a param afaict.

packages/valory/skills/registration_abci/behaviours.py
0xArdi: So this behaviour is for agents to share configs with each other and learn about their tm configs. Before this behaviour is finished the tm nodes are not connected to each other, in the e2e tests they are connected, but this will not be the case when this feature is used in reality. So what CAN happen is we receive a block before the reset, but that can be from the badly configured (unconnected) tendermint node. We want blocks from the properly connected tm node.

Adamantios (approved):

Adamantios (approved):

DavidMinarsch (approved):

PR #1659, opened 2022-12-13
[0.6.0] feat: add pre- and post- conditions on db at fsm level

DavidMinarsch:
## Proposed changes

We can check for presence of data at beginning and end of an ABCI app. We do so by introducing pre- and post-conditions on the db for each initial and final state. These can then be verified during chaining of the apps.

The FSM needs to list all expectations it has on data, including those available on the base interface.

TODO:
-> unify the various randomness & select keeper behaviours/rounds. move those properties into the base synchronized data interface.

-> registered_addresses -> need to match with all_participants

-> unify code: https://github.com/valory-xyz/open-autonomy/pull/1645#discussion_r1045265699

## Fixes

## Types of changes

## Checklist

## Further comments

-


autonomy/fsm/scaffold/templates/components.py
```diff
@@ -70,9 +70,9 @@ class {RoundCls}({ABCRoundCls}):
         \"\"\"{RoundCls}\"\"\"
 
         {todo_abstract_round_cls}
-        round_id: str = "{round_id}"
         allowed_tx_type: Optional[TransactionType] = {PayloadCls}.transaction_type
-        payload_attribute: str = "{round_id}"
```
DavidMinarsch, line 75: This is just wrong

autonomy/fsm/scaffold/templates/components.py
```diff
@@ -92,8 +92,6 @@ def process_payload(self, payload: {PayloadCls}) -> None:
     class {RoundCls}({ABCRoundCls}):
         \"\"\"{RoundCls}\"\"\"
 
-        round_id: str = "{round_id}"
```
DavidMinarsch, line 95: removed as computed automatically now

autonomy/fsm/scaffold/templates/components.py
```diff
@@ -158,9 +156,6 @@ def params(self) -> Params:
     class {BehaviourCls}({BaseBehaviourCls}):
         \"\"\"{BehaviourCls}\"\"\"
 
-        # TODO: set the following class attributes
-        state_id: str
-        behaviour_id: str = "{behaviour_id}"
```
DavidMinarsch, line 163: same

autonomy/fsm/scaffold/templates/components.py
```diff
@@ -158,9 +156,6 @@ def params(self) -> Params:
     class {BehaviourCls}({BaseBehaviourCls}):
         \"\"\"{BehaviourCls}\"\"\"
 
-        # TODO: set the following class attributes
-        state_id: str
```
DavidMinarsch, line 162: pointless - no idea why it was added

packages/valory/skills/abstract_round_abci/abci_app_chain.py
```diff
@@ -108,6 +116,88 @@ def chain(  # pylint: disable=too-many-locals,too-many-statements
                 f"Found non-initial state {value} specified in abci_app_transition_mapping."
             )
```
DavidMinarsch, line 118: This whole method is now massive. Can be eventually refactored.

packages/valory/skills/abstract_round_abci/base.py
```diff
@@ -1848,6 +1849,28 @@ def _check_initial_states_and_final_states(
             "final round classes must be subclasses of the DegenerateRound class",
         )
 
+    @classmethod
+    def _check_db_constraints_consistency(mcs, abci_app_cls: Type["AbciApp"]) -> None:
+        """Check that the pre and post constraints on the db are consistent with the initial and final states."""
+        expected = abci_app_cls.initial_states
+        actual = abci_app_cls.db_pre_conditions.keys()
+        invalid_initial_states = (
```
DavidMinarsch, line 1857: Note, if the pre/post conditions are not set we are not checking anything

packages/valory/skills/abstract_round_abci/tests/data/dummy_abci/rounds.py
```diff
@@ -129,7 +129,6 @@ class DummyAbciApp(AbciApp[Event]):
     """DummyAbciApp"""
 
     initial_round_cls: AppState = DummyStartingRound
-    initial_states: Set[AppState] = {DummyStartingRound}
```
DavidMinarsch, line 132: need to check why I removed it

packages/valory/skills/abstract_round_abci/tests/test_abci_app_chain.py
```diff
@@ -126,6 +132,8 @@ class AbciApp2(AbciApp):
             }
             final_states = {self.round_2c}
             event_to_timeout = {self.event_timeout2: self.timeout2}
+            db_pre_conditions = {self.round_2a: [self.key_1]}  # type: ignore
+            db_post_conditions = {self.round_2c: [self.key_2]}  # type: ignore
```
DavidMinarsch, line 136: Question is if we should force the dev to make the post conditions strictly include the pre conditions or not. I say we could force that post conditions do **not** include pre conditions as there is no mechanism to delete data from db atm and it is just unnecessary duplication. Also the app does not "generate" this data (it explicitly has it as a pre condition).

packages/valory/skills/registration_abci/rounds.py
```diff
@@ -166,3 +165,19 @@ class AgentRegistrationAbciApp(AbciApp[Event]):
...
+    db_post_conditions: Dict[AppState, List[str]] = {
+        FinishedRegistrationRound: [
+            get_name(BaseSynchronizedData.participants),
+            get_name(BaseSynchronizedData.all_participants),
+            get_name(BaseSynchronizedData.period_count),
+        ],
+        FinishedRegistrationFFWRound: [
+            get_name(BaseSynchronizedData.participants),
+            get_name(BaseSynchronizedData.all_participants),
+            get_name(BaseSynchronizedData.period_count),
```
DavidMinarsch, line 181: There's a question whether we should say `period_count` is a precondition instead. We could also have a set of preconditions that are just provided framework side.

packages/valory/skills/termination_abci/rounds.py
```diff
@@ -200,3 +200,9 @@ class PostTerminationTxAbciApp(AbciApp[Event]):
 TerminationAbciApp.transition_function[BackgroundRound] = {
     Event.TERMINATE: TransactionSubmissionAbciApp.initial_round_cls,
 }
+TerminationAbciApp.db_pre_conditions = {
```
DavidMinarsch, line 203: Actually, I need to move this app to the post termination app and not have it on the chained app

packages/valory/skills/registration_abci/rounds.py
DavidMinarsch: Strictly speaking `period_count` is not used in this app so I will remove it and provide it framework side. Otherwise this is a bit arbitrary.

DavidMinarsch (commented):
Still quite some work to do - but: `pytest packages/valory/skills/` passes

packages/valory/skills/abstract_round_abci/tests/test_abci_app_chain.py
DavidMinarsch: Yes enforced now.

packages/valory/skills/registration_abci/rounds.py
DavidMinarsch: The three values `nb_participant`, `round_count` and `period_count` are special values provided framework side at all times.

DavidMinarsch (commented):
We need to think throw how this feature should interact with `cross_period_persisted_keys`.

packages/valory/skills/abstract_round_abci/base.py
```diff
@@ -736,6 +736,10 @@ class BaseSynchronizedData:
     This is the relevant data constructed and replicated by the agents.
     """
 
+    # Keys always set by default
+    # TODO: round_count and period_count need to be guaranteed to be synchronized too.
```
DavidMinarsch, line 740: @Adamantios thoughts on this in context of safe deployment removal?

packages/valory/skills/abstract_round_abci/base.py
```diff
@@ -736,6 +736,10 @@ class BaseSynchronizedData:
...
+    default_db_keys: List[str] = ["round_count", "period_count", "nb_participants"]
```
Adamantios, line 741: `nb_participants` does not exist in the db. Instead, it is inferred from the `participants`. I think that `round_count`, `period_count`, `participants`, and `all_participants` should be `default_db_keys` and also they should always be pre and post-conditions. Therefore, we might need to simplify the code and make it optional for the FSM to specify these as pre and post-conditions and automatically assume it and fail if that is not the case.

Adamantios: I see you already suggested what I pointed out in https://github.com/valory-xyz/open-autonomy/pull/1659#discussion_r1047713895.

packages/valory/skills/abstract_round_abci/base.py
DavidMinarsch: We can add it here once you have changed the functionality on the new PRs

packages/valory/skills/abstract_round_abci/base.py
DavidMinarsch: `nb_participants` is a weird case. I suggest we keep it as a default pre condition as it defaults to 0 when no `participants` are present.

Adamantios: Agreed! Let's keep it in mind for `all_participants` to add them as a default key in the future.

packages/valory/skills/reset_pause_abci/rounds.py
```diff
@@ -125,3 +124,8 @@ class ResetPauseAbciApp(AbciApp[Event]):
         Event.ROUND_TIMEOUT: 30.0,
         Event.RESET_AND_PAUSE_TIMEOUT: 30.0,
     }
+    db_pre_conditions: Dict[AppState, List[str]] = {ResetAndPauseRound: []}
+    db_post_conditions: Dict[AppState, List[str]] = {
+        FinishedResetAndPauseRound: [],
+        FinishedResetAndPauseErrorRound: [],
+    }
```
Adamantios, line 131: I think we can remove this.

packages/valory/skills/safe_deployment_abci/rounds.py
```diff
@@ -192,4 +191,12 @@ class SafeDeploymentAbciApp(AbciApp[Event]):
         Event.VALIDATE_TIMEOUT: 30.0,
         Event.DEPLOY_TIMEOUT: 30.0,
     }
-    cross_period_persisted_keys = [get_name(SynchronizedData.safe_contract_address)]
+    cross_period_persisted_keys: List[str] = [
+        get_name(SynchronizedData.safe_contract_address)
+    ]
+    db_pre_conditions: Dict[AppState, List[str]] = {
+        RandomnessSafeRound: [get_name(BaseSynchronizedData.participants)]
+    }
```
Adamantios, line 199:
```suggestion
    db_pre_conditions: Dict[AppState, List[str]] = {
        RandomnessSafeRound: [get_name(BaseSynchronizedData.participants)],
        DeploySafeBehaviour: [get_name(SynchronizedData.most_voted_keeper_address)]
    }
```

packages/valory/skills/abstract_round_abci/abci_app_chain.py
```diff
@@ -184,13 +184,18 @@ def get_paths(
...
+                    set(next_app.db_pre_conditions[next_initial_state]),
+                    accumulated_post_conditions,
+                )
+                if len(diff) != 0:
+                    raise ValueError(
+                        f"Pre conditions '{diff}' of app '{next_app}' not a post condition of app '{current_app}' or any preceding app in path {path}."
+                    )
+            else:
+                _default_logger.warning(
+                    f"No pre-conditions have been set for {next_initial_state}!"
```
Adamantios, line 198: Otherwise, we were getting a non-informative `KeyError`, e.g., [[0.6.0] feat: add pre- and post- conditions on db at fsm level #6358](https://github.com/valory-xyz/open-autonomy/actions/runs/3703352842/jobs/6274692428). I updated the code so that the dev can choose not to use any pre-conditions, e.g., they might be empty. We now just log a warning about it, as I thought it was more appropriate than raising.

packages/valory/skills/abstract_round_abci/base.py
DavidMinarsch: I'd replace the todo with this comment

packages/valory/skills/abstract_round_abci/abci_app_chain.py
DavidMinarsch: It should not be allowed to have no pre_conditions set for all initial states! They should explicitly set empty preconditions.

packages/valory/skills/abstract_round_abci/abci_app_chain.py
```diff
@@ -108,6 +116,90 @@ def chain(  # pylint: disable=too-many-locals,too-many-statements
...
+    # Ensure all DB pre- and post-conditions are consistent
+    # Since we know which app is the "entry-point" we can
+    # simply work forward from there through all branches. When
+    # we loop back on an earlier node we stop.
+    initial_state_to_app: Dict[AppState, Type[AbciApp]] = {}
+    for value in abci_app_transition_mapping.values():
+        for app in abci_apps:
+            if value in app.initial_states or value == app.initial_round_cls:
+                initial_state_to_app[value] = app
+                break
```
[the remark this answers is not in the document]
DavidMinarsch: I don't see what problem you are describing. `initial_round_cls` always must be defined and `initial_states` is optionally defined. Can you expand?

packages/valory/skills/reset_pause_abci/rounds.py
DavidMinarsch: With your code change below yes. It makes it less explicit.

packages/valory/skills/abstract_round_abci/abci_app_chain.py
Adamantios: Then let's convert the warning with a `ValueError` and the message `f"No pre-conditions have been set for {next_initial_state}!"`. Otherwise, the `KeyError` is confusing.

packages/valory/skills/abstract_round_abci/base.py
Adamantios: Done in c8731c93530f42dff8522e15be7b1062b38a6a57.

packages/valory/skills/abstract_round_abci/abci_app_chain.py
Adamantios: Done in b70515b4968d78a13358c311c7e43f9e82868a2e.

packages/valory/skills/abstract_round_abci/base.py
0xArdi: Actually I think it's a good idea to move them in `SynchronizedData`, they are a bit messy atm. Reset index is both defined on `AbciApp` and `AbciAppDB`. The one from `AbciApp` is used in generating the root hash. Where as the other one is used in `reset_pause_abci`. In the latter case its sent as a payload, but we never update the db. I think it would make sense to update the db with majority of the `reset_index`. Also, I don't think it makes sense to assume that all agents have the same `len(self._data)` in `AbciAppDB`, especially with rejoins (recovery) in mind.

0xArdi: One reason why we want to have `round_count` in `SynchronizedData` is becuase of background behaviours. It would help us avoid possible race conditions with concurrent rounds. We have yet to see this happen, but I think there edge cases when it could happen.

PR #1663, opened 2022-12-14
[0.7.0]: Remove the safe deployment abci

Adamantios:
## Proposed changes

Removes the safe deployment abci. This is a breaking change and should be included in `v0.7.0`.

## Fixes

n/a

## Types of changes

## Checklist

## Further comments

n/a


tests/test_docs/test_code_blocks.py
```diff
@@ -133,19 +133,21 @@ class TestYamlSnippets(BaseTestDocCode):
...
-        },
+        # TODO uncomment and update this doc, when the safe-related rounds get removed from the `price-oracle`.
+        # "docs/demos/price_oracle_fsms.md": {  # flake8: noqa: E800
+        #     "code_files": [  # flake8: noqa: E800
+        #         "packages/valory/skills/registration_abci/fsm_specification.yaml",  # flake8: noqa: E800
+        #         "packages/valory/skills/transaction_settlement_abci/fsm_specification.yaml",  # flake8: noqa: E800
+        #         "packages/valory/skills/reset_pause_abci/fsm_specification.yaml",  # flake8: noqa: E800
+        #     ],  # flake8: noqa: E800
+        #     "skip_blocks": [2, 3, 6],  # flake8: noqa: E800
+        # },  # flake8: noqa: E800
```
Adamantios, line 144: This may get out of sync at some point. Skipped the check temporarily because a new FSM will be introduced to the `price-oracle` now that the safe deployment has been removed.

DavidMinarsch (commented):
You edited the `mints` folder - please revert these changes

tests/test_docs/test_code_blocks.py
DavidMinarsch: I don't follow

packages/valory/skills/termination_abci/rounds.py
```diff
@@ -57,7 +59,7 @@ class SynchronizedData(BaseSynchronizedData):
     """
 
     @property
-    def safe_contract_address(self) -> Optional[str]:
+    def safe_contract_address(self) -> Optional[str]:  # type: ignore
```
DavidMinarsch, line 62: why not removed? it's available on base now

packages/valory/skills/abstract_round_abci/base.py
```diff
@@ -878,6 +883,11 @@ def participant_to_votes(self) -> Mapping:
         """Check whether keeper is set."""
         return cast(Dict, self.db.get_strict("participant_to_votes"))
 
+    @property
+    def safe_contract_address(self) -> str:
```
DavidMinarsch, line 887: Please check it is removed from all synchronised data children

tests/test_docs/test_code_blocks.py
Adamantios: This is a check for the code in the price oracle demo's documentation. It now fails because the safe deployment abci has been removed from this repo, but it is still in the docs' code blocks. However, the price oracle still contains it, so I am not sure if it would be correct to remove it from the demo's docs at the moment. Therefore, I skipped the test in order for us to update the docs after the safe deployment gets removed from the price oracle too.

packages/valory/skills/termination_abci/rounds.py
Adamantios: As I mentioned in f5d052115626e341099caabe478f8c71669403d7, It is still overridden here, because it is necessary for the `termination_abci` to function correctly. In essence, we cannot assume that the registration round is over when the `safe_contract_address` is called by the `TerminationRound` since it runs concurrently, from the beginning of the app.

@0xArdi can provide more insight about this.

packages/valory/skills/abstract_round_abci/base.py
Adamantios: See the comment above.

DavidMinarsch (approved):

packages/valory/skills/termination_abci/rounds.py
0xArdi: As @Adamantios said, it might happen that this is accessed before registration is over.

packages/valory/skills/safe_deployment_abci/rounds.py
```diff
@@ -1,202 +0,0 @@
...
-    }
-    final_states: Set[AppState] = {FinishedSafeRound}
-    event_to_timeout: Dict[Event, float] = {
-        Event.ROUND_TIMEOUT: 30.0,
-        Event.VALIDATE_TIMEOUT: 30.0,
-        Event.DEPLOY_TIMEOUT: 30.0,
-    }
-    cross_period_persisted_keys: List[str] = [
-        get_name(SynchronizedData.safe_contract_address)
-    ]
```
0xArdi, line 196: Now that this is removed here, do we need to add it somewhere else. I don't see any AbciApp that defines this as cross_period. Do we need to add it somewhere?

Adamantios: Yes, that would be a good idea, for the reason that you mentioned in https://github.com/valory-xyz/price-oracle/pull/23#discussion_r1053645614. Maybe we should set some default cross-period keys, and the safe address would be one of them?

DavidMinarsch: Yes let’s do that.

DavidMinarsch: All default pre conditions

packages/valory/skills/termination_abci/rounds.py
DavidMinarsch: This doesn’t explain why we need it. Now the address is a pre condition so the property as defined on the base class is sufficient. It can never be None

Adamantios: Yes, this sounds reasonable since the data will now contain the safe contract (must be in the setup data) from the db's initialization!

https://github.com/valory-xyz/open-autonomy/blob/10b215f6848c7bc25ca178038f4678cc57fca0db/packages/valory/skills/abstract_round_abci/base.py#L610-L612

packages/valory/skills/safe_deployment_abci/rounds.py
Adamantios: Setting all default pre-conditions will not work, unfortunately, because they are defined by property name:

https://github.com/valory-xyz/open-autonomy/blob/10b215f6848c7bc25ca178038f4678cc57fca0db/packages/valory/skills/abstract_round_abci/base.py#L741-L746

Instead, the cross-period keys are implemented in such a way that they must be given by their key name in the db. This also means that there is a bug. In essence, if someone creates a property with a different name than the db key that the property is accessing, and then uses the `get_name()` to set it as a cross-period persisted key, then it will not work:

https://github.com/valory-xyz/open-autonomy/blob/aa9481cab797c2fe6ae5f6a0df4ea17b014a8a71/packages/valory/skills/reset_pause_abci/rounds.py#L58-L70

Adamantios: Moreover, it does not make sense to have something called `cross_period_persisted_keys` in the db and let the `reset_pause_round` handle it. It might be better to handle it in the db's `create()` method.

packages/valory/skills/termination_abci/rounds.py
Adamantios: Done in 3d6d752e68f2d51c02150b023be951e6ec2d9e6c.

packages/valory/skills/safe_deployment_abci/rounds.py
Adamantios: Done only for the safe address in 0466577aecd6596dfede187b4de9c02fa275be51. I believe that we should open an issue for the above comments.

packages/valory/skills/abstract_round_abci/base.py
Karrenbelt: this is not covered by a unit test 
```
+    @property
+    def safe_contract_address(self) -> str:
+        """Get the safe contract address."""
+        return cast(str, self.db.get_strict("safe_contract_address"))   # <----
```

0xArdi (approved):

DavidMinarsch (approved):

PR #1665, opened 2022-12-15
[WIP]: Insert only synchronized data in the db

Adamantios:
## Proposed changes

Refactors the db and the synchronized data in such a way that the `self._data` only gets initialized after an initialization method is called by the `RegistrationStartupRound`. The initialization inserts the setup data after they have been synchronized among the agents and can only be called once, otherwise raises. Moreover, if the initialization has not been completed, other CRUD operations cannot be performed.

## Fixes

n/a

## Types of changes

## Checklist

## Further comments

n/a

### DIFF
--- a/packages/valory/skills/abstract_round_abci/base.py
+++ b/packages/valory/skills/abstract_round_abci/base.py
@@ -600,6 +600,7 @@ class AbciAppDB:
         :param setup_data: the setup data
         :param cross_period_persisted_keys: data keys that will be kept after a new period starts
         """
+        self._initialized = False
         AbciAppDB._check_data(setup_data)
         self._setup_data = deepcopy(setup_data)
         self._cross_period_persisted_keys = (
@@ -607,15 +608,15 @@ class AbciAppDB:
             if cross_period_persisted_keys is not None
             else []
         )
-        self._data: Dict[int, Dict[str, List[Any]]] = {
-            RESET_COUNT_START: self.setup_data  # the key represents the reset index
-        }
+        self._data: Dict[int, Dict[str, List[Any]]] = {}
         self._round_count = ROUND_COUNT_DEFAULT  # ensures first round is indexed at 0!
 
     @property
     def setup_data(self) -> Dict[str, Any]:
         """
         Get the setup_data without entries which have empty values.
+
+        Please keep in mind that the setup data are not synced among the agents and contain local values.
 
         :return: the setup_data
         """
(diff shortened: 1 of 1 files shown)


packages/valory/skills/abstract_round_abci/base.py
```diff
@@ -811,6 +841,27 @@ def nb_participants(self) -> int:
...
+    def initialize(self, synced_setup_data: Dict[str, List[Any]]) -> None:
+        """Initialize the synchronized data.
+
+        This method should only be used to initialize the data, once,
+        and only after the agents have reached consensus about the initial setup data.
+        `RegistrationStartupRound` is meant to utilize this method.
+
+        :param synced_setup_data: the synchronized setup data to initialize the database with.
+        """
+        self.db.initialize(synced_setup_data)
```
Adamantios, line 853: Here we initialize with synchronized data.

packages/valory/skills/abstract_round_abci/base.py
```diff
@@ -811,6 +841,27 @@ def nb_participants(self) -> int:
...
+        missing_keys = []
+        for mandatory_key in self.default_db_keys:
+            # we do not check if the attribute exists since this is done during the chaining
+            if not getattr(self, mandatory_key):
+                missing_keys.append(mandatory_key)
+
+        enforce(
+            not missing_keys,
+            f"There are missing setup data {missing_keys} after initialization!",
+        )
```
Adamantios, line 863: We enforce the `default_db_keys` to be non-null after synchronizing the setup data.

packages/valory/skills/abstract_round_abci/base.py
```diff
@@ -648,8 +649,32 @@ def cross_period_persisted_keys(self) -> List[str]:
...
+        :param synced_setup_data: the synchronized setup data to initialize the database with.
+        """
+        enforce(not self._initialized, "The database has already been initialized!")
+        enforce(
+            self.reset_index == RESET_COUNT_START,
+            "The database contains data before initialization!",
+        )
+        self._data = {
+            RESET_COUNT_START: synced_setup_data  # the key represents the reset index
+        }
```
Adamantios, line 668: The db's data are first populated here, during the initialization, with synchronized setup data.

Adamantios:
This approach is abandoned because we need these data before the startup round and Tendermint will never function properly if they are incorrect. So if we reach this point of the execution, then it is guaranteed that they are synchronized.

PR #1667, opened 2022-12-16
[0.6.0] Add support for specifying external hostname for tendermint P2P listener

angrybayblade:
## Proposed changes

This PR updates the base model params and registration skill to support specifying the external hostname for the tendermint P2P connection

## Fixes

## Types of changes

## Checklist


packages/valory/skills/abstract_round_abci/models.py
```diff
@@ -122,8 +122,8 @@ def __init__(self, *args: Any, **kwargs: Any) -> None:
         self.share_tm_config_on_startup = kwargs.pop(
             "share_tm_config_on_startup", False
         )
-        self.tendermint_p2p_port = kwargs.pop(
-            "tendermint_p2p_port", DEFAULT_TENDERMINT_P2P_PORT
+        self.tendermint_p2p_url = kwargs.pop(
+            "tendermint_p2p_url", DEFAULT_TENDERMINT_P2P_URL
```
angrybayblade, line 126: This will be helpful when establishing the tendermint chain using external/machine hosts. Since the agent and the tm node will be running in the same machine the tendermint RPC url (`tedermint_url`) can use localhost as the host but when establishing the chain we need to share either the IP address of the machine or the hostname of the machine. The `tendermint_p2p_url` will contain either the IP address or the hostname of the machine which other nodes can use to communicate.

0xArdi: Just as a side note, we don't need this now. But it would be good to update the external_address on config.toml with this address at some point. Although since we don't rely on PEX, it doesn't really matter at this point.

0xArdi (dismissed):

Adamantios (approved):

8ball030 (dismissed):
This looks like it will work to me. 
I am eager to test it

DavidMinarsch (approved):

PR #1672, opened 2022-12-16
[0.7.0]: Validate setup data

Adamantios:
## Proposed changes

Adds validation for the setup data, so that we fail early if necessary data are not provided instead of waiting for the first round to happen, and we give an informative error.

Moreover, removes the FFW round because it is not meaningful now that the setup data cannot be empty.

## Fixes

n/a

## Types of changes

## Checklist

## Further comments

n/a


packages/valory/skills/abstract_round_abci/models.py
```diff
@@ -126,13 +127,26 @@ def __init__(self, *args: Any, **kwargs: Any) -> None:
         self.setup_params = setup_params
         super().__init__(*args, **kwargs)
 
+        if not self.context.is_abstract_component:
+            # setup data are mandatory for non-abstract skills,
+            # and they should always contain at least the `safe_contract_address`
```
DavidMinarsch, line 132: And `all_participants`. Please discuss with Viraj how we can configure CLI to directly pass both values in. For e2e tests we will have to have some changes to in order to properly populate this.

packages/valory/skills/registration_abci/rounds.py
```diff
@@ -64,23 +59,16 @@ class RegistrationStartupRound(CollectSameUntilAllRound):
...
-                synchronized_data_class=self.synchronized_data_class,
-            )
-            return synchronized_data, Event.DONE
-        return None
+        if not self.collection_threshold_reached:
+            return None
+
+        synchronized_data = self.synchronized_data.update(
+            participants=frozenset(self.collection),
+            all_participants=frozenset(self.collection),
```
DavidMinarsch, line 67: This needs to change. As we know `all_participants` from the start now.

packages/valory/skills/abstract_round_abci/models.py
Adamantios: We would simply need to add `all_participants` to the `setup` params in the config and create a command like `_try_update_safe_contract_address` for `all_participants`:

https://github.com/valory-xyz/open-autonomy/blob/6395ee932295b185597f64a3934eeae56c0ae978/autonomy/deploy/base.py#L229

Adamantios:
e2e failure does not make much sense:

```
Error: Package loading error: An error occurred while loading skill valory/registration_abci:0.1.0: Traceback (most recent call last):
  File "/tmp/tmpur1u9xt_/agent_00003/vendor/valory/skills/abstract_round_abci/models.py", line 144, in _ensure_setup
    enforce(bool(self.setup_params), "`setup` params contain no values!")
  File "/home/runner/work/open-autonomy/open-autonomy/.tox/packages-py3.10-linux/lib/python3.10/site-packages/aea/exceptions.py", line 106, in enforce
    raise exception_class(exception_text)
aea.exceptions.AEAEnforceError: `setup` params contain no values!
```

However the `registration_abci` skill is abstract:

https://github.com/valory-xyz/open-autonomy/blob/8be71a59f41c7bdd896a6338ab059b7a169ba1ce/packages/valory/skills/registration_abci/skill.yaml#L130

and therefore should not execute a check for the `setup` params:

https://github.com/valory-xyz/open-autonomy/blob/5c37dc902fd7a5a896c43b464b88e95b921dc16c/packages/valory/skills/abstract_round_abci/models.py#L130-L133

packages/valory/skills/abstract_round_abci/models.py
DavidMinarsch: I suggest you do this in a new PR on top - it will affect a lot of tests

Adamantios: Done in https://github.com/valory-xyz/open-autonomy/pull/1677.

DavidMinarsch (approved):

0xArdi (approved):

PR #1685, opened 2022-12-20
Fix: use gRPC with Tendermint

Karrenbelt:
## Proposed changes

Currently, we cannot use gRPC with our Tendermint setup. This PR attempts to resolve this issue.

## Fixes

## Types of changes

## Checklist

## Further comments


deployments/Dockerfiles/tendermint/tendermint.py
```diff
@@ -162,8 +170,6 @@ def _build_init_command(self) -> List[str]:
             "tendermint",
             "init",
         ]
-        if self.params.use_grpc:
-            cmd += ["--abci=grpc"]
```
Karrenbelt, line 166: should not be used on `tendermint init`

deployments/Dockerfiles/tendermint/tendermint.py
```diff
@@ -111,6 +118,7 @@ def build_node_command(self, debug: bool = False) -> List[str]:
             f"--p2p.laddr={self.p2p_laddr}",
             f"--p2p.seeds={p2p_seeds}",
             f"--consensus.create_empty_blocks={str(self.consensus_create_empty_blocks).lower()}",
+            f"--abci={'grpc' if self.use_grpc else 'socket'}",
```
Karrenbelt, line 121: should be used on `tendermint node`. We make it explicit, also when using the default: TCP (socket)

packages/valory/connections/abci/connection.py
```diff
@@ -1358,7 +1369,9 @@ def _process_tendermint_params(self) -> None:
         if not self.use_tendermint:
             return
         tendermint_config = self.configuration.config.get("tendermint_config", {})
-        rpc_laddr = cast(str, tendermint_config.get("rpc_laddr", DEFAULT_RPC_PORT))
+        rpc_laddr = cast(
+            str, tendermint_config.get("rpc_laddr", DEFAULT_RPC_LISTEN_ADDRESS)
+        )
```
Karrenbelt, line 1374: should be full address, not only port

packages/valory/connections/abci/tests/test_abci.py
```diff
@@ -169,6 +170,17 @@ def _update_dialogues(self, request: AbciMessage) -> AbciDialogue:
...
+    def echo(self, request: AbciMessage) -> AbciMessage:
+        """Process echo request"""
+
+        abci_dialogue = self._update_dialogues(request)
+        reply = abci_dialogue.reply(
+            performative=AbciMessage.Performative.RESPONSE_ECHO,
+            target_message=request,
+            message=request.message,
+        )
+        return cast(AbciMessage, reply)
```
Karrenbelt, line 182: we need to respond to Tendermint echo request, but was not implemented.

packages/valory/connections/abci/tests/test_abci.py
```diff
@@ -327,8 +339,19 @@ def make_app(self) -> ABCIAppTest:
...
+    container = image.create()
+    container.start()
+    logging.info(f"Setting up image {image.image}...")
+    success = image.wait()
+    logging.info(f"TendermintDockerImage running: {success}...")
+
+
 @pytest.mark.integration
-@pytest.mark.usefixtures("tendermint")
 class BaseTestABCITendermintIntegration(BaseThreadedAsyncLoop, ABC):
```
Karrenbelt, line 355: We cannot use the Tendermint autouse fixture, because gRPC only sends a single echo request. Hence, the connection needs to be ready to receive it when the docker image is launched.

packages/valory/connections/abci/tests/test_abci.py
```diff
@@ -346,17 +369,22 @@ class BaseTestABCITendermintIntegration(BaseThreadedAsyncLoop, ABC):
...
             "name", address=self.ADDRESS, public_key=self.PUBLIC_KEY
         )
+
         self.configuration = ConnectionConfig(
             connection_id=ABCIServerConnection.connection_id,
             host=DEFAULT_LISTEN_ADDRESS,
             port=DEFAULT_ABCI_PORT,
             target_skill_id=self.TARGET_SKILL_ID,
             use_tendermint=False,
+            use_grpc=False,
```
Karrenbelt, line 385: still at False, which is was, now simply made explicit. I've been attempting to get this to work with `use_grpc=True`

packages/valory/connections/abci/connection.py
```diff
@@ -1353,12 +1364,16 @@ def _process_tendermint_params(self) -> None:
         self.use_tendermint = cast(
             bool, self.configuration.config.get("use_tendermint")
         )
-        self.use_grpc = cast(bool, self.configuration.config.get("use_grpc"))
+        if not isinstance(self.use_tendermint, bool):
+            raise TypeError(f"use_tendermint must be bool, got {self.use_tendermint}")
+        self.use_grpc = cast(bool, self.configuration.config.get("use_grpc", False))
 
         if not self.use_tendermint:
             return
```
Karrenbelt, line 1372: use_grpc was assumed to be a boolean. This was not necessarily so. `(cast` should be used carefully. Since we assume this I decided to add a check, forcing assignment and not silently passing when `None`.

packages/valory/connections/abci/tests/test_abci.py
```diff
@@ -344,19 +370,26 @@ class BaseTestABCITendermintIntegration(BaseThreadedAsyncLoop, ABC):
     PUBLIC_KEY = "agent_public_key"
     tendermint_port = DEFAULT_TENDERMINT_PORT  # noqa: F811
 
-    def setup(self) -> None:
+    @pytest.fixture(autouse=True, params=[False, True])
+    def setup_and_teardown(self, request: SubRequest) -> Generator:
```
Karrenbelt, line 374: @Adamantios

0xArdi (approved):

deployments/Dockerfiles/tendermint/tendermint.py
```diff
@@ -30,9 +30,10 @@
 from typing import Any, Dict, List, Optional
 
 
+_TCP = "tcp://"
 ENCODING = "utf-8"
-DEFAULT_P2P_LISTEN_ADDRESS = "tcp://0.0.0.0:26656"
-DEFAULT_RPC_LISTEN_ADDRESS = "tcp://0.0.0.0:26657"
+DEFAULT_P2P_LISTEN_ADDRESS = f"{_TCP}0.0.0.0:26656"
+DEFAULT_RPC_LISTEN_ADDRESS = f"{_TCP}0.0.0.0:26657"
```
Adamantios, line 36: Why create a constant for that and not also for `0.0.0.0` for example? Such as in:

https://github.com/valory-xyz/open-autonomy/blob/51c6881b10fb44f7e3273cf90f5366d268b66c93/packages/valory/connections/abci/connection.py#L99

packages/valory/connections/abci/tests/test_abci.py
```diff
@@ -369,9 +402,23 @@ def setup(self) -> None:
             self.process_incoming_messages()
         )
 
+        # connection must be established,
+        # unlike TCP, only a single ECHO request is send with gRPC
+        time.sleep(5)
```
Adamantios, line 407: Why 5?

Adamantios (approved):

deployments/Dockerfiles/tendermint/tendermint.py
Karrenbelt: not relevant, it's a remnant from earlier introduction of `grpc://` as a prefix. Would be nice though to collect all of this in one place and using something like [urlparse](https://docs.python.org/3/library/urllib.parse.html)

packages/valory/connections/abci/tests/test_abci.py
Karrenbelt: I'm not sure this is true (that we cannot fire up the image before starting the connecting), but I was informed by @0xArdi this was the case. Doesn't align with implementation in #1687 though where the autouse fixture fires before agents set and start up.

DavidMinarsch (approved):

PR #1708, opened 2023-01-03
contract packages moved from data dir to packages

solarw:
## Proposed changes

contract packages moved from data dir to packages

## Types of changes

## Checklist

## Further comments


setup.py
```diff
@@ -97,7 +97,10 @@ def parse_readme():
                 "data/Dockerfiles/hardhat/*",
                 "data/Dockerfiles/tendermint/*",
                 "test_tools/data/*",
-            ]
+                "../packages/valory/contracts/component_registry/*"
+                "../packages/valory/contracts/registries_manager/*"
+            ],
+
         },
```
DavidMinarsch, line 104: ref

autonomy/chain/mint.py
```diff
@@ -118,9 +118,10 @@ def verify_and_fetch_token_id_from_event(
 
 def get_contract(public_id: PublicId) -> Contract:
     """Load contract for given public id."""
-
-    package_dir = CONTRACTS_DIR / public_id.name
-    return Contract.from_dir(directory=package_dir)
+    contract_dir = RELATIVE_CONTRACTS_DIR / public_id.name
+    if contract_dir.exists():
+        return contract_dir
+    return ROOT_DIR / contract_dir
```
DavidMinarsch, line 124: Shouldn't we look in the data directory in that case? Because this will be the case of the framework being installed via pypi and so you have the packages in the data dir: see comment "ref"

solarw: @DavidMinarsch @angrybayblade 
PR updated:
it uses extra package named "packages" and intalled with `autonomy`
so code works and contracts relative palce stays the same.

moving some packages to `data` dir during the pacakge buuld process looks quite tricky if possible

DavidMinarsch (approved):

PR #1722, opened 2023-01-09
[WIP] Refactor/payload

Karrenbelt:
## Proposed changes

Base PR showcasing draft implementation of payload redesign. 
Please look at current implementation of [`BaseTxPayload`](https://github.com/valory-xyz/open-autonomy/blob/refactor/payload/packages/valory/skills/abstract_round_abci/base.py#L205-L311) and [`Transaction`](https://github.com/valory-xyz/open-autonomy/blob/4d623d97896c7ad3675cb82de07d3dcd7099c38a/packages/valory/skills/abstract_round_abci/base.py#L314-L361) for comparison.
WD can be found [here](https://docs.google.com/document/d/1rb8MOJwQngmDo3tKTWQPYZFCgGyaWlmVQu2HFbQ9-w8)

TODO (not into this PR):
1. replace `transaction_type` -> `payload_class` & remove `payload_attribute` & `allowed_tx_type`.
   - remove `transaction_type` entirely from `Payload` and `Round`
   - set `payload_class` on `Round` instead.
   - make necessary changes to `_MetaPayload`. Use `key = cls.__name__, cls.__module__` for the class registry.
   - all places where as use `getattr(payload, self.payload_attribute)` to obtain the content to instead fetch all the fields of the payload. The entire payload as a dict. This will be a considerable change, now we may use `getattr` to retrieve `Any`, a primitive for example. Instead we will retrieve a dict containing all fields of the payload. We re-instantiate a payload for clarity and type checking purposes (though fields like `sender` will be meaningless here, perhaps can derive a `TypedDict` from the dataclass definition directly, will have to see what is possible here).  
2. remove remaining methods from `BaseTxPayload`
   - all serialization logic to `Transaction`. Data class should be as (c)lean as possible.
   - move uuid logic outside of payload (prefix in [`deliver_tx` ](https://github.com/valory-xyz/open-autonomy/blob/main/packages/valory/skills/abstract_round_abci/handlers.py#L192), remove prefix when reading. Purpose is to ensure Tendermint doesn't detect the payload as a unique one)
3. expose `payload_class` directly on `BaseBehaviour` via `matching_round` to accommodate payload initialization.
4. put back `Optional` on Registration payload `initialisation` (see https://github.com/valory-xyz/open-autonomy/pull/1734#discussion_r1067162954)

## Fixes

## Types of changes

## Checklist

## Further comments


packages/valory/skills/abstract_round_abci/base.py
```diff
@@ -355,6 +361,55 @@ def __eq__(self, other: Any) -> bool:
         return self.payload == other.payload and self.signature == other.signature
 
 
+@dataclass(frozen=True)
+class NewBaseTxPayload(ABC, metaclass=_MetaPayload):
+    """This class represents a base class for transaction payload classes."""
+
+    sender: str
+    round_count: int = field(default=ROUND_COUNT_DEFAULT, init=False)
```
Karrenbelt, line 369: just a plain dataclass and nothing more. No extra methods for serialization or anything. 
- The `round_count` field cannot be set during initialization and hence will have the default value, which we set in `send_a2a_transaction`. Reason for providing it like this has two reasons:
  1. `round_count` is for internal use and not generally relevant to devs, neither should they set it in the first place.
      https://github.com/valory-xyz/open-autonomy/blob/4d623d97896c7ad3675cb82de07d3dcd7099c38a/packages/valory/skills/abstract_round_abci/behaviour_utils.py#L703
  2. providing a default with `init=False` allows to subclass without the necessity to provide default value (would be necessary without default value) and without having to specify the `round_count`. To illustrate
        ```
        @dataclass(frozen=True)
        class MyPayload(NewBaseTxPayload):
            a : int
        ```
        would not be possible to initialize as follows
        ```
        MyPayload("s", 2)
        >>> TxPay(sender='s', round_count=-1, a=2)
        ```
- 


 Current implementation here
https://github.com/valory-xyz/open-autonomy/blob/4d623d97896c7ad3675cb82de07d3dcd7099c38a/packages/valory/skills/abstract_round_abci/base.py#L205-L311

packages/valory/skills/abstract_round_abci/base.py
```diff
@@ -355,6 +361,55 @@ def __eq__(self, other: Any) -> bool:
...
+
+
+@dataclass
+class NewTransaction:
+    """Class to represent a transaction for the ephemeral chain of a period."""
+
+    payload: NewBaseTxPayload
+    signature: Optional[str] = None
```
Karrenbelt, line 377: This class I propose to treat as a dataclass as well. It can contain the immutable payload which can be signed by an agent. Here we can move all operations related to serialization, deserialization and verification of the payload.

packages/valory/skills/abstract_round_abci/base.py
```diff
@@ -355,6 +361,55 @@ def __eq__(self, other: Any) -> bool:
...
+    def __bytes__(self) -> bytes:
+        """Serialize the transaction to bytes"""
+
+        payload = {self.payload.__class__.__name__: asdict(self.payload)}
+        tx_data = dict(payload=payload, signature=self.signature)
+        encoded_data = json.dumps(tx_data, sort_keys=True).encode()
+        if sys.getsizeof(encoded_data) > MAX_READ_IN_BYTES:
+            msg = f"Transaction must be smaller than {MAX_READ_IN_BYTES} bytes"
+            raise ValueError(msg)
+        return encoded_data
```
Karrenbelt, line 388: simply serialize to json. Note that, unlike with pickle, this does not allow for a one-to-one mapping between datastructures (e.g. `assert isinstance(json.loads(json.dumps(tuple())), list)`).

packages/valory/skills/abstract_round_abci/base.py
```diff
@@ -355,6 +361,55 @@ def __eq__(self, other: Any) -> bool:
...
+        """Deserialize the transaction from bytes"""
+
+        data = json.loads(obj.decode())
+        payload, signature = data["payload"], data["signature"]
+        cls_name, kwargs = payload.popitem()
+        round_count = kwargs.pop("round_count")
+        payload_cls = _MetaPayload.transaction_type_to_payload_cls[cls_name]
+        payload = payload_cls(**kwargs)
+        object.__setattr__(payload, "round_count", round_count)
+        return cls(payload=cast(NewBaseTxPayload, payload), signature=signature)
```
Karrenbelt, line 401: `round_count` must be set using `object` since its own setter will raise a `dataclasses.FrozenInstanceError`.

packages/valory/skills/abstract_round_abci/base.py
```diff
@@ -952,6 +1007,8 @@ def _check_required_class_attributes(
                 f"'allowed_tx_type' not set on {abstract_round_cls}"
             )
         if not hasattr(abstract_round_cls, "payload_attribute"):
+            if hasattr(abstract_round_cls, "payload"):
+                return
```
Karrenbelt, line 1011: should not be here, part of current work relevant to next PR.

packages/valory/skills/abstract_round_abci/base.py
Adamantios: This should be a comment in the code.

Adamantios: Why set `init=False` here:

https://github.com/valory-xyz/open-autonomy/blob/4d623d97896c7ad3675cb82de07d3dcd7099c38a/packages/valory/skills/abstract_round_abci/base.py#L369

instead of setting it to `True` and not using `object.__setattr__` but keep the `round_count` in the `kwargs`?

Karrenbelt: please see my previous comment here: https://github.com/valory-xyz/open-autonomy/pull/1722#discussion_r1065169704 where I explain that we can otherwise not use non-default arguments in any of the concrete payload implementations.

packages/valory/skills/abstract_round_abci/behaviour_utils.py
```diff
@@ -699,9 +699,10 @@ def send_a2a_transaction(
         :yield: the responses
         """
         stop_condition = self.is_round_ended(self.matching_round.auto_round_id())
-        payload.round_count = cast(
+        round_count = cast(
             SharedState, self.context.state
         ).synchronized_data.round_count
+        object.__setattr__(payload, "round_count", round_count)
```
DavidMinarsch, line 705: Can we wrap this in a helper method on the base class? Just to expose less internals..

Karrenbelt: I can make it a private method to dismay the developer from using it

Karrenbelt: Not a good idea, then we expose a method which we don't want developers to use. It should only be used in `abstract_round_abci`, which is the case currently, and outside of tests / test tools the `behaviour_utils.py` is the the only place it is used, once.

packages/valory/skills/abstract_round_abci/base.py
```diff
@@ -198,113 +197,59 @@ def _get_field(mcs, cls: Type, field_name: str) -> Any:
...
-        """
-        self.id_ = uuid.uuid4().hex if id_ is None else id_
-        self._round_count = round_count
-        self._sender = sender
-        # kwargs exists for typing only
-        enforce(len(kwargs) == 0, f"kwargs passed to {type(self)} must be empty")
+    sender: str
+    transaction_type: Any = field(init=False)
+    round_count: int = field(default=ROUND_COUNT_DEFAULT, init=False)
+    id_: str = field(default_factory=lambda: uuid.uuid4().hex, init=False)
```
Karrenbelt, line 207: `transaction_type` and `id_` to be removed later.

packages/valory/skills/abstract_round_abci/base.py
```diff
@@ -198,113 +197,59 @@ def _get_field(mcs, cls: Type, field_name: str) -> Any:
...
 
     @property
-    def sender(self) -> str:
-        """Get the sender."""
-        return self._sender
+    def data(self) -> Dict[str, Any]:
+        """Data"""
+        excluded = ["sender", "round_count", "id_", "transaction_type"]
+        return {k: v for k, v in asdict(self).items() if k not in excluded}
```
Karrenbelt, line 213: `content` might be a better nomer. It concerns everything that is relevant for use in the application. ignoring `id_` and `transaction_type` as they will be removed, both `sender` and `round_count` are only relevant for payload processing (metadata so you will); it is not part of what the agents need to reach consensus on.

packages/valory/skills/abstract_round_abci/base.py
```diff
@@ -198,113 +197,59 @@ def _get_field(mcs, cls: Type, field_name: str) -> Any:
...
-
-    @classmethod
-    def decode(cls, obj: bytes) -> "BaseTxPayload":
-        """Decode the payload."""
-        return cls.from_json(DictProtobufStructSerializer.decode(obj))
+    def json(self) -> Dict[str, Any]:
+        """Json"""
+        data = asdict(self)
+        data["transaction_type"] = str(data["transaction_type"])
+        return data
```
Karrenbelt, line 221: serialization can be dealt with here (class currently not used)
https://github.com/valory-xyz/open-autonomy/blob/22dcb96cbb033ad0c6e7109f15c70c5141250239/packages/valory/skills/abstract_round_abci/base.py#L305-L344

packages/valory/skills/abstract_round_abci/common.py
```diff
@@ -130,7 +130,7 @@ def async_act(self) -> Generator:
                 self.context.logger.info(f"Retrieved DRAND values: {observation}.")
 
         if observation:
-            payload = self.payload_class(
+            payload = self.payload_class(  # type: ignore
```
Karrenbelt, line 133: because it thinks it's a `Type[BaseTxPayload]` it does not recognize the arguments passed for anything else but the `sender`. Associating this directly via the `AbstractRound` and making accessible via `BaseBehaviour.matching_round` should presumably alleviate this.

packages/valory/skills/hello_world_abci/payloads.py
```diff
@@ -39,131 +39,46 @@ def __str__(self) -> str:
...
 
 
+@dataclass(frozen=True)
 class BaseHelloWorldAbciPayload(BaseTxPayload, ABC):
     """Base class for the Hello World abci demo."""
 
-    def __hash__(self) -> int:
-        """Hash the payload."""
-        return hash(tuple(sorted(self.data.items())))
-
```
Karrenbelt, line 48: whole base class can be removed now

packages/valory/skills/hello_world_abci/tests/test_payloads.py
```diff
@@ -76,12 +74,12 @@ def test_print_message_payload() -> None:
 def test_reset_payload() -> None:
     """Test `ResetPayload`"""
 
-    payload = ResetPayload(sender="sender", period_count=1, id_="id")
+    payload = ResetPayload(sender="sender", period_count=1)
 
     assert payload.period_count == 1
-    assert payload.id_ == "id"
+    assert payload.id_
```
Karrenbelt, line 80: can no longer be set on initialization. Will be automatically generated on instantiation.

packages/valory/skills/hello_world_abci/tests/test_payloads.py
```diff
@@ -76,12 +74,12 @@ def test_print_message_payload() -> None:
...
     assert payload.data == {"period_count": 1}
-    assert hash(payload) == hash(tuple(sorted(payload.data.items())))
+    assert hash(payload)
```
Karrenbelt, line 82: no more custom `__hash__` method present. Should remain hashable though since the data structure now is immutable, and that is what we care about

packages/valory/skills/registration_abci/behaviours.py
```diff
@@ -72,7 +72,7 @@ def async_act(self) -> Generator:
             initialisation = (
                 json.dumps(self.synchronized_data.db.setup_data, sort_keys=True)
                 if self.synchronized_data.db.setup_data != {}
-                else None
+                else ""
```
DavidMinarsch, line 75: See previous comment. Suggest to make empty dict serialised

packages/valory/skills/abstract_round_abci/base.py
```diff
@@ -355,6 +302,48 @@ def __eq__(self, other: Any) -> bool:
         return self.payload == other.payload and self.signature == other.signature
 
 
+@dataclass
+class NewTransaction:
```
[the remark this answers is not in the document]
Karrenbelt: this was here to indicate intention, to highlight that I planned to move all serialization logic from payload to transaction. Can be removed (and kept on shelf for later).

packages/valory/skills/abstract_round_abci/base.py
```diff
@@ -948,10 +937,14 @@ def _check_required_class_attributes(
                 f"'synchronized_data_class' not set on {abstract_round_cls}"
             )
         if not hasattr(abstract_round_cls, "allowed_tx_type"):
+            if hasattr(abstract_round_cls, "payload"):
+                return
```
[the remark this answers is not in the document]
Karrenbelt: also a remnant related to a subsequent PR where I replaced `payload_attribute` and so on with `payload(_class)`
see https://github.com/valory-xyz/open-autonomy/pull/1722#discussion_r1065184772

packages/valory/skills/registration_abci/behaviours.py
Karrenbelt: ok, so effectively `json.dumps({}, sort_keys=True)` when `setup_data` is an empty dict.

Karrenbelt:
1. error in unit test is something related to hypothesis time outs - I'm not touching that
2. e2e test failure - didn't investigate
    ```
    2023-01-11T22:56:03.4392900Z =================================== FAILURES ===================================
    2023-01-11T22:56:03.4393248Z _________________________ TestTermination.test_run[4] __________________________
    2023-01-11T22:56:03.4393425Z 
    2023-01-11T22:56:03.4393687Z self = <packages.valory.agents.register_termination.tests.test_register_reset.TestTermination object at 0x7f1585d08610>
    ```

packages/valory/skills/abstract_round_abci/common.py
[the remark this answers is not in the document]
Karrenbelt: hmm no I cannot cast it since the payload is not known at abstract base class level. There was an implicit assumption here that went unnoticed by mypy but has come to light now that we use dataclasses

packages/valory/skills/abstract_round_abci/base.py
```diff
@@ -198,113 +199,59 @@ def _get_field(mcs, cls: Type, field_name: str) -> Any:
...
-            return NotImplemented
-        return (
-            self.id_ == other.id_
-            and self.round_count == other.round_count
-            and self.sender == other.sender
-            and self.data == other.data
-        )
+    def encode(self) -> bytes:
+        """Encode"""
+        encoded_data = json.dumps(self.json).encode()
```
DavidMinarsch, line 245:
```suggestion
        encoded_data = json.dumps(self.json, sort_keys=True).encode()
```

packages/valory/skills/abstract_round_abci/base.py
```diff
@@ -198,113 +199,59 @@ def _get_field(mcs, cls: Type, field_name: str) -> Any:
...
+        object.__setattr__(payload, "round_count", round_count)
+        object.__setattr__(payload, "id_", id_)
+        return payload
 
     def with_new_id(self) -> "BaseTxPayload":
         """Create a new payload with the same content but new id."""
-        return type(self)(
-            self.sender, id_=uuid.uuid4().hex, round_count=self.round_count, **self.data
-        )
+        new = type(self)(sender=self.sender, **self.data)  # type: ignore
```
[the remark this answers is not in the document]
Karrenbelt: it will still complain about the extra arguments `**self.data`:
`packages/valory/skills/abstract_round_abci/base.py:239: error: Too many arguments for "BaseTxPayload"`

Karrenbelt:
- e2e failure: `TestHelloWorldABCIFourAgentsCatchupOnRegister.test_run[4]`
    ```
    2023-01-12T11:01:00.5778251Z [9432]stdout>[2023-01-12 10:28:43,252] [DEBUG] [agent_00001] Handling envelope: Envelope(to=valory/hello_world_abci:0.1.0, sender=valory/http_client:0.23.0, protocol_specification_id=valory/http:1.0.0, message=Message(sender=valory/http_client:0.23.0,to=valory/hello_world_abci:0.1.0,body=b'{\n  "jsonrpc": "2.0",\n  "id": -1,\n  "error": {\n    "code": -32603,\n    "message": "Internal error",\n    "data": "tx (357C638121E4BF89C1EFB1131F0D20DC824A5249C037C2E601E1B9B91EA0E9EC) not found"\n  }\n}',dialogue_reference=('50fd4fa99bf3ae6f16faf501ff8d8312b037ce368c5cf68b394d942eb83161fe', '2ee922cc0c09ec3fb2bcd8606d6791de8acf69f05119893ce037fa64092d770d'),headers=Content-Type: application/json
    ```

DavidMinarsch (commented):
Please update scaffolding to match

packages/valory/skills/abstract_round_abci/base.py
DavidMinarsch, line 239:
```suggestion
        new = type(self)(sender=self.sender, **self.data)  # type: ignore
```
is missing the `round_count` and `id_`!

packages/valory/skills/registration_abci/payloads.py
```diff
@@ -34,33 +34,9 @@ def __str__(self) -> str:
         return self.value
 
 
+@dataclass(frozen=True)
 class RegistrationPayload(BaseTxPayload):
     """Represent a transaction payload of type 'registration'."""
 
+    initialisation: str
```
DavidMinarsch, line 41: Why did we change this away from `Optional[str]`?

packages/valory/skills/hello_world_abci/payloads.py
DavidMinarsch: Yes please remove it

packages/valory/skills/registration_abci/payloads.py
Karrenbelt: because we cannot serialize `None` using protobuf. Can revert this and move away from protobuf

packages/valory/skills/abstract_round_abci/base.py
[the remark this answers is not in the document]
DavidMinarsch: You could use https://docs.python.org/3/library/dataclasses.html#init-only-variables and add `kwargs: InitVar[Optional[Dict]] = None` on `BaseTxPayload` to enable `kwargs` without adding extra fields accessible on the inheriting class.

packages/valory/skills/abstract_round_abci/base.py
```diff
@@ -266,7 +266,7 @@ def encode(self) -> bytes:
         """Encode the transaction."""
 
         data = dict(payload=self.payload.json, signature=self.signature)
-        encoded_data = DictProtobufStructSerializer.encode(data)
+        encoded_data = json.dumps(data, sort_keys=True).encode()
```
DavidMinarsch, line 266: Please remove the whole file `serializer.py`

packages/valory/skills/abstract_round_abci/base.py
Karrenbelt: - `id_`
https://github.com/valory-xyz/open-autonomy/blob/7a18eb4479a26891fe3e807abebea4fd4d5f9e32/packages/valory/skills/abstract_round_abci/base.py#L206
- `round_count`
https://github.com/valory-xyz/open-autonomy/blob/7a18eb4479a26891fe3e807abebea4fd4d5f9e32/packages/valory/skills/abstract_round_abci/base.py#L234-L238

packages/valory/skills/abstract_round_abci/common.py
Karrenbelt: yes. You know this is where protocols come in nicely. Not the kind we use in our framework for defining speech acts and so on, but the types one uses to define the structure of an interface (as an alternative to an abstract base class).

DavidMinarsch (approved):

Karrenbelt:
same e2e test issue as before, involving `TestHelloWorldABCIFourAgentsCatchupOnRegister.test_run[4]` with the same types of failures as reported here: https://github.com/valory-xyz/open-autonomy/pull/1722#issuecomment-1380309271

Adamantios (approved):

PR #1720, opened 2023-01-09
[WIP]: Refactor tm protocol

Adamantios:
## Proposed changes

Refactor tendermint protocol and handler (which are in fact for ACN communication), in order to add support for sharing the recovery parameters.

## Fixes

n/a

## Types of changes

What types of changes does your code introduce? (A **breaking change** is a fix or feature that would cause existing functionality and APIs to not work as expected.)
_Put an `x` in the box that applies_

- [ ] Non-breaking fix (non-breaking change which fixes an issue)
- [ ] Breaking fix (breaking change which fixes an issue)
- [ ] Non-breaking feature (non-breaking change which adds functionality)
- [ ] Breaking feature (breaking change which adds functionality)
- [x] Refactor (non-breaking change which changes implementation)
- [ ] Messy (mixture of the above - requires explanation!)

## Checklist

_Put an `x` in the boxes that apply._

- [x] I have read the [CONTRIBUTING](../blob/main/CONTRIBUTING.md) doc
- [x] I am making a pull request against the `main` branch (left side). Also you should start your branch off our `main`.
- [x] Lint and unit tests pass locally with my changes
- [x] I have added tests that prove my fix is effective or that my feature works
- [ ] I have locally run services that could be impacted and they do not present failures derived from my changes

## Further comments

n/a

PR #1723, opened 2023-01-10
fix: flaky fuzzer tests

0xArdi:
## Fixes

Flaky fuzzer abci tests.

## Types of changes

What types of changes does your code introduce? (A **breaking change** is a fix or feature that would cause existing functionality and APIs to not work as expected.)
_Put an `x` in the box that applies_

- [x] Non-breaking fix (non-breaking change which fixes an issue)
- [ ] Breaking fix (breaking change which fixes an issue)
- [ ] Non-breaking feature (non-breaking change which adds functionality)
- [ ] Breaking feature (breaking change which adds functionality)
- [ ] Refactor (non-breaking change which changes implementation)
- [ ] Messy (mixture of the above - requires explanation!)

## Checklist

_Put an `x` in the boxes that apply._

- [x] I have read the [CONTRIBUTING](../blob/main/CONTRIBUTING.md) doc
- [x] I am making a pull request against the `main` branch (left side). Also you should start your branch off our `main`.
- [x] Lint and unit tests pass locally with my changes
- [x] I have added tests that prove my fix is effective or that my feature works
- [x] I have locally run services that could be impacted and they do not present failures derived from my changes

PR #1724, opened 2023-01-10
Refactor recovery parameters

Adamantios:
## Proposed changes

Refactors the recovery parameters.

The `TendermintRecoveryParams` were holding the class of the round for the `reset_from_round`. However, in order to share this information via the ACN, we need to be able to serialize it. Therefore, in this PR, it is replaced with the round id instead.

Moreover, tests are added in this PR. These concern the new methods of the tm handler (introduced in #1720) which will enable us to share the recovery parameters information.

## Fixes

n/a

## Types of changes

## Checklist

## Further comments

n/a


0xArdi (approved):

0xArdi:
macOS failures are resolved by #1723 . The windows failure is known to be flaky.

packages/valory/skills/abstract_round_abci/behaviour_utils.py
```diff
@@ -1854,7 +1854,7 @@ def reset_tendermint_with_wait(  # pylint: disable=too-many-locals, too-many-sta
                             reset_params=reset_params,
                             reset_index=reset_index,
                             round_count=round_count,
-                            reset_from_round=restart_from_round,
+                            reset_from_round=restart_from_round.auto_round_id(),
```
DavidMinarsch, line 1857: Was this a bug?

packages/valory/skills/abstract_round_abci/models.py
```diff
@@ -175,7 +174,7 @@ def __init__(
         self.abci_app_cls._is_abstract = skill_context.is_abstract_component
         self._round_sequence: Optional[RoundSequence] = None
         self.tm_recovery_params: TendermintRecoveryParams = TendermintRecoveryParams(
-            self.abci_app_cls.initial_round_cls
+            self.abci_app_cls.initial_round_cls.auto_round_id()
```
DavidMinarsch, line 177: Ok so we now use the round id

packages/valory/skills/abstract_round_abci/behaviour_utils.py
Adamantios: No, it was not, but:

> The TendermintRecoveryParams were holding the class of the round for the reset_from_round. However, in order to share this information via the ACN, we need to be able to serialize it. Therefore, in this PR, it is replaced with the round id instead.

PR #1734, opened 2023-01-11
Refactor: payloads throughout skills

Karrenbelt:
## Proposed changes

Changing all payload to dataclasses throughout

TODO:
- merge main
- metaclass and test
- mypy

## Fixes

## Types of changes

## Checklist

## Further comments


packages/valory/skills/abstract_round_abci/base.py
```diff
@@ -164,43 +164,14 @@ def __new__(mcs, name: str, bases: Tuple, namespace: Dict, **kwargs: Any) -> Typ
...
-    ) -> None:
-        """Check that a transaction type is not already associated to a concrete payload class."""
-        if transaction_type in mcs.transaction_type_to_payload_cls:
-            previous_payload_cls = mcs.transaction_type_to_payload_cls[transaction_type]
-            if new_payload_cls.__name__ != previous_payload_cls.__name__:
-                raise ValueError(
-                    f"transaction type with name {transaction_type} already "
-                    f"used by class {previous_payload_cls}, and cannot be "
-                    f"used by class {new_payload_cls} "
-                )
```
Karrenbelt, line 195: So there either is an issue in the current design or I misunderstand it. Now `transaction_type` and `payload.__name__` need to match. However, that allows for an overwrite which I believe should not happen. To illustrate, the following is possible:

```
class Payload(BaseTxPayload):
    transaction_type = "tx_type"

class Payload(BaseTxPayload):
    transaction_type = "tx_type"
    other_field = "other"
```

packages/valory/skills/abstract_round_abci/base.py
```diff
@@ -164,43 +164,14 @@ def __new__(mcs, name: str, bases: Tuple, namespace: Dict, **kwargs: Any) -> Typ
...
                 f"class {name} must inherit from {BaseTxPayload.__name__}"
             )
-        new_cls = cast(Type[BaseTxPayload], new_cls)
-
-        transaction_type = str(mcs._get_field(new_cls, "transaction_type"))
-        mcs._validate_transaction_type(transaction_type, new_cls)
-        # remember association from transaction type to payload class
-        mcs.transaction_type_to_payload_cls[transaction_type] = new_cls
 
+        mcs.transaction_type_to_payload_cls[new_cls.__name__] = new_cls
```
Karrenbelt, line 172: What I tried to use is the following, but it results in errors as we have several modules where we define payloads with the same name as defined elsewhere. However I believe that this is what we want, otherwise we cannot guarantee we can load the payload from bytes correctly.
```
cls = mcs.transaction_type_to_payload_cls.get(new_cls.__name__)
if cls and cls is not new_cls:
    raise ValueError
```

packages/valory/skills/abstract_round_abci/tests/test_base.py
```diff
@@ -353,48 +341,6 @@ def test_dict_serializer_is_deterministic(obj: Any) -> None:
         assert obj == DictProtobufStructSerializer.decode(obj_bytes)
 
 
-class TestMetaPayloadUtilityMethods:
-    """Test _MetaPayload private utility methods."""
```
Karrenbelt, line 357: will need to be implemented in next PR when the purpose is clarified.

packages/valory/skills/registration_abci/payloads.py
```diff
@@ -34,33 +34,9 @@ def __str__(self) -> str:
         return self.value
 
 
+@dataclass(frozen=True)
 class RegistrationPayload(BaseTxPayload):
     """Represent a transaction payload of type 'registration'."""
 
+    initialisation: str
```
Karrenbelt, line 41: No longer optional. Optional fields should not be used in the current implementation, since the DictToProtobuf serializer cannot deal with `None`s

packages/valory/skills/abstract_round_abci/base.py
Karrenbelt: I actually was casting to dataclass here initially, as it removed the need for devs to do this
`freezing_payload_class = dataclass(new_cls, frozen=True)`, however the linters don't like it; they don't understand that the class attribute are in fact parameters of the initializer.

packages/valory/skills/registration_abci/payloads.py
DavidMinarsch: We're removing DictToProtobuf though, so just add a TODO so we can fix this in the next wave

packages/valory/skills/abstract_round_abci/base.py
DavidMinarsch: You highlight an important issue. We have the same question with Events. It would totally hinder composability if the developer had to think about clashes outside their own skill. So I propose we include two pieces of info into the mapping: class name & skill id (without hash). That way, we can guarantee uniqueness even when two payloads have the same class name in different skills.

packages/valory/skills/abstract_round_abci/base.py
```diff
@@ -164,43 +164,14 @@ def __new__(mcs, name: str, bases: Tuple, namespace: Dict, **kwargs: Any) -> Typ
...
         return new_cls
 
-    @classmethod
-    def _validate_transaction_type(
```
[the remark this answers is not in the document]
DavidMinarsch: See my other response, keep it and add the skill id to the chck

packages/valory/skills/abstract_round_abci/base.py
Karrenbelt: Then we end up with the same issue though, we just move it to the transaction type. Now you can name payload classes identically, but not their transaction type. What I'm saying in [my previous comment](https://github.com/valory-xyz/open-autonomy/pull/1734#discussion_r1067084050) is that currently we can overwrite the payload class in the metaclass registry when they have matching name and payload_attribute. This is problematic since we assume that transaction_type is unique in the application (e.g. when we check if a particular paylaod is allowed during a specific round).

Adamantios (approved):

PR #1736, opened 2023-01-11
Refactor failure recovery

Adamantios:
## Proposed changes

Refactors the recovery mechanism in order to first retrieve the Tendermint recovery parameters via the ACN before attempting to reset.

## Fixes

n/a

## Types of changes

## Checklist

## Further comments

n/a


packages/valory/skills/abstract_round_abci/behaviour_utils.py
```diff
@@ -1699,6 +1707,86 @@ def __parse_rpc_error(error: str) -> RPCResponseStatus:
...
+            dialogues = cast(TendermintDialogues, self.context.tendermint_dialogues)
+            message, _ = dialogues.create(
+                counterparty=address, performative=performative
+            )
+            message = cast(TendermintMessage, message)
+            context = EnvelopeContext(connection_id=P2P_LIBP2P_CLIENT_PUBLIC_ID)
+            self.context.outbox.put_message(message=message, context=context)
+
+        # we wait for the `address_to_acn_deliverable` to be populated with the responses (done by the tm handler)
+        yield from self.sleep(self.params.sleep_time)
```
0xArdi, line 1737: maybe we can replace this with a condition check, with a timeout?

packages/valory/skills/abstract_round_abci/handlers.py
```diff
@@ -729,9 +729,9 @@ def _response_recovery_params(
         try:
             recovery_params = json.loads(message.params)
             shared_state = cast(SharedState, self.context.state)
-            shared_state.tm_recovery_params = TendermintRecoveryParams(
-                **recovery_params
-            )
+            shared_state.address_to_acn_deliverable[
+                message.sender
+            ] = TendermintRecoveryParams(**recovery_params)
```
Adamantios, line 734: https://github.com/valory-xyz/open-autonomy/pull/1720#discussion_r1065962500

packages/valory/skills/abstract_round_abci/behaviour_utils.py
```diff
@@ -1699,6 +1707,86 @@ def __parse_rpc_error(error: str) -> RPCResponseStatus:
...
+            SharedState, self.context.state
+        ).address_to_acn_deliverable = dict.fromkeys(addresses)
+
+        for i in range(self.params.max_attempts):
+            self.context.logger.debug(
+                f"ACN attempt {i + 1}/{self.params.max_attempts}."
+            )
+            yield from self._acn_request_from_pending(performative)
+
+            result = cast(SharedState, self.context.state).get_acn_result()
```
0xArdi, line 1762: Should we set this to None after getting it once?

packages/valory/skills/abstract_round_abci/behaviour_utils.py
```diff
@@ -1971,6 +2059,14 @@ def _handle_unhealthy_tm(self) -> Generator:
...
         # since we have reached this point that means that the cause of blocks not being received
         # cannot be attributed to a lack of peers in the network
+        # therefore, we request the recovery parameters via the ACN, and if we succeed, we use them to recover
+        acn_communication_success = yield from self.request_recovery_params()
+        if not acn_communication_success:
+            self.context.logger.error(
+                "Failed to get the recovery parameters via the ACN. Cannot reset Tendermint."
+            )
+            return
+
```
0xArdi, line 2069: :+1:

packages/valory/skills/abstract_round_abci/models.py
```diff
@@ -207,6 +210,23 @@ def synchronized_data(self) -> BaseSynchronizedData:
         """Get the latest synchronized_data if available."""
         return self.round_sequence.latest_synchronized_data
 
+    def get_acn_result(self) -> Any:
+        """Get the majority of the ACN deliverables."""
+        if len(self.address_to_acn_deliverable) == 0:
+            return None
+
+        threshold = consensus_threshold(self.synchronized_data.nb_participants)
```
0xArdi, line 218: this here means (2/3) + 1 + 1. Since the agent doing the call is not sending a payload. In the future when we might want to consider the case when more than 1/3 of participants are down this will need adjustment.

DavidMinarsch: Yes suggest fixing this now. Just add: -1

packages/valory/skills/abstract_round_abci/behaviour_utils.py
Adamantios: Sure, but this would require that we replace the retry logic:

https://github.com/valory-xyz/open-autonomy/blob/0e3cf01ec9a882f401868049fd419e807f577f71/packages/valory/skills/abstract_round_abci/behaviour_utils.py#L1756-L1766

So I would suggest that we address this on a separate PR at some point.

packages/valory/skills/abstract_round_abci/behaviour_utils.py
Adamantios: Do you mean the `address_to_acn_deliverable`? This is reset every time we begin a new ACN request at the start of this method.

packages/valory/skills/abstract_round_abci/models.py
Adamantios: 435afd87e35c3a99ed13c426e2d9e99ebe651e8f

0xArdi:
Interesting e2e failure:
```
>       assert missing_agent_logs == "", missing_agent_logs  # nosec
E       AssertionError: Strings ['local height == remote height; continuing execution...'] didn't appear in agent_00001 output.
E       ''registration_startup_round' round is done with event: Event.DONE' appeared only 0 out of 1 times
```

0xArdi (approved):

PR #1767, opened 2023-01-23
Tests for the component registry contract

angrybayblade:
## Proposed changes

This PR adds tests for the component registry contract

## Fixes

## Types of changes

## Checklist


plugins/aea-test-autonomy/aea_test_autonomy/base_test_classes/contracts.py
```diff
@@ -53,6 +54,7 @@ class BaseContractTest(ABC):
     identifier: str = EthereumCrypto.identifier
     deployer_crypto: Crypto
     contract_address: Optional[str] = None
+    deploy_contract: bool = True
```
angrybayblade, line 57: Deploying the contracts should be optional, the registries image comes with all of required contracts pre deployed

packages/valory/contracts/component_registry/tests/test_contract.py
```diff
@@ -0,0 +1,116 @@
...
+            return_value=mock.MagicMock(
+                events=mock.MagicMock(
+                    CreateUnit=mock.MagicMock(
+                        createFilter=lambda **_: mock.MagicMock(
+                            get_all_entries=lambda *_: []
+                        )
+                    )
+                )
+            ),
+        ):
```
0xArdi, line 79: This is OK, but I don't think this is the same as having the actual contract, and interacting with it here. You are assuming that what this is returning here matches how you are handling it.

plugins/aea-test-autonomy/aea_test_autonomy/base_test_classes/contracts.py
0xArdi: why not set it to false by default then?

plugins/aea-test-autonomy/aea_test_autonomy/fixture_helpers.py
```diff
@@ -584,6 +584,23 @@ def _build_image(cls) -> DockerImage:
...
+    def _build_image(cls) -> DockerImage:
+        """Build the image."""
+        client = docker.from_env()
+        return RegistriesDockerImage(
+            client=client,
+            addr=cls.addr,
+            port=cls.port,
+        )
+
+
```
0xArdi, line 603: What does this accomplish that `UseRegistries` doesn't?

plugins/aea-test-autonomy/aea_test_autonomy/base_test_classes/contracts.py
angrybayblade: Because I thought setting it false by default would introduce a breaking change since other contract require deployment. If that's not the case I can change this

plugins/aea-test-autonomy/aea_test_autonomy/fixture_helpers.py
angrybayblade: I was not aware of the base class, I tried searching for base test class but couldn't find one. Now I know why, I'll update the code

plugins/aea-test-autonomy/aea_test_autonomy/base_test_classes/contracts.py
0xArdi: Okay I see what you mean, you could do something like this in the child class if you dont want to delpoy:
https://github.com/valory-xyz/autonomous-fund/blob/main/packages/balancer/agents/autonomous_fund/tests/test_contracts/test_managed_pool/test_contract.py#L74

plugins/aea-test-autonomy/aea_test_autonomy/fixture_helpers.py
angrybayblade: I tried using the `UseRegistries` but there is an issue, it's not designed for testing the contracts. The  `HardHatBaseTest ` and all the classes derived from it are compatible with the `BaseContractTest`.

0xArdi (approved):

plugins/aea-test-autonomy/aea_test_autonomy/base_test_classes/contracts.py
angrybayblade: I tried and I get error

```
function_abi = {'inputs': [{'internalType': 'string', 'name': '_name', 'type': 'string'}, {'internalType': 'string', 'name': '_symbol...nternalType': 'string', 'name': '_baseURI', 'type': 'string'}], 'stateMutability': 'nonpayable', 'type': 'constructor'}
args = (), kwargs = {}

    def merge_args_and_kwargs(
        function_abi: ABIFunction, args: Sequence[Any], kwargs: Dict[str, Any]
    ) -> Tuple[Any, ...]:
        """
        Takes a list of positional args (``args``) and a dict of keyword args
        (``kwargs``) defining values to be passed to a call to the contract function
        described by ``function_abi``.  Checks to ensure that the correct number of
        args were given, no duplicate args were given, and no unknown args were
        given.  Returns a list of argument values aligned to the order of inputs
        defined in ``function_abi``.
        """
        # Ensure the function is being applied to the correct number of args
        if len(args) + len(kwargs) != len(function_abi.get('inputs', [])):
>           raise TypeError(
                "Incorrect argument count.  Expected '{0}'.  Got '{1}'".format(
                    len(function_abi['inputs']),
                    len(args) + len(kwargs),
                )
            )
E           TypeError: Incorrect argument count.  Expected '3'.  Got '0'

../.local/share/virtualenvs/open-autonomy-VW6Otnns/lib/python3.10/site-packages/web3/_utils/abi.py:431: TypeError
```

plugins/aea-test-autonomy/aea_test_autonomy/base_test_classes/contracts.py
```diff
@@ -131,6 +135,12 @@ class BaseHardhatAMMContractTest(BaseContractTest, HardHatAMMBaseTest):
     """Base test case for testing AMM contracts on Hardhat."""
 
 
+class BaseRegistriesContractsTest(BaseContractTest, RegistriesBaseTest):
```
DavidMinarsch, line 138: This is not allowed here! Violation of separation of concerns

plugins/aea-test-autonomy/aea_test_autonomy/fixture_helpers.py
```diff
@@ -584,6 +584,23 @@ def _build_image(cls) -> DockerImage:
         return AMMNetDockerImage(client, cls.addr, cls.port)
 
 
+class RegistriesBaseTest(HardHatBaseTest):
```
DavidMinarsch, line 587: See above

DavidMinarsch (changes requested):

plugins/aea-test-autonomy/aea_test_autonomy/base_test_classes/contracts.py
angrybayblade: I don't understand, the `BaseRegistriesContractsTest` is a class for testing the contract so shouldn't it belong in the base contract test classes module?

DavidMinarsch (approved):

PR #1762, opened 2023-01-20
feat: add serialized db state on `TendermintRecoveryParams`

0xArdi:
## Proposed changes

This PR adds the serialized db to the `TendermintRecoveryParams`.

## Types of changes

What types of changes does your code introduce? (A **breaking change** is a fix or feature that would cause existing functionality and APIs to not work as expected.)
_Put an `x` in the box that applies_

- [x] Non-breaking fix (non-breaking change which fixes an issue)
- [ ] Breaking fix (breaking change which fixes an issue)
- [ ] Non-breaking feature (non-breaking change which adds functionality)
- [ ] Breaking feature (breaking change which adds functionality)
- [ ] Refactor (non-breaking change which changes implementation)
- [ ] Messy (mixture of the above - requires explanation!)

## Checklist

_Put an `x` in the boxes that apply._

- [ ] I have read the [CONTRIBUTING](../blob/main/CONTRIBUTING.md) doc
- [ ] I am making a pull request against the `main` branch (left side). Also you should start your branch off our `main`.
- [ ] Lint and unit tests pass locally with my changes
- [ ] I have added tests that prove my fix is effective or that my feature works
- [ ] I have locally run services that could be impacted and they do not present failures derived from my changes

## Further comments

If this is a relatively large or complex change, kick off the discussion by explaining why you chose the solution you did and what alternatives you considered, etc...

PR #1754, opened 2023-01-19
[0.9.0]: Make database serializable and hashable

Adamantios:
## Proposed changes

A breaking change to the database to make it serializable and hashable. This means that the data inserted into the database are now enforced to be primitive or non-primitive built-in types only, except for `sets` and `frozensets`. In essence, the data should be `json` serializable. 

Therefore, a lot of unit tests are expected to fail, because the rounds need to be refactored to not store unserializable data to the db, and the corresponding properties in the synchronized data should now convert them back to the desired type.

Making the database hashable, will allow us to use the digest for the Tendermint app hash. This is a better alternative for the app hash because it guarantees that the state is the same for all the agents at the end of each round.

## Fixes

n/a

## Types of changes

What types of changes does your code introduce? (A **breaking change** is a fix or feature that would cause existing functionality and APIs to not work as expected.)
_Put an `x` in the box that applies_

- [ ] Non-breaking fix (non-breaking change which fixes an issue)
- [ ] Breaking fix (breaking change which fixes an issue)
- [ ] Non-breaking feature (non-breaking change which adds functionality)
- [x] Breaking feature (breaking change which adds functionality)
- [ ] Refactor (non-breaking change which changes implementation)
- [ ] Messy (mixture of the above - requires explanation!)

## Checklist

_Put an `x` in the boxes that apply._

- [x] I have read the [CONTRIBUTING](../blob/main/CONTRIBUTING.md) doc
- [x] I am making a pull request against the `main` branch (left side). Also you should start your branch off our `main`.
- [x] Lint and unit tests pass locally with my changes
- [x] I have added tests that prove my fix is effective or that my feature works
- [x] I have locally run services that could be impacted and they do not present failures derived from my changes

## Further comments

n/a

PR #1775, opened 2023-01-26
Use db for the app hash

Adamantios:
## Proposed changes

Use the db's hash introduced in #1754 for the app hash.

This is expected to cause a lot of e2e test failures and make the apps unusable. The agents are not able to join the network at the end of the registration startup when using the new app hash. This is because they all have different app hashes. The reason for that is the way that we have developed our registration startup's logic. The agents are updating the synchronized data *from the behaviour's side*! The fact that it fails when we use the new app hash is, of course, a good thing, because we should never update the synced data anywhere except a round, because we may end up with different states, and the new logic guarantees that.

A new PR will be opened on top to fix the registration startup.

Moreover, the automated recovery logic will also start failing, because the `TendermintRecoveryParams` needs to be updated to include the serialized db's state. This is done on #1762.

## Fixes

n/a

## Types of changes

## Checklist

## Further comments

n/a


Adamantios:
0e5305d66a9da022658e1aca3e2c6b4a1e5ebe0e covers gap on #1776 and 0af9381c99ced56f625fcdaf9d9f499fcb96c959 covers gap on #1762.

0xArdi (approved):

DavidMinarsch (approved):

PR #1791, opened 2023-02-01
`autonomy analyse service` for checking service readiness

angrybayblade:
## Proposed changes

This PR adds a command which performs following checks on a service 

- Checks for on-chain state using the token ID if provided (using `--token-id` flag)
- Check required overrides for an ABCI skill
- Checks if the agent package is published on the IPFS registry
- Cross verifies component overrides between agent and service configs to make sure service does not define any extra overrides
- Checks if the agent dependencies are published on the IPFS registry

## Fixes

## Types of changes

## Checklist


Adamantios (approved):
Looks good, though it would be nice if we could also verify the env vars. For example, a typo in the agents' env vars might end up in a service override not working.

Do I understand correctly that `verify_overrides` does not check for this? If yes, can we add this on a separate PR?

DavidMinarsch (approved):

PR #1802, opened 2023-02-06
[0.9.0]: Fix `ApiSpecs`

Adamantios:
## Proposed changes

The type for the `headers` and the `parameters` was defined as `List[Tuple[str, str]]` in the `ApiSpecs`. However, a `tuple` will never occur when loading from a `yaml`. Moreover, conceptually it does not make sense to use a `list` for these configurations. Therefore, they were both converted to a mapping, which is more intuitive.

## Fixes

n/a

## Types of changes

## Checklist

## Further comments

n/a


packages/valory/skills/abstract_round_abci/tests/test_models.py
```diff
@@ -165,8 +166,8 @@ def test_get_spec(
         actual_specs = {
             "url": "http://dummy",
             "method": "GET",
-            "headers": [("Dummy-Header", "dummy_value")],
-            "parameters": [("Dummy-Param", "dummy_param")],
+            "headers": {"Dummy-Header": "dummy_value"},
+            "parameters": {"Dummy-Param": "dummy_param"},
```
DavidMinarsch, line 170: @angrybayblade originally introduced the list type I believe. Please check if you're happy

packages/valory/skills/abstract_round_abci/models.py
```diff
@@ -541,11 +542,11 @@ def __init__(self, *args: Any, **kwargs: Any) -> None:
         self.url: str = self._ensure("url", kwargs, str)
         self.api_id: str = self._ensure("api_id", kwargs, str)
         self.method: str = self._ensure("method", kwargs, str)
-        self.headers: List[Tuple[str, str]] = self._ensure(
-            "headers", kwargs, List[Tuple[str, str]]
+        self.headers: Dict[str, str] = dict(
```
DavidMinarsch, line 545: why not OrderedDict twice?

Adamantios: Well, I thought it would not make much sense to require an `OrderedDict` for all the HTTP requests with `python >=3.7`.

0xArdi (approved):

DavidMinarsch (approved):

Issue #1655, opened 2022-12-13
Remove try/except block and excess logs from valory/connection/abci


0xArdi:
There are multiple `info` logs in the `abci` connection. They are **temporarily** put there to see how the connection behaves while running for longer periods of time.
This issue will be mentioned as `TODO` close to the logs to be removed.

PR #1804, opened 2023-02-06
fix/misc catchup problems

0xArdi:
## Fixes

This PR fixes various issues found while testing rejoining of agents across multiple periods (reset indexes).
The issues that it fixes:
- Restarts tendermint whenever there's a connection drop, from the abci or otherwise.
- Makes sure monitoring is performed even when tendermint is reset by the tendermint server. Previous to this PR, we wouldn't see from tendermint if it had been reset due to a failure, like the ones noted above.
- Clean up timeouts and `last_timestamp` before trying to restore state received from peers.
- Don't wait for the observation interval in cases when reset is performed for recovery. This was the case before as well, now it's more explicit, and avoids running into `last_timestamp is None` errors when recovering.
- Enables ACN communication for rejoining agents, by accepting requests/responses from `all_participants`. 
- Fixes https://github.com/valory-xyz/open-autonomy/issues/1655.

## Types of changes

## Checklist


packages/valory/skills/abstract_round_abci/models.py
```diff
@@ -532,6 +541,15 @@ class TendermintRecoveryParams(TypeCheckMixin):
     reset_params: Optional[List[Tuple[str, str]]] = None
     serialized_db_state: Optional[str] = None
 
+    def __hash__(self) -> int:
+        """Hash the object."""
+        return hash(
+            self.reset_from_round
+            + str(self.round_count)
+            + str(self.serialized_db_state)
+            + json.dumps(self.reset_params)
```
DavidMinarsch, line 550: Sort keys?

Adamantios: Not necessary for determinism since we are using python >= 3.7

0xArdi: This is a list of tuples actually, so that shouldnt make a difference.

DavidMinarsch: It's recursive afaik and we should always use it imo. Technically, there could be a non-Python client interacting. This just makes it all more transparent.

0xArdi: After merging `fix/api-specs` the `reset_params` is actually `Dict[str, str]`, `sort_keys=True` was added.

packages/valory/agents/register_reset/tests/test_register_reset.py
```diff
@@ -56,7 +56,6 @@
 
 
 HAPPY_PATH = (
-    RoundChecks(RegistrationStartupRound.auto_round_id()),
```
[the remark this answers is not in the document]
0xArdi: Yes, the rejoining agent will not go through this.

DavidMinarsch (approved):

PR #1805, opened 2023-02-07
Remove `consensus` from the configuration

Adamantios:
## Proposed changes

Removes consensus from the skills' configurations and replaces the `max participants` with the length of the `all_participants` list.

The deployments also need to be refactored (by @angrybayblade ) and therefore this PR is marked as a draft until another PR is merged on this one, containing the necessary changes for that.

## Fixes

n/a

## Types of changes

## Checklist

## Further comments

n/a


packages/valory/skills/abstract_round_abci/base.py
```diff
@@ -423,16 +423,6 @@ def consensus_threshold(self) -> int:
         """Get the consensus threshold."""
```
DavidMinarsch, line 423: Actually, I realise that the on-chain protocol let's you define the consensus_threshold as >= this value here. So we should also fetch this via the CLI from the on-chain protocol to ensure it matches...

packages/valory/skills/abstract_round_abci/models.py
```diff
@@ -409,7 +424,8 @@ def __init__(
     def setup(self) -> None:
         """Set up the model."""
         self._round_sequence = RoundSequence(self.abci_app_cls)
-        consensus_params = cast(BaseParams, self.context.params).consensus_params
+        # consensus parameters will not be available if the current skill is abstract
+        consensus_params = getattr(self.context.params, "consensus_params", None)
```
DavidMinarsch, line 428: I thought we remove them?

packages/valory/skills/abstract_round_abci/base.py
Adamantios: So revert the PR, make the CLI also fetch the `max_aprticipants` and populate the config for that, and add a check for `max_participants == len(all_participants)`?

packages/valory/skills/abstract_round_abci/models.py
Adamantios: We do not remove the consensus params from the model, but remove them from the configuration, because we can infer the `max_participants` from `all_participants` and create the consensus params in the model using the latter.

packages/valory/skills/abstract_round_abci/base.py
DavidMinarsch: No, I mean that we should also fetch the threshold value via config and pass in.

packages/valory/skills/abstract_round_abci/models.py
DavidMinarsch: I know we can infer `max_participants`. But `threshold` should actually match what's defined onchain. So we will need to pass it in via config.

Adamantios: Ok, I see what you mean. So we do not even need the `ConsensusParams` class. The `consensus_threshold` property will become a configured parameter (`setup` param would make the most sense I believe), which will be overridden by the CLI, and retrieved from the on-chain protocol.

Adamantios:
@angrybayblade There are 2 deployment related changes required to be implemented on top:

1. We need to remove the `max_participants` from the deployments.
2. The `from_token` deployment needs to replace the `consensus_threshold` with the value provided by the on-chain protocol - if provided - in a similar way that it does with `all_participants`.

0xArdi (approved):

0xArdi (approved):

packages/valory/skills/abstract_round_abci/models.py
```diff
@@ -305,44 +301,47 @@ def __init__(self, *args: Any, **kwargs: Any) -> None:
...
+        self.setup_params: Dict[str, Any] = self._ensure("setup", kwargs, dict)
 
         # we sanitize for null values as these are just kept for schema definitions
         skill_id = kwargs["skill_context"].skill_id
-        setup_params: Dict[str, List[Any]] = {
-            key: check_val(val, skill_id)
-            for key, val in setup_params_.items()
-            if val is not None
-        }
-        self.setup_params = setup_params
```
DavidMinarsch, line 325: Why this change?

packages/valory/skills/abstract_round_abci/models.py
```diff
@@ -443,10 +441,12 @@ def setup(self) -> None:
                     cross_period_persisted_keys=self.abci_app_cls.cross_period_persisted_keys,
                 )
             ),
-            consensus_params,
             self.context.logger,
         )
-        self.initial_tm_configs = dict.fromkeys(self.synchronized_data.all_participants)
+        if not self.context.is_abstract_component:
```
DavidMinarsch, line 446: Why this change?

packages/valory/skills/reset_pause_abci/rounds.py
```diff
@@ -54,7 +54,11 @@ def end_block(self) -> Optional[Tuple[BaseSynchronizedData, Event]]:
         """Process the end of the block."""
         if self.threshold_reached:
             extra_kwargs = {}
-            for key in self.synchronized_data.db.cross_period_persisted_keys:
+            cross_period_persisted_keys = (
+                self.synchronized_data.db.cross_period_persisted_keys
+            )
+            cross_period_persisted_keys.update({"consensus_threshold"})
```
DavidMinarsch, line 60: Why? Isn't the default set of `cross_period_persisted_keys` hard coded elsewhere and should be updated there?

packages/valory/skills/reset_pause_abci/skill.yaml
```diff
@@ -110,11 +108,7 @@ models:
       round_timeout_seconds: 30.0
       service_id: reset_pause_abci
       service_registry_address: null
-      setup:
-        all_participants:
-        - - '0x0000000000000000000000000000000000000000'
-        safe_contract_address:
-        - '0x0000000000000000000000000000000000000000'
```
DavidMinarsch, line 117: Why removed?

packages/valory/skills/termination_abci/skill.yaml
```diff
@@ -120,11 +118,7 @@ models:
       round_timeout_seconds: 30.0
       service_id: termination_abci
       service_registry_address: '0x48b6af7B12C71f09e2fC8aF4855De4Ff54e775cA'
-      setup:
```
DavidMinarsch, line 123: Same, don't understand why we remove it in some

packages/valory/skills/transaction_settlement_abci/skill.yaml
```diff
@@ -126,9 +124,7 @@ models:
       round_timeout_seconds: 30.0
       service_id: registration
       service_registry_address: null
-      setup:
-        all_participants:
-        - - '0x0000000000000000000000000000000000000000'
+      setup: {}
```
DavidMinarsch, line 127: see above

packages/valory/skills/abstract_round_abci/models.py
Adamantios: Added type check for the setup params below, so this check was duplicate and I removed it.

packages/valory/skills/abstract_round_abci/models.py
Adamantios: Abstract components do not need to specify `all_participants` or set the `initial_tm_configs`.

Therefore, in this PR when I had to add the new `consensus_params` in the setup of the skills' configurations, I decided to remove the setup from the abstract skills instead of adding the new one, since they were all unnecessary.

packages/valory/skills/reset_pause_abci/skill.yaml
Adamantios: https://github.com/valory-xyz/open-autonomy/pull/1805/files/9e28626f640e4e99f16f687ee97deeedabb27bc6#r1105375171

packages/valory/skills/termination_abci/skill.yaml
Adamantios: https://github.com/valory-xyz/open-autonomy/pull/1805#discussion_r1105375171

packages/valory/skills/transaction_settlement_abci/skill.yaml
Adamantios: https://github.com/valory-xyz/open-autonomy/pull/1805#discussion_r1105375171

packages/valory/skills/reset_pause_abci/rounds.py
Adamantios: Yes, this is set here:

https://github.com/valory-xyz/open-autonomy/blob/ee71f2f577c7baa906508e4003c0b9090a3168b7/packages/valory/skills/abstract_round_abci/base.py#L2065

However, each abci app can overwrite this, e.g., in [price oracle deployment](https://github.com/valory-xyz/price-oracle/blob/f42f53add155a74e89d1bd1e9d47c3e9456daa4c/packages/valory/skills/oracle_deployment_abci/rounds.py#L225-L228).

Therefore I performed this weird change. Also, a reason I opened [[Refactor]: db creation in `ResetAndPauseRound`](https://trello.com/c/nBsvTlEI).

PR #1807, opened 2023-02-08
Add support for validation on basic overrides required for a service deployment

angrybayblade:
## Proposed changes

This PR 

- Separates validation process for on-chain services and services which are not published yet
- Added validation for model params which are required (parsed by _ensure method), look at this [comment](https://github.com/valory-xyz/open-autonomy/pull/1807#discussion_r1101123897)
- Adds cross verification of parameters between service<->agent<->skill
- Updates validation process for the ledger connection to make it flexible for third party ledger plugins

## Fixes

## Types of changes

## Checklist


autonomy/analyse/service.py
```diff
@@ -36,15 +43,101 @@
...
+                                }
+                            },
+                            "required": [
+                                "setup",
+                                "tendermint_url",
+                                "tendermint_com_url",
+                                "service_registry_address",
+                                "share_tm_config_on_startup",
+                                "on_chain_service_id",
+                            ],
```
DavidMinarsch, line 122: This is incomplete afaik. Check what we have in scaffolding too. Also the two should be forced to stay in sync

autonomy/analyse/service.py
```diff
@@ -36,15 +43,101 @@
...
+                        },
+                    },
+                    "required": ["ethereum"],
+                }
+            },
+            "required": ["ledger_apis"],
+        },
+    },
+    "required": ["config"],
+}
```
DavidMinarsch, line 93: This is not generic. Very much scoped to ethereum version of ledger connection..

autonomy/analyse/service.py
```diff
@@ -120,42 +213,81 @@ def verify_overrides(self, agent_config: AgentConfig) -> None:
...
+        ):
+            cls._validate_override(
+                validator=ABCI_CONNECTION_VALIDATOR,
+                overrides=override,
+                has_multiple_overrides=has_multiple_overrides,
+                error_message="ABCI connection validation failed; {error}",
+            )
+        if (
+            component_id.component_type == ComponentType.CONNECTION
+            and component_id.name == "ledger"
```
DavidMinarsch, line 259: magic

autonomy/analyse/service.py
```diff
@@ -120,42 +213,81 @@ def verify_overrides(self, agent_config: AgentConfig) -> None:
...
+    ) -> None:
+        """Validate override"""
+
+        if component_id.component_type == ComponentType.SKILL:
+            cls._validate_override(
+                validator=ABCI_SKILL_VALIDATOR,
+                overrides=override,
+                has_multiple_overrides=has_multiple_overrides,
+                error_message="ABCI skill validation failed; {error}",
+            )
```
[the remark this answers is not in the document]
angrybayblade: Addressed https://github.com/valory-xyz/open-autonomy/pull/1807/commits/477b3cc4bd7edc1563faf009b198ac2cd5475f1f#diff-739012fc1855ea8228dad566ed5f58aaf94767441f8391c8962368907ca155b2R239-R242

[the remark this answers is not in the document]
angrybayblade: But we'll have to make sure every ABCI skill does contain the `abci` in the name and every non-abci skill does contain it

autonomy/analyse/service.py
angrybayblade: Addressed https://github.com/valory-xyz/open-autonomy/pull/1807/commits/477b3cc4bd7edc1563faf009b198ac2cd5475f1f#diff-739012fc1855ea8228dad566ed5f58aaf94767441f8391c8962368907ca155b2R70

autonomy/analyse/service.py
angrybayblade: We don't want to redefine every single parameter in the service configuration, we need to make sure it contains the basic overrides needed for a deployment not every single override that is defined in the skill config

autonomy/analyse/service.py
```diff
@@ -36,15 +43,100 @@
...
+LEDGER_CONNECTION_SCHEMA = {
+    "type": "object",
+    "properties": {
+        "config": {
+            "type": "object",
+            "properties": {
+                "ledger_apis": {
+                    "type": "object",
+                    "patternProperties": {
+                        "^(ethereum|fetchai|cosmos)$": {
```
DavidMinarsch, line 70: That change makes it no better. Firstly, now we restrict to these 3. Secondly, these three have not the same params defined.

Instead, revert to ethereum only validation. But rather than fail, print a warning if non-ethereum configs are used.

autonomy/analyse/service.py
DavidMinarsch: All the overrides defined for the scaffold are needed in each skill. O/w they cannot be over-riden from the agent/service & also the skill instantiation will fail if the concrete skill does not list them. Speak to @Adamantios who can point out all the "enforce" that will raise if various params are not defined.

autonomy/analyse/service.py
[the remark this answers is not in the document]
DavidMinarsch: We can detect whether a skill is `abci` type by whether or not it contains certain classes. Let's do that rather than use a convention.

DavidMinarsch (changes requested):

autonomy/analyse/service.py
angrybayblade: Addressed [771ce53](https://github.com/valory-xyz/open-autonomy/pull/1807/commits/771ce53807f351bca782686a1fa2d37cd4948d2e)

autonomy/analyse/service.py
```diff
@@ -21,29 +21,168 @@
...
+        "request_retry_delay",
+        "tx_timeout",
+        "max_attempts",
+        "service_registry_address",
+        "on_chain_service_id",
+        "share_tm_config_on_startup",
+        "tendermint_p2p_url",
+        "setup",
+    ],
+}
```
angrybayblade, line 175: @DavidMinarsch make sure we have every required model parameters in the `skill.yaml`

autonomy/cli/helpers/analyse.py
```diff
@@ -88,59 +94,199 @@ def run_dialogues_check(
...
 
-    with tempfile.TemporaryDirectory() as agent_dir:
-        agent_config_content = ipfs_tool.client.cat(
-            f"{agent.hash}/{agent.name}/{DEFAULT_AEA_CONFIG_FILE}"
+
+def _get_chained_abci_skill(
+    agent_config: AgentConfig,
+    package_manager: PackageManagerV1,
+    is_on_chain_check: bool = False,
+) -> Optional[SkillConfig]:
```
angrybayblade, line 140: @DavidMinarsch @Adamantios Method for finding the chained ABCI app skill for an agent

autonomy/cli/helpers/analyse.py
```diff
@@ -88,59 +94,199 @@ def run_dialogues_check(
...
+        ),
+    )
 
-        click.echo("Checking if the agent package is published")
-        service_analyser.check_agent_package_published(ipfs_pins=ipfs_pins)
+    skill_config = _get_chained_abci_skill(
+        agent_config=agent_config,
+        package_manager=package_manager,
+        is_on_chain_check=is_on_chain_check,
+    )
```
angrybayblade, line 264: If token ID is provided we check the components which are published on the IPFS registry else we validate the local components

autonomy/analyse/service.py
```diff
@@ -120,42 +271,112 @@ def verify_overrides(self, agent_config: AgentConfig) -> None:
...
 
-        for param in REQUIRED_PARAM_VALUES:
-            if param not in override["models"]["params"]["args"]:
-                raise ServiceValidationFailed(
-                    f"`{param}` needs to be defined in the `models:params:args` parameter"
+                (unknown_ledger,) = cast(
+                    re.Match, UNKNOWN_LEDGER_RE.match(msg)
+                ).groups()
+                logging.warning(
+                    f"Unknown ledger configuration found with name `{unknown_ledger}`"
```
angrybayblade, line 339: If the defined ledger is other than `ethereum` we print warning

autonomy/analyse/service.py
```diff
@@ -119,43 +328,175 @@ def verify_overrides(self, agent_config: AgentConfig) -> None:
...
+                has_multiple_overrides,
+            ) = Service.process_metadata(copy.deepcopy(override))
+            if skill_config.component_id == component_id:
+                if not has_multiple_overrides:
+                    self._check_overrides_recursively(
+                        check_from=skill_override_from_service,
+                        check_with=skill_override_from_agent,
+                        check_from_name=str(self.service_config.package_id),
+                        check_with_name=str(agent_config.package_id),
+                    )
```
angrybayblade, line 353: Verify overrides between the agent and the service

autonomy/analyse/service.py
```diff
@@ -119,43 +328,175 @@ def verify_overrides(self, agent_config: AgentConfig) -> None:
...
+                    self.logger.info(
+                        "Cross verifying overrides between skill and service"
+                    )
+                    self._check_overrides_recursively(
+                        check_from=skill_override_from_service,
+                        check_with=skill_config_to_check,
+                        check_from_name=str(self.service_config.package_id),
+                        check_with_name=str(skill_config.package_id),
+                    )
```
angrybayblade, line 362: Verify overrides defined on the service component and the actual skill config

autonomy/analyse/service.py
```diff
@@ -119,43 +328,175 @@ def verify_overrides(self, agent_config: AgentConfig) -> None:
...
+
+        self.logger.info("Cross verifying overrides between skill and agent")
+        self._check_overrides_recursively(
+            check_from=skill_override_from_agent,
+            check_with=skill_config_to_check,
+            check_from_name=str(agent_config.package_id),
+            check_with_name=str(skill_config.package_id),
+        )
```
angrybayblade, line 391: Verify overrides defined on the agent config and the actual skill

autonomy/analyse/service.py
```diff
@@ -119,43 +328,175 @@ def verify_overrides(self, agent_config: AgentConfig) -> None:
...
-                    f"`{setup_param}` needs to be defined in the `models:params:args:setup` parameter"
-                )
+        self.logger.info("Validating skill overrides")
+        model_params = skill_config.models.read("params")
+        self._validate_override(
+            validator=ABCI_SKILL_MODEL_PARAMS_VALIDATOR,
+            overrides=model_params.args,
+            has_multiple_overrides=False,
+            error_message="ABCI skill model parameter validation failed; {error}",
+        )
```
angrybayblade, line 472: Validate `skill.models.params.args`

Adamantios (approved):
I played around with the tool at the price-oracle. It works nicely. Just one observation:
The report seems too verbose. On a next PR I suggest that we create two levels of verbosity:
1. for top to bottom
2. for both ways

So if `1` is the verbosity level then the warnings are not reported for example if a skill has something defined but an agent does not override it. Moreover, `1` makes more sense as a default because:
- there will almost always be more configurations as we go down the hierarchy as a dev would rarely need to override everything.
- it is more important to spot missing configurations which are attempted to be overridden.

Another way to do this without using different verbosity levels would be by sorting them by importance (first `1`, then `2`), and/or special formatting, e.g., colors/prefixes, to the reported information.

DavidMinarsch (approved):

Issue #1766, opened 2023-01-21
Hardcoded runtime image registry means third parties are unable to push image.


8ball030:
**Is your feature request related to a problem? Please describe.**
all run time images assume that the author is valory.

This means that unless you have access to the valory docker repository, you are unable to make use of any deployment option other than building locally and running docker-compose.

**Describe the solution you'd like**
it would be great if there was either;

A) a flag that could be passed to autonomy build-image & autonomy deploy build

C) a way that i could configure the aea cli tool so that i could initialise with a docker registry.

i.e. aea init --docker-user-name 8ball030 

**Describe alternatives you've considered**

**Additional context**

angrybayblade:
Addressed https://github.com/valory-xyz/open-autonomy/pull/1818

PR #1818, opened 2023-02-13
[WIP] image_author option for cli image-build and deploy

solarw:
## Proposed changes

 image_autho option for cli image-build and deply

## Fixes

https://github.com/valory-xyz/open-autonomy/issues/1766

## Types of changes

## Checklist

## Further comments

### DIFF
--- a/autonomy/constants.py
+++ b/autonomy/constants.py
@@ -56,6 +56,6 @@ DEFAULT_SERVICE_REGISTRY_CONTRACTS_IMAGE = (
     f"{SERVICE_REGISTRY_IMAGE_NAME}:{SERVICE_REGISTRY_IMAGE_VERSION}"
 )
 ACN_IMAGE_NAME = os.environ.get("ACN_IMAGE_NAME", "valory/open-acn-node")
-
-OAR_IMAGE = "valory/oar-{agent}:{version}"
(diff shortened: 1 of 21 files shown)


DavidMinarsch (changes requested):

autonomy/constants.py
```diff
@@ -56,6 +56,6 @@
     f"{SERVICE_REGISTRY_IMAGE_NAME}:{SERVICE_REGISTRY_IMAGE_VERSION}"
 )
 ACN_IMAGE_NAME = os.environ.get("ACN_IMAGE_NAME", "valory/open-acn-node")
-
-OAR_IMAGE = "valory/oar-{agent}:{version}"
+DEFAULT_DOCKER_IMAGE_AUTHOR = "valory"
```
angrybayblade, line 59: I think the default author should be set using `aea.cli.utils.config.get_default_author_from_cli_config`

solarw: @angrybayblade 
dont think it's a good idea to perform on module load operations with config reading.
it will be done every time module loaded even no need in this variable.

also not sure image author and agent author are the same.

@8ball030  any ideas?

angrybayblade: I agree, but I think it'll be really confusing if we have multiple types of authors. I propose we don't use the `DEFAULT_DOCKER_IMAGE_AUTHOR` variable for default author, we pass down `None` if the author name is not provided on the command and use `get_default_author_from_cli_config` method where the author name is required

solarw: @angrybayblade updated

angrybayblade (approved):
LGTM

DavidMinarsch (approved):

PR #1827, opened 2023-02-15
[0.10.0]: `AbciAppDB`'s `create` refactoring

Adamantios:
## Proposed changes

This is a breaking refactor.

With this PR, the `AbciAppDB`'s `create()` is responsible for setting the cross-period keys for the new period and converting the corresponding data to the correct format. The skills using the `create()` method now do not need to manually set the data for the next period as this is handled automatically via the cross-period keys.

## Fixes

n/a

## Types of changes

- [x] Breaking Refactor (breaking change which changes implementation)

## Checklist

## Further comments

n/a


packages/valory/skills/abstract_round_abci/base.py
```diff
@@ -491,16 +502,27 @@ def __init__(
...
-            cross_period_persisted_keys.copy()
-            if cross_period_persisted_keys is not None
-            else set()
-        )
         self._data: Dict[int, Dict[str, List[Any]]] = {
             RESET_COUNT_START: self.setup_data  # the key represents the reset index
         }
         self._round_count = ROUND_COUNT_DEFAULT  # ensures first round is indexed at 0!
 
+        self._cross_period_persisted_keys = self.default_cross_period_keys.union(
```
DavidMinarsch, line 510: the resulting will be set not frozenset afaik

packages/valory/skills/abstract_round_abci/base.py
```diff
@@ -491,16 +502,27 @@ def __init__(
...
+
+    def _cross_period_check(self) -> None:
+        """Check the cross period keys against the setup data."""
+        not_in_cross_period = set(self._setup_data).difference(
+            self.cross_period_persisted_keys
+        )
+        if not_in_cross_period:
+            _logger.warning(
+                f"The setup data ({self._setup_data.keys()}) contain keys that are not in the "
+                f"cross period persisted keys ({self.cross_period_persisted_keys}): {not_in_cross_period}"
```
DavidMinarsch, line 523: Hm noisy?

packages/valory/skills/abstract_round_abci/base.py
```diff
@@ -573,6 +595,11 @@ def validate(data: Any) -> None:
                 "You may use `AbciAppDB.validate(your_data)` to validate your data for the `AbciAppDB`."
             )
 
+    @staticmethod
+    def deterministic(value: Iterable) -> Tuple:
+        """Converts an iterable to a sorted tuple."""
+        return tuple(sorted(value))
```
DavidMinarsch, line 601: I think this should not have a method. It's totally unrelated to the concerns of the object. If anything have a standalone method, but I think might be cleaner to do it in place.

Adamantios: Yes, I agree. The only reason I added it here is that I thought it might become relevant in the future when we address [[Feature] Ensure order of elements before storing in db](https://trello.com/c/kyijw5h9). In the future, we could extend this method to support more types and also nested data. 

I'll remove it for now since it is irrelevant.

packages/valory/skills/abstract_round_abci/base.py
Adamantios: We currently have all the setup data in the default cross-period keys, so this should not warn. In the future, we might extend the `setup` data and forget to update the default cross-period keys to contain the new item(s). At this point, it would start warning and the app would also fail if someone tried to use the new item(s) in a period > 0. This warning might be useful to troubleshoot faster :thinking: 

What do you think?

packages/valory/skills/abstract_round_abci/base.py
Adamantios:
```
assert isinstance(
    frozenset().union({"test"}), frozenset
), "`frozenset().union` should produce a `frozenset`."
```

packages/valory/skills/abstract_round_abci/base.py
Adamantios: Done on b813831dd314af80bcc981cc0996dcec4bfa7889.

0xArdi (dismissed):

0xArdi (approved):

0xArdi (approved):

dvilelaf (approved):

PR #1848, opened 2023-02-23
[0.10.0]: Set the requests equal to the limits for the resources

Adamantios:
## Proposed changes

This will ensure that an agent will never try to use more resources than expected. By setting `requests=limits`, then assuming all the agents in the pod are using the max allowed resources at some point, we still do not utilize more than the pod has available.

## Fixes

n/a

## Types of changes

## Checklist

## Further comments

n/a

### DIFF
--- a/autonomy/deploy/generators/kubernetes/templates.py
+++ b/autonomy/deploy/generators/kubernetes/templates.py
@@ -204,11 +204,11 @@ spec:
         imagePullPolicy: Always
         resources:
           limits:
-            memory: "1512Mi"
-            cpu: "1"
+            memory: "350Mi"
+            cpu: "0.05"
           requests:
-            cpu: "1"
-            memory: "1512Mi"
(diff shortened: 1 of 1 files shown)


8ball030:
I think we can get away with much much less resource usage on the tm side. 

Lets go for;


image


the cpu for the node as is, just both aligned to the limits.

For the memory for the node;

image


I think 350M for limits and requests 


# The aea

## Memory
### limits and requests at 1GB please.
image


For the CPU;


image


Please limit it to 0.5


Above this would be quite excessive and if required we should look to somehow provide that as a deployment option.

DavidMinarsch:
For the purpose of this PR hardcoding it is fine. But @Adamantios please open a Trello ticket for Icebox to allow the overriding of these values with custom ones.

8ball030 (approved):

angrybayblade (approved):

PR #1850, opened 2023-02-24
Fix/doc code tests

dvilelaf:
## Proposed changes

The test to detect outdated code block in our docs had two issues that prevented us from parsing all code blocks. This PR makes the following changes:
* Removes mistune, as it is not compatible with admonitions and therefore prevented a correct parsing
* Implements custom code block parsing via regex
* Fixes the test adding all the files and code blocks that were not being properly checked

## Fixes

## Types of changes

## Checklist

## Further comments

n/a


docs/demos/price_oracle_fsms.md
```diff
@@ -28,7 +28,6 @@ The summary of the constituent FSMs is as follows:
 | FSM                     | States  | Start states  | Final states  | Events  | Non-trivial transitions (*)   |
 |-----------------------  |-------: |-------------: |-------------: |-------: |------------------------:  |
 | AgentRegistration       |      4  |            2  |            2  |      3  |                       3   |
-| SafeDeployment          |      5  |            1  |            1  |      8  |                      14   |
```
dvilelaf, line 31: The safe deployment was removed from the oracle

scripts/check_doc_links.py
```diff
@@ -49,6 +49,8 @@
 HTTP_SKIPS = [
     "http://www.fipa.org/repository/ips.php3",
     "http://host.docker.internal:8545",
+    "http://host.docker.internal:8545",  # internal (ERR_NAME_NOT_RESOLVED)
+    "http://host.docker.internal:8545}",  # regex issue
```
jmoreira-valory, line 53: Why repeated 3 times?

dvilelaf: There are two internal docker links in our docs: the ones set as environment variable (written inside brackets) and the "normal ones". There's two problems with those 
* They are HTTP links
* The regex gets the } as part of the link in the ones inside brackets
So I needed to add both of them to both HTTP and HTTPS skips.

tests/test_docs/helper.py
```diff
@@ -50,36 +46,23 @@ class CodeType(Enum):
...
-
-
-def type_filter(type_: Optional[str], b: Dict) -> bool:
-    """
-    Check Mistune code block is of a certain type.
-
-    If the field "info" is None, return False.
-    If type_ is None, this function always return true.
+def dedent_code(code: str, n_spaces: int) -> str:
+    """Removes n spaces at the beggining of each line"""
```
jmoreira-valory, line 50: begining

tests/test_docs/helper.py
```diff
@@ -50,36 +46,23 @@ class CodeType(Enum):
...
+    lines = [line[n_spaces:] for line in code.split("\n")]
+    return "\n".join(lines)
```
jmoreira-valory, line 52: To be more precise, the code removes the firs 'n' characters of each line (not necessarily spaces)

jmoreira-valory (approved):

scripts/check_doc_links.py
angrybayblade: The URL on lines 51 and 52 are the same

scripts/check_doc_links.py
```diff
@@ -41,14 +41,14 @@
 
 
 MAX_WORKERS = 10
-URL_REGEX = r'(https?:\/\/(?:www\.|(?!www))[a-zA-Z0-9][a-zA-Z0-9-]+[a-zA-Z0-9]\.[^\s)"]{2,}|www\.[a-zA-Z0-9][a-zA-Z0-9-]+[a-zA-Z0-9]\.[^\s)"]{2,}|https?:\/\/(?:www\.|(?!www))[a-zA-Z0-9]+\.[^\s)"]{2,}|www\.[a-zA-Z0-9]+\.[^\s)"]{2,})'
+URL_REGEX = r'(https?:\/\/(?:www\.|(?!www))[a-zA-Z0-9][a-zA-Z0-9-]+[a-zA-Z0-9]\.[^\s})"]{2,}|www\.[a-zA-Z0-9][a-zA-Z0-9-]+[a-zA-Z0-9]\.[^\s})"]{2,}|https?:\/\/(?:www\.|(?!www))[a-zA-Z0-9]+\.[^\s)"}]{2,}|www\.[a-zA-Z0-9]+\.[^\s)"}]{2,})'
```
dvilelaf, line 44: I have now fixed this regex so it does not get the invalid char }

scripts/check_doc_links.py
dvilelaf: Fixed

jmoreira-valory (approved):
LGTM.
Hope we can integrate the mapping and skipped files in the .md as "magic comments" in a future improvement.

Adamantios (approved):

PR #1852, opened 2023-03-01
Deprecate the usage of `--force` flag for overwriting the existing builds

### LINKED ISSUE #1851: This makes quite the assumtion that the only way to deploy from token id is into docker-compose
https://github.com/valory-xyz/open-autonomy/blob/496d7185589845a2125bba39a708db6c20961cbe/autonomy/cli/helpers/deployment.py#L266

For now, i am hacking my way around this to use kube but this is none optiomal.

It would be great if either i could;

A) pass --kubernetes
B) pass --deploy False

angrybayblade:
## Proposed changes

This PR

- Removes the support for `--force` flag on the `autonomy deploy build` command
- Updates the test to remove the usage of `--force` flag
- Updates the documentation for the same

**Breaking change**

- `autonomy deploy build` does not support the usage of `--force` flag to remove the existing build directories anymore

## Fixes

#1851 

## Types of changes

## Checklist


autonomy/cli/helpers/deployment.py
```diff
@@ -144,10 +142,9 @@ def build_deployment(  # pylint: disable=too-many-arguments, too-many-locals
     image_author: Optional[str] = None,
 ) -> None:
     """Build deployment."""
+
     if build_dir.is_dir():  # pragma: no cover
-        if not force_overwrite:
-            raise click.ClickException(f"Build already exists @ {build_dir}")
-        shutil.rmtree(build_dir)
```
angrybayblade, line 150: I think we should just leave this to the users since this a very trivial thing to do. Keeping this piece of code here will require some platform and user group specific hacks.

angrybayblade (commented):
@8ball030 I've checked [workflow.yaml](https://github.com/valory-xyz/internal-infrastructure/blob/main/charts/agent-service/templates/workflows.yaml) We don't use this flag anywhere so it should not be a problem for you

8ball030:
Thanks for checking the workflows :)

jmoreira-valory (approved):

PR #1874, opened 2023-03-09
[WIP] This PR adds support for using hardware wallets for interacting with the on-chain contracts

angrybayblade:
## Proposed changes

Describe the big picture of your changes here to communicate to the maintainers why we should accept this pull request.

## Fixes

If it fixes a bug or resolves a feature request, be sure to link to that issue.

## Types of changes

What types of changes does your code introduce? (A **breaking change** is a fix or feature that would cause existing functionality and APIs to not work as expected.)
_Put an `x` in the box that applies_

- [ ] Non-breaking fix (non-breaking change which fixes an issue)
- [ ] Breaking fix (breaking change which fixes an issue)
- [ ] Non-breaking feature (non-breaking change which adds functionality)
- [ ] Breaking feature (breaking change which adds functionality)
- [ ] Refactor (non-breaking change which changes implementation)
- [ ] Messy (mixture of the above - requires explanation!)

## Checklist

_Put an `x` in the boxes that apply._

- [ ] I have read the [CONTRIBUTING](../blob/main/CONTRIBUTING.md) doc
- [ ] I am making a pull request against the `main` branch (left side). Also you should start your branch off our `main`.
- [ ] Lint and unit tests pass locally with my changes
- [ ] I have added tests that prove my fix is effective or that my feature works
- [ ] I have locally run services that could be impacted and they do not present failures derived from my changes

## Further comments

If this is a relatively large or complex change, kick off the discussion by explaining why you chose the solution you did and what alternatives you considered, etc...

PR #1879, opened 2023-03-13
Make the hardware wallet installation optional

angrybayblade:
## Proposed changes

This PR makes the usage of the hardware wallet optional

## Fixes

https://github.com/valory-xyz/open-autonomy/pull/1874#discussion_r1132900605

## Types of changes

## Checklist

### DIFF
--- a/autonomy/cli/helpers/chain.py
+++ b/autonomy/cli/helpers/chain.py
@@ -28,8 +28,6 @@ from aea.configurations.loader import load_configuration_object
 from aea.crypto.base import Crypto, LedgerApi
 from aea.crypto.registries import crypto_registry, ledger_apis_registry
 from aea_ledger_ethereum.ethereum import EthereumApi, EthereumCrypto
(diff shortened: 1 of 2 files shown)


autonomy/cli/helpers/chain.py
```diff
@@ -61,6 +59,15 @@
 from autonomy.configurations.base import PACKAGE_TYPE_TO_CONFIG_CLASS, Service
 
 
+try:
+    from aea_ledger_ethereum_hwi.exceptions import HWIError
```
solarw, line 63: as next steps: probly apply plugin discovery approach used for aea plugins?
implemented here `https://github.com/valory-xyz/open-aea/blob/main/aea/crypto/plugin.py`
instead of hardcode `try..import`


probably will require plugin updates, so can not be implemented adhoc.

solarw (approved):
small optional comment
LGTM

PR #1923, opened 2023-04-12
Automatic tendermint P2P url update at runtime

### LINKED ISSUE #1903: Tendermint service config should be handled by the framework
See [here](https://github.com/valory-xyz/IEKit/pull/37#discussion_r1151914545): 

> This should be handled by autonomy framework, please open an issue on open-autonomy to fix this and on this repo to rewrite the service config once the release is out with the fix

angrybayblade:
## Proposed changes

This adds support for updating the tendermint P2P url at the runtime

## Fixes

#1903 

## Types of changes

## Checklist

## Further comments


autonomy/deploy/base.py
```diff
@@ -318,6 +326,13 @@ def _update_tendermint_params(
                 param_args[TENDERMINT_URL_PARAM] = TENDERMINT_NODE.format(idx)
                 param_args[TENDERMINT_COM_URL_PARAM] = TENDERMINT_COM.format(idx)
 
+            if TENDERMINT_P2P_URL_PARAM not in param_args:
+                tm_p2p_url = os.environ.get(
+                    TENDERMINT_P2P_URL_ENV_VAR.format(idx),
+                    TENDERMINT_P2P_URL.format(idx, TENDERMINT_P2P_PORT),
+                )
+                param_args[TENDERMINT_P2P_URL_PARAM] = tm_p2p_url
```
angrybayblade, line 334: @dvilelaf @Adamantios @0xArdi This will address the issue of manually setting the `tendermint_p2p_url`, so you don't have to unnecessarily define multi overrides if not required. Things to keep in mind

1. The update won't be performed if `tendermint_p2p_url` is already defined in the abci skill overrides
2. The update method will first check if the env var is present or not for the `tendermint_p2p_url`, if there's an env var available for the override the env var will be used. The env var needs to follow `TM_P2P_NODE_URL_<AGENT_ID>` format. Eg

If you expose `TM_P2P_NODE_URL_0`, the service builder will update the `tendermint_p2p_url` parameter for agent 1 using this env var
3. If there's not env var available it'll generate the URL using `node<AGENT_ID>:<PORT>` format

angrybayblade: This update will be performed at the runtime on the service component, I don't understand the first request.

> Also, what happens if the user specifies an env var with a different format than TM_P2P_NODE_URL_<AGENT_ID> in the service configuration?

It won't get picked and the value will be set to auto generated URL in `node<ID>:<PORT>` format

angrybayblade: > tm p2p value that ends up being used

Here are some examples

```yaml
(...)
---
public_id: valory/impact_evaluator_abci:0.1.0
type: skill
0:
  models:
    params:
      args:
        (...)
        tendermint_p2p_url: localhost:8000
```

Since the `tendermint_p2p_url` is present in the override at the service level `localhost:8000` will be used

```yaml
(...)
---
public_id: valory/impact_evaluator_abci:0.1.0
type: skill
0:
  models:
    params:
      args:
        (...)
        tendermint_p2p_url: ${SOME_ENV_VAR_NAME:str:localhost:8000}
```

Since the `tendermint_p2p_url` is present in the override at the service level and represented by an env var, if the value for env var `SOME_ENV_VAR_NAME` is present it will be used or  `localhost:8000` will be used

```yaml
(...)
---
public_id: valory/impact_evaluator_abci:0.1.0
type: skill
0:
  models:
    params:
      args:
        (...)
```

Since the `tendermint_p2p_url` is not present in the override at the service level, there are two scenarios here

1. If the env var `TM_P2P_NODE_URL_0` is present the env var value will be used or the auto generated value which is `node0:26656` will be used

angrybayblade: I'm sorry looks like I got confused with the question, the statement in the second comment is correct.

dvilelaf (approved):

Adamantios (approved):

PR #1925, opened 2023-04-12
Create slashing structures

Adamantios:
## Proposed changes

Creates some structures necessary for introducing slashing.

## Fixes

n/a

## Types of changes

What types of changes does your code introduce? (A **breaking change** is a fix or feature that would cause existing functionality and APIs to not work as expected.)
_Put an `x` in the box that applies_

- [ ] Non-breaking fix (non-breaking change which fixes an issue)
- [ ] Breaking fix (breaking change which fixes an issue)
- [x] Non-breaking feature (non-breaking change which adds functionality)
- [ ] Breaking feature (breaking change which adds functionality)
- [ ] Refactor (non-breaking change which changes implementation)
- [ ] Messy (mixture of the above - requires explanation!)

## Checklist

_Put an `x` in the boxes that apply._

- [x] I have read the [CONTRIBUTING](../blob/main/CONTRIBUTING.md) doc
- [x] I am making a pull request against the `main` branch (left side). Also you should start your branch off our `main`.
- [x] Lint and unit tests pass locally with my changes
- [x] I have added tests that prove my fix is effective or that my feature works
- [x] I have locally run services that could be impacted and they do not present failures derived from my changes
- [x] Public-facing documentation has been updated with the changes affected by this PR. Even if the provided contents are not in their final form, all significant information must be included.
- [x] Any backwards-incompatible/breaking change has been clearly documented in the upgrading document.

## Further comments

n/a

PR #1929, opened 2023-04-13
Update models for slashing

Adamantios:
## Proposed changes

Initializes offense-related attributes in the round sequence and creates a method to map agent addresses to validator addresses for slashing.

## Fixes

n/a

## Types of changes

## Checklist

## Further comments

n/a


packages/valory/skills/abstract_round_abci/tests/test_base.py
```diff
@@ -1859,9 +1859,11 @@ def test_add(max_length: int, hypothesis_data: Any) -> None:
...
 
     @staticmethod
-    @pytest.mark.parametrize("max_length", (0,))
-    @pytest.mark.parametrize("num_positive", (0,))
-    @pytest.mark.parametrize("num_negative", (0,))
+    @given(
+        max_length=integers(min_value=0, max_value=30_000),
+        num_positive=integers(min_value=0),
+        num_negative=integers(min_value=0),
+    )
```
Adamantios, line 1866: Unrelated to the PR, but forgot to do this on #1925.

packages/valory/skills/abstract_round_abci/tests/test_models.py
```diff
@@ -490,14 +490,8 @@ def test_synchronized_data_db(self, *_: Any) -> None:
...
-            assert (
-                shared_state.synchronized_data.db.get_strict("safe_contract_address")
-                == "0xsafe"
-            )
-            assert (
-                shared_state.synchronized_data.db.get_strict("oracle_contract_address")
-                == "0xoracle"
-            )
+            for key, value in mock_params.setup_params.items():
+                assert shared_state.synchronized_data.db.get_strict(key) == value
```
Adamantios, line 494: Unrelated to the PR, but a check for `all_participants` was missing, so I updated the test to include it.

packages/valory/skills/abstract_round_abci/models.py
```diff
@@ -427,6 +431,58 @@ def __init__(
         kwargs["skill_context"] = skill_context
         super().__init__(*args, **kwargs)
 
+    @property
+    def validator_to_agent(self) -> Dict[str, str]:
```
dvilelaf, line 435: What is the difference between a validator and an agent?

0xArdi: Validator is the tendermint part, it is identified with a different key than the agent.
This key is used for consensus in tendermint itself. Because malicious behavior can be comitted in consensus-level as well, we need a way to punish the agent who's validator committed the act.

0xArdi (approved):

dvilelaf (approved):

PR #1976, opened 2023-05-22
Fix tm disconnection issue

Adamantios:
## Proposed changes

This pull request addresses the problem of tm nodes' disconnection by introducing a gentle reset as the initial step before proceeding with a hard reset.

Moreover, it resolves an issue that was revealed by the changes introduced on 8346638c70a89068d764e7a73a2fe2abe94433bf:
https://github.com/valory-xyz/open-autonomy/pull/1976#issuecomment-1563290036

## Fixes

n/a

## Types of changes

## Checklist

## Further comments

n/a


Adamantios:
This PR revealed an issue with our *pre-existing* hard reset logic.

After inspecting the [e2e logs](https://github.com/valory-xyz/open-autonomy/actions/runs/5080202476/jobs/9127158218), it seems that it was flaky and was caused by the following situation:
1. A hard reset is successfully performed. 
2. As a result, the `_is_healthy` flag is set to `True`:
https://github.com/valory-xyz/open-autonomy/blob/5c74b2429053ef2e7fe99968c91bfd916bf1dc0f/packages/valory/skills/abstract_round_abci/behaviour_utils.py#L1853
but never reset to `False`.
3. We fail to retrieve the status here and return `False`:

    https://github.com/valory-xyz/open-autonomy/blob/5c74b2429053ef2e7fe99968c91bfd916bf1dc0f/packages/valory/skills/abstract_round_abci/behaviour_utils.py#L1967-L1975

4. We re-enter the method, but the hard reset is effectively bypassed in all subsequent reset attempts, due to this logic:

    https://github.com/valory-xyz/open-autonomy/blob/5c74b2429053ef2e7fe99968c91bfd916bf1dc0f/packages/valory/skills/abstract_round_abci/behaviour_utils.py#L1904

A unit test was edited to fail because of the issue described above on 5c2fb3c61f7b105de8e4fb159def408f95fb6f6d, and was fixed on c2462d40daea487216f00bc6c12d172ee0ae3592.

Moreover, we could probably remove this part:

https://github.com/valory-xyz/open-autonomy/blob/5c74b2429053ef2e7fe99968c91bfd916bf1dc0f/packages/valory/skills/abstract_round_abci/behaviour_utils.py#L1840

as it will always be the case and does not make sense because of the condition required to enter this part of the code:

https://github.com/valory-xyz/open-autonomy/blob/5c74b2429053ef2e7fe99968c91bfd916bf1dc0f/packages/valory/skills/abstract_round_abci/behaviour_utils.py#L1833

cc @0xArdi

0xArdi (approved):

dvilelaf (approved):

PR #1982, opened 2023-05-25
[v0.10.6] Make the chained `ABCI` skill filter more strict on the service analyser

angrybayblade:
## Proposed changes

Adds check to make sure the chained `ABCI` skill has `valory/abstract_round_abci` skill as a dependency so we don't mistake other skills with `is_abstract` set to `false` as chained `ABCI` skills

## Fixes

## Types of changes

## Checklist


autonomy/cli/helpers/analyse.py
```diff
@@ -338,6 +351,10 @@ def _get_chained_abci_skill(
         if skill_config.is_abstract:
             continue
 
+        # Check if the skill has the `abstract_round_abci` skill as a dependency
+        if not _has_abstract_round_abci_skill_as_dependency(skill_config=skill_config):
+            continue
```
angrybayblade, line 356: Makes the abci skill filter even stricter so we don't confuse every skill with `is_abstract` set to `false` as a chained abci skill

0xArdi (approved):

PR #1994, opened 2023-06-12
[WIP][TBD]: Implement pending offences background app

Adamantios:
## Proposed changes

Implements background behaviour, round, and payload for synchronizing the pending offences.

## Fixes

n/a

## Types of changes

## Checklist

## Further comments

n/a


Adamantios:
The failing test on Windows is because the `timegm` returns the time in milliseconds instead of seconds as in other platforms.
The issue is only present on the test, not the actual code, as the `timegm` is used to generate the timestamps to feed the hypothesis `builds` strategy.

A condition to check for the platform should be added there and divide the timestamp by `1000` if on Windows, however, we leave this test failing for now.

0xArdi (approved):

dvilelaf (approved):

PR #1996, opened 2023-06-19
fix: timeout even when there's no message from the connection

0xArdi:
## Fixes

Make sure the timeout gets checked even when there's no message from connection.
Backported from https://github.com/valory-xyz/mech/pull/74

## Types of changes

## Checklist

- [x]  I have read the [CONTRIBUTING](../blob/main/CONTRIBUTING.md) doc


angrybayblade (approved):

packages/valory/skills/abstract_round_abci/behaviour_utils.py
```diff
@@ -357,6 +357,8 @@ def __handle_waiting_for_message(self) -> None:
                 # wait for the next message
                 self.__notified = False
                 self.__message = None
+        else:
+            self.__get_generator_act().send(None)
```
Adamantios, line 361: Cannot `StopIteration` also get raised here?

0xArdi: yes

Adamantios: Then how about:

```python
def __handle_waiting_for_message(self) -> None:
    """Handle an 'act' tick, when waiting for a message."""
    try:
        self.__get_generator_act().send(self.__message)
    except StopIteration:
        self.__handle_stop_iteration()
    finally:
        # wait for the next message
        self.__notified = False
        self.__message = None
```

packages/valory/skills/abstract_round_abci/behaviour_utils.py
```diff
@@ -358,7 +358,10 @@ def __handle_waiting_for_message(self) -> None:
                 self.__notified = False
                 self.__message = None
         else:
-            self.__get_generator_act().send(None)
+            try:
+                self.__get_generator_act().send(self.__message)
+            except StopIteration:
+                self.__handle_stop_iteration()
```
Adamantios: This is more readable:
https://github.com/valory-xyz/open-autonomy/pull/1996#discussion_r1236967696

Adamantios (approved):

PR #2003, opened 2023-06-23
[TBD]: Automatically sync offences

Adamantios:
## Proposed changes

Refactors the core so that it automatically syncs the pending offenses, without requiring the user to add the corresponding background app to the composed abci app.

In order to perform this refactor, the implementation of the pending offenses had to be moved to base.

This was the only way to avoid circular imports when utilizing the pending offenses to the core. The reason is because of the `base.py` which contains a lot of unseparated functionality on a single file.

Ideally, the file should be split into multiple files and the implementation of the pending offenses would not have to be moved.

## Fixes

n/a

## Types of changes

## Checklist

## Further comments

n/a


0xArdi (approved):

packages/valory/skills/abstract_round_abci/background/pending_offences/behaviour.py
dvilelaf, line 1: Why are all these files deleted?

Adamantios: In order to perform this refactor, the implementation of the pending offenses had to be moved to base.

More info about that decision can be seen in the description.

dvilelaf (approved):

PR #2016, opened 2023-07-25
Add support for multiple service deployments on a same device

angrybayblade:
## Proposed changes

This PR adds support for running multiple services on a same device.

## Fixes

fixes https://github.com/valory-xyz/open-autonomy/issues/1857

## Types of changes

What types of changes does your code introduce? (A **breaking change** is a fix or feature that would cause existing functionality and APIs to not work as expected.)
_Put an `x` in the box that applies_

- [ ] Non-breaking fix (non-breaking change which fixes an issue)
- [ ] Breaking fix (breaking change which fixes an issue)
- [ ] Non-breaking feature (non-breaking change which adds functionality)
- [ ] Breaking feature (breaking change which adds functionality)
- [ ] Refactor (non-breaking change which changes implementation)
- [ ] Messy (mixture of the above - requires explanation!)

## Checklist

_Put an `x` in the boxes that apply._

- [ ] I have read the [CONTRIBUTING](../blob/main/CONTRIBUTING.md) doc
- [ ] I am making a pull request against the `main` branch (left side). Also you should start your branch off our `main`.
- [ ] Lint and unit tests pass locally with my changes
- [ ] I have added tests that prove my fix is effective or that my feature works
- [ ] I have locally run services that could be impacted and they do not present failures derived from my changes
- [ ] Public-facing documentation has been updated with the changes affected by this PR. Even if the provided contents are not in their final form, all significant information must be included.
- [ ] Any backwards-incompatible/breaking change has been clearly documented in the upgrading document.

## Further comments

If this is a relatively large or complex change, kick off the discussion by explaining why you chose the solution you did and what alternatives you considered, etc...

PR #2019, opened 2023-07-28
Release v0.10.10.post1

angrybayblade:
## Release summary

Version number: v0.10.10.post1

## Release details

Autonomy:
- Update `tendermint` testnet command generator to accommodate for new container naming format 

## Checklist

- [ ] I have read the [CONTRIBUTING](../main/CONTRIBUTING.md) doc
- [ ] I am making a pull request against the `main` branch (left side), from `develop`
- [ ] I've updated the dependencies versions to the latest, wherever is possible.
- [ ] Lint and unit tests pass locally (please run tests also manually, not only with `tox`)
- [ ] I built the documentation and updated it with the latest changes
- [ ] I've added an item in `HISTORY.md` for this release
- [ ] I bumped the version number in the `__init__.py` file.
- [ ] I published the latest version on TestPyPI and checked that the following command work:
       ```pip install project-name==<version-number> --index-url https://test.pypi.org/simple --force --no-cache-dir --no-deps```
- [ ] After merging the PR, I'll publish the build also on PyPI. Then, I'll make sure the following
      command will work:
      ```pip install project-name-template==<version_number> --force --no-cache-dir --no-deps```  
- [ ] After merging the PR, I'll tag the repo with `v${VERSION_NUMVER}` (e.g. `v0.1.2`)

## Further comments

Write here any other comment about the release, if any.


tests/test_base.py
```diff
@@ -62,7 +62,7 @@ def get_test_files(package_type: PackageType) -> List[Path]:
 
 def test_version() -> None:
     """Test the version."""
-    assert autonomy.__version__ == "0.10.10"
+    assert autonomy.__version__ == "0.10.10.post1.post1"
```
Adamantios, line 65:
```suggestion
    assert autonomy.__version__ == "0.10.10.post1"
```

autonomy/deploy/generators/docker_compose/base.py
```diff
@@ -215,13 +215,11 @@ def generate_config_tendermint(self) -> "DockerComposeGenerator":
...
-                [
-                    f"--hostname=node{k}"
-                    for k in range(self.service_builder.service.number_of_agents)
-                ]
-            ),
+        hosts = " \\\n".join(
+            [
+                "--hostname=" + self.service_builder.get_tm_container_name(index=k)
+                for k in range(self.service_builder.service.number_of_agents)
+            ]
```
Adamantios, line 222: Can you explain how this resolves the issue?

angrybayblade: Before the fix the hostnames were being formatted as `--hostname=node{NODE_ID}`, with this change they'll be formatted as `--hostname={SERVICE_NAME}_tm_{NODE_ID}`

dvilelaf (commented):
Has this been tested against an actual service? For example, the hello world or simple abci service?

autonomy/deploy/generators/docker_compose/base.py
Adamantios: I understand this part. I do not understand the situation though. I do not know what issue this PR is trying to solve. The description does not help. Is it simply something that was missed on #2016? What is the symptom of the issue? Why was `hosts` a `tuple` and now it is a string?

dvilelaf (approved):
Just run this with IEKit. Can confirm it seems to be working now.

autonomy/deploy/generators/docker_compose/base.py
angrybayblade: So, when building deployments that uses pre-established tendermint networks we build a testnet using `tendermint testnet` command and use this a said pre-established tendermint network. Here when defining the number of validators we also define the hostnames for every node using `--hostname=<NAME>`. And this was missed during the last release.

Adamantios (approved):

PR #2040, opened 2023-09-01
feat: add safe nonce check [no ci]

0xArdi:
## Proposed changes

This PR adds support for checking the safe nonce before re-sending a transaction. 

## Types of changes

## Checklist

## Further comments


packages/valory/skills/transaction_settlement_abci/behaviours.py
```diff
@@ -538,10 +548,33 @@ def _check_tx_history(
...
+
+        self.context.logger.info(
+            f"A transaction with nonce {safe_nonce} has already been sent. "
+        )
         self.context.logger.info(
-            f"Starting check for the transaction history: {self.history}."
+            f"Starting check for the transaction history: {self.history}. "
         )
-        for tx_hash in self.history[::-1]:
+        for tx_hash in self.history:
```
0xArdi, line 577: we check even the current transaction, just to account for any issues that could've been in the prev. round

Adamantios: Why not in reverse order? The logic here was that the more gas we spend in fees, the more probable for the tx to have passed.

Adamantios (approved):

packages/valory/skills/transaction_settlement_abci/behaviours.py
0xArdi: I see, that makes sense. I can revert this.

angrybayblade (approved):

PR #2068, opened 2023-09-28
Make fallback handler address configurable when deploying a service

### LINKED ISSUE #2036: Account for service Safe contracts' callback handlers when created on other chains
As of now, the default Safe multisig is created via the CLI with the default callback handler specified in the following file:
https://github.com/safe-global/safe-deployments/blob/main/src/assets/v1.3.0/compatibility_fallback_handler.json

The address of the handler is hardcoded in the default payload:
https://github.com/valory-xyz/open-autonomy/blob/42f100826f447b89b3f2e719ece165c329a331dc/autonomy/chain/service.py#L53.

However, that callback handler is not the same for every chain. Please refer to the same file to find out differences in addresses on other chains and be aware that the adjustment to the default payload is needed once the protocol is deployed on other chains.

angrybayblade:
## Proposed changes

This PR replaces the deployment payload CLI argument with the fallback handler address argument

## Fixes

Fixes #2036 

## Types of changes

## Checklist

### DIFF
--- a/autonomy/cli/service.py
+++ b/autonomy/cli/service.py
@@ -142,10 +142,10 @@ def _register(  # pylint: disable=too-many-arguments
 @hwi_flag
 @password_decorator
 @click.option(
-    "-d",
-    "--deployment-payload",
-    type=int,
-    help="Deployment payload value",
+    "-f",
+    "--fallback-handler",
+    type=str,
+    help="Fallback handler address for the gnosis safe multisig",
 )
 def _deploy(  # pylint: disable=too-many-arguments
     ctx: Context,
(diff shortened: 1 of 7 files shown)


autonomy/cli/service.py
```diff
@@ -142,10 +142,10 @@ def _register(  # pylint: disable=too-many-arguments
...
 @password_decorator
 @click.option(
-    "-d",
-    "--deployment-payload",
-    type=int,
-    help="Deployment payload value",
+    "-f",
+    "--fallback-handler",
+    type=str,
+    help="Fallback handler address for the gnosis safe multisig",
```
angrybayblade, line 148: Provide custom fallback handler using `-f/--fallback-handler` flag, eg

`autonomy deploy service ... --fallback-handler ADDRESS`

default is `0xf48f2b2d2a534e402487b3ee7c18c33aec0fe5e4`

jmoreira-valory (approved):

kupermind (approved):

PR #2109, opened 2023-11-20
Fix/multichain support

0xArdi:
## Fixes

This PR fixes multiple issues with multichain support. 

## Types of changes

## Checklist


packages/valory/skills/transaction_settlement_abci/skill.yaml
```diff
@@ -140,7 +140,7 @@ models:
       use_slashing: false
       use_termination: false
       validate_timeout: 1205
-      default_chain_name: ethereum
+      default_chain_id: ethereum
```
0xArdi, line 143: breaking

DavidMinarsch (approved):

packages/valory/skills/transaction_settlement_abci/skill.yaml
DavidMinarsch: So why are we changing this? Chain id is usually an integer. Chain name is the ecosystem

0xArdi: the ledger connection expects it as `chain_id`. I agree `chain_name` makes more sense.

DavidMinarsch: ok - well let's open an issue so we don't forget. Shouldn't block us now to move forward.

angrybayblade (approved):

packages/valory/skills/transaction_settlement_abci/skill.yaml
0xArdi: https://github.com/valory-xyz/open-aea/issues/694

PR #2110, opened 2023-11-21
Support for logging to console on tendermint com server

angrybayblade:
## Proposed changes

Adds support for logging to console on the tendermint com server

## Fixes

## Types of changes

## Checklist


angrybayblade (commented):
Tested this locally with trader

deployments/Dockerfiles/tendermint/tendermint.py
```diff
@@ -117,47 +118,48 @@ def build_node_command(self, debug: bool = False) -> List[str]:
         ]
         if debug:
             cmd.append("--log_level=debug")
-
         if self.home is not None:  # pragma: nocover
             cmd += ["--home", self.home]
         return cmd
 
     @staticmethod
-    def get_node_command_kwargs(monitoring: bool = False) -> Dict:
```
DavidMinarsch, line 126: Hm this approach was more generic. Why are we forcing the approach now?

deployments/Dockerfiles/tendermint/tendermint.py
```diff
@@ -174,35 +176,64 @@ def init(self) -> None:
         cmd = self._build_init_command()
         subprocess.call(cmd)  # nosec
 
-    def start(self, start_monitoring: bool = False, debug: bool = False) -> None:
-        """Start a Tendermint node process."""
-        self._start_tm_process(start_monitoring, debug)
-        if start_monitoring:
-            self._start_monitoring_thread()
+    def _monitor_tendermint_process(
```
DavidMinarsch, line 179: Is this a complete copy from below?

DavidMinarsch (commented):
Not sure what the reason was for removing the option to select the type of monitoring. Isn't it simpler to just modify the defaults?

solarw:
@angrybayblade 
tendermint already works well with printing lofs to stdout
i just set log file env variable to /dev/null to avoid on-disk files usage.

so now i can just set WRITE_TO_LOG=false?  and it will stop writing logs to file?

angrybayblade:
> @angrybayblade tendermint already works well with printing lofs to stdout i just set log file env variable to /dev/null to avoid on-disk files usage.
> 
> so now i can just set WRITE_TO_LOG=false? and it will stop writing logs to file?

Yes

deployments/Dockerfiles/tendermint/tendermint.py
angrybayblade: Yes

solarw (approved):
LGTM

deployments/Dockerfiles/tendermint/tendermint.py
DavidMinarsch: No monitoring was a flag that got passed down. So it was possible to configure.

DavidMinarsch (changes requested):
What exactly is this PR trying to achieve?

It does change some default behaviour and adds support to configure via env variable and moves some code around. Can we be specific on what we actually want from this PR?

DavidMinarsch (approved):

solarw (approved):
lgtm

0xArdi (approved):

0xArdi (approved):

PR #2126, opened 2023-12-11
Support for tendermint image on arm platforms

angrybayblade:
## Proposed changes

This PR fixes the Dockerfile to install platform specific tendermint binaries

## Fixes

## Types of changes

## Checklist

### DIFF
--- a/deployments/Dockerfiles/tendermint/install.sh
+++ b/deployments/Dockerfiles/tendermint/install.sh
@@ -1,12 +1,18 @@
 # Install architecture specific version of tendermint
 
-if [[ "$(uname -m)" == "amd64" ]]
+echo Fetching binaries for $(uname -m)
+if [[ "$(uname -m)" == "aarch64" ]]
 then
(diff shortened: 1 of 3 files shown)


deployments/Dockerfiles/tendermint/install.sh
```diff
@@ -0,0 +1,12 @@
+# Install architecture specific version of tendermint
+
+if [[ "$(uname -m)" == "amd64" ]]
+then
+    curl -L https://github.com/tendermint/tendermint/releases/download/v0.34.19/tendermint_0.34.19_linux_amd64.tar.gz > tendermint.tar.gz
+else
```
0xArdi, line 6: why not specifically check for arm? and error out on else?

angrybayblade: I see, that would be futile since we only setup for building images on arm and amd platforms

0xArdi: the current check assumes that if its not amd64 its arm64

0xArdi (approved):

PR #2153, opened 2024-01-30
Support for specifying resources using CLI flags and framework level environment variables

angrybayblade:
## Proposed changes

## Fixes

## Types of changes

## Checklist

### DIFF
--- a/autonomy/cli/deploy.py
+++ b/autonomy/cli/deploy.py
@@ -1,7 +1,7 @@
 # -*- coding: utf-8 -*-
 # ------------------------------------------------------------------------------
 #
-#   Copyright 2022-2023 Valory AG
+#   Copyright 2022-2024 Valory AG
 #
 #   Licensed under the Apache License, Version 2.0 (the "License");
 #   you may not use this file except in compliance with the License.
--- a/autonomy/cli/deploy.py
+++ b/autonomy/cli/deploy.py
@@ -162,6 +162,8 @@ def deploy_group(
     help="Use local tendermint chain setup.",
 )
 @click.option("--image-version", type=str, help="Define runtime image version.")
+@click.option("--agent-cpu", type=float, help="Set agent CPU usage limit.")
+@click.option("--agent-memory", type=int, help="Set agent memory usage limit.")
 @registry_flag()
 @password_option(confirmation_prompt=True)
 @image_author_option
@@ -185,6 +187,8 @@ def build_deployment_command(  # pylint: disable=too-many-arguments, too-many-lo
     use_acn: bool = False,
     use_tm_testnet_setup: bool = False,
     image_author: Optional[str] = None,
+    agent_cpu: Optional[float] = None,
+    agent_memory: Optional[int] = None,
 ) -> None:
     """Build deployment setup for n agents."""
     if password is not None:  # pragma: nocover
@@ -237,6 +241,7 @@ def build_deployment_command(  # pylint: disable=too-many-arguments, too-many-lo
             use_acn=use_acn,
             use_tm_testnet_setup=use_tm_testnet_setup,
             image_author=image_author,
+            resources={"agent": {"cpu": agent_cpu, "memory": agent_memory}},
         )
     except (NotValidKeysFile, FileNotFoundError, FileExistsError) as e:
         shutil.rmtree(build_dir)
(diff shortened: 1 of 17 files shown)


docs/advanced_reference/commands/autonomy_deploy.md
```diff
@@ -157,6 +163,37 @@ export OPEN_AUTONOMY_PRIVATE_KEY_PASSWORD=PASSWORD
...
+        resources:
+          limits:
+            cpu: '1.0'
+            memory: 2048Mi
+          requests:
+            cpu: '0.5'
+            memory: 1024Mi
+```
+
+on kubernetes deployment.
```
angrybayblade, line 195: @solarw @jmoreira-valory @Adamantios

docs/advanced_reference/commands/autonomy_deploy.md
```diff
@@ -157,6 +163,37 @@ export OPEN_AUTONOMY_PRIVATE_KEY_PASSWORD=PASSWORD
...
+
+```bash
+autonomy deploy build --agent-cpu 1.0 --agent-memory 2048
+```
+
+will produce agent nodes with following resource constrains
+
+```yaml
+  xyz_abci_0:
+    mem_reservation: 256M
```
DavidMinarsch, line 178: How does one specify the reservation/requests? Seems like this PR should also allow for this.

autonomy/deploy/generators/kubernetes/templates.py
```diff
@@ -252,11 +252,11 @@
...
         resources:
           limits:
-            memory: "1512Mi"
-            cpu: "0.5"
+            memory: "{agent_memory}Mi"
+            cpu: "{agent_cpu}"
           requests:
             cpu: "0.5"
-            memory: "1512Mi"
+            memory: "1024Mi"
```
DavidMinarsch, line 259: Why do we not also support the modification of the requests/reservation in the same PR?

DavidMinarsch: Also, we are changing the default here which might have unexpected consequences.

autonomy/cli/deploy.py
```diff
@@ -162,6 +162,8 @@ def deploy_group(
     help="Use local tendermint chain setup.",
 )
 @click.option("--image-version", type=str, help="Define runtime image version.")
+@click.option("--agent-cpu", type=float, help="Set agent CPU usage limit.")
+@click.option("--agent-memory", type=int, help="Set agent memory usage limit.")
```
DavidMinarsch, line 166: See my comments elsewhere. These are limits and we should be explicit about the name imo. Because there are also "requests" which imo should be configurable too.

autonomy/deploy/generators/kubernetes/templates.py
angrybayblade: Sorry, this was a mistake

angrybayblade: Reverted

solarw:
@angrybayblade @DavidMinarsch what if set requested and limit the same?

angrybayblade:
> @angrybayblade @DavidMinarsch what if set requested and limit the same?

Do you mean what the resource request values and limit values are the same?

docs/advanced_reference/commands/autonomy_deploy.md
```diff
@@ -157,6 +169,37 @@ export OPEN_AUTONOMY_PRIVATE_KEY_PASSWORD=PASSWORD
...
+        resources:
+          limits:
+            cpu: '1.0'
+            memory: 2048Mi
+          requests:
+            cpu: '0.5'
+            memory: 1024Mi
+```
+
+on kubernetes deployment.
```
angrybayblade, line 201: @DavidMinarsch addresses your comment

autonomy/deploy/base.py
```diff
@@ -90,6 +91,11 @@
     ]
 }
 
+DEFAULT_AGENT_MEMORY_LIMIT = int(os.environ.get("AUTONOMY_AGENT_MEMORY_LIMIT", 1024))
```
DavidMinarsch, line 94: Note that for kubernetes the default was 1512, is that an issue?

DavidMinarsch (approved):

solarw:
> > @angrybayblade @DavidMinarsch what if set requested and limit the same?
> 
> Do you mean what the resource request values and limit values are the same?

yes, probably can use one value, but need to clarify with eseleos

DavidMinarsch:
> > > @angrybayblade @DavidMinarsch what if set requested and limit the same?
> > 
> > 
> > Do you mean what the resource request values and limit values are the same?
> 
> yes, probably can use one value, but need to clarify with eseleos

Why would we hardcode when we now have an option to modify.

autonomy/deploy/base.py
angrybayblade: I don't think so, event if it is it can be overridden very easily

solarw (approved):
lgtm

PR #2179, opened 2024-03-03
fix: relax dotenv dependency

dvilelaf:
## Proposed changes

* Relax dotenv dependency

## Fixes

n/a

## Types of changes

## Checklist

## Further comments
n/a


autonomy/analyse/service.py
```diff
@@ -1,7 +1,7 @@
 # -*- coding: utf-8 -*-
 # ------------------------------------------------------------------------------
 #
-#   Copyright 2023 Valory AG
+#   Copyright 2023-2024 Valory AG
```
0xArdi, line 4: Why the change?

Pipfile
```diff
@@ -41,7 +41,7 @@ hypothesis = "==6.21.6"
 protobuf = "<4.25.0,>=4.21.6"
 pytz = "==2022.2.1"
 py-ecc = "==6.0.0"
-python-dotenv = ">=0.14.5,<0.18.0"
+python-dotenv = ">=0.14.5,<0.22.0"
```
0xArdi, line 44: I think we will still need the open-aea release first. Otherwise, this wont take effect.

0xArdi (commented):
I think there are a bunch of accidental changes.

Pipfile
dvilelaf: Yes, this is a WIP

autonomy/analyse/service.py
dvilelaf: This happened after running generators. Too many files changed.

angrybayblade (approved):

PR #2212, opened 2024-04-01
Add support for specifying chain_id on `get_transaction_receipt`

### LINKED ISSUE #2209: Default chain cannot be overridden everywhere
Such an example is the `_send_transaction_receipt_request`.

As a result, the transaction settlement does not work properly as the validation round cannot succeed for transactions sent using a non-default chain. It needs several minutes so it can proceed to the check history hashes round which utilizes the `get_contract_api_response` and can use the non-default chain.

There might be more methods that do not support using a different chain than the default one.

angrybayblade:
## Proposed changes

## Fixes

fixes #2209

## Types of changes

## Checklist


packages/valory/skills/abstract_round_abci/behaviour_utils.py
```diff
@@ -1631,7 +1634,11 @@ def get_transaction_receipt(
         :yield: LedgerApiMessage object
         :return: transaction receipt data
         """
-        self._send_transaction_receipt_request(tx_digest, retry_timeout, retry_attempts)
+        if chain_id is None:
+            chain_id = self.params.default_chain_id
+        self._send_transaction_receipt_request(
+            tx_digest, retry_timeout, retry_attempts, chain_id=chain_id
+        )
```
angrybayblade, line 1641: @Adamantios I looked for the methods where we're interacting with ledger/contract, `_send_transaction_receipt_request` was the only one where we could not specify the `chain_id`, if there are any other please let me know

angrybayblade:
The test failures are addressed here - https://github.com/valory-xyz/open-autonomy/pull/2213

Adamantios (dismissed):

Adamantios:
Btw this should not close the referenced issue. If you close this, please open a new one for the transaction settlement skill part.

Adamantios (approved):

PR #2222, opened 2024-04-23
fix: store status offences only when slashing enabled

0xArdi:
## Fixes

This PR makes it s.t. offence status changes are only stored in the DB if slashing is enabled.

## Types of changes

## Checklist


angrybayblade (approved):

packages/valory/skills/registration_abci/rounds.py
```diff
@@ -58,17 +59,23 @@ class RegistrationStartupRound(CollectSameUntilAllRound):
...
     def end_block(self) -> Optional[Tuple[BaseSynchronizedData, Event]]:
         """Process the end of the block."""
         if not self.collection_threshold_reached:
             return None
 
         try:
             _ = self.context.state.round_sequence.offence_status
+            # only use slashing if it is configured and the `use_slashing` is set to True
+            if self.params.use_slashing:
+                self.context.state.round_sequence.enable_slashing()
```
Adamantios, line 76: Why inside the `try`?

0xArdi: Just a style matter, otherwise i need this in after the except:
```
        except SlashingNotConfiguredError:
            self.context.logger.warning("Slashing has not been enabled!")
        else:
            if self.params.use_slashing:
                self.context.state.round_sequence.enable_slashing()
```

Adamantios (approved):

PR #2261, opened 2024-10-02
Docs (autonomy service): Add chain profile options

### LINKED ISSUE #2052: Change the management of all chains to an homogeneous way as profiles.
Currently, the framework has some chains defined with a profile (e.g., Ethereum) and some other chains to be defined with "custom profile" that requires environment variables to be defined (e.g., Polygon).

The suggestion is to use profiles homogeneously for all the chains that we control, and leave "custom chain profile" only for actually user-defined chains.

The reasoning for adopting this solution was to avoid multitude of flags on the CLI (--use-ethereum etc.) However, altough this solves the problem of multitude of flag, it introduces the burden of keeping the user informed of the on-chain changes through the documentation, and the user needs to update a significative number of environment variables .

The current "workflow" would be
1- Some contract gets updated
2- Framework documentation gets updated accordingly
3- **User needs to be aware of this update in the docs and change their environment variables accordingly.** 
4- Use the commands

With the proposed approach, if we have profiles for all relevant chains that the DAO deploys the contracts, the workflow would be:
1- Some contract gets updated
2- Framework updates the profile
3- User needs to use the commands with the latest version of the library. No need to redefine environment variables, etc...

Obiously, both approaches have drawbacks, but the first one seems to be more error-prone for the user, as they need to be very aware of the changes ongoing, and where to find this information. The proposed solution requires "only" that the user has the library updated.

**Bonus:**
Although this would not bring much benefit from the point of view of implementation, seems that the ideal solution would be having a flag --chain-config <profile>  instead of multiple mutually-exclusive flags on the commands. This however, might introduce backwards-incompatible changes.

OjusWiZard:
## Proposed changes

Update the `autonomy service` docs to add the available options for chain profiles (same as in `autonomy mint`)

Closes #2052 

## Types of changes

## Checklist

### DIFF
--- a/docs/advanced_reference/commands/autonomy_service.md
+++ b/docs/advanced_reference/commands/autonomy_service.md
@@ -11,15 +11,46 @@ This command group consists of a number of functionalities to manage [the life c
 `--dry-run`
 : Perform a dry run for the transaction.
 
+`--use-celo`                      
+:   Use the `Celo` chain profile to interact with the Autonolas Protocol registry contracts.
+
+`--use-base-sepolia`              
+:   Use the `Base Sepolia` profile to interact with the Autonolas Protocol registry contracts.
+
+`--use-base`                      
+:   Use the `Base` chain profile to interact with the Autonolas Protocol registry contracts.
+
+`--use-optimistic-sepolia`        
+:   Use the `Optimistic Sepolia` chain profile to interact with the Autonolas Protocol registry contracts.
+
+`--use-optimistic`                
+:   Use the `Optimistic` chain profile to interact with the Autonolas Protocol registry contracts.
+
+`--use-arbitrum-sepolia`          
+:   Use the `Arbitrum Sepolia` chain profile to interact with the Autonolas Protocol registry contracts.
+
+`--use-arbitrum-one`              
+:   Use the `Arbitrum One` chain profile to interact with the Autonolas Protocol registry contracts.
+
+`--use-chiado`                    
+:   Use the `Chiado` chain profile to interact with the Autonolas Protocol registry contracts.
+
+`--use-gnosis`                    
+:   Use the `Gnosis` chain profile to interact with the Autonolas Protocol registry contracts.
+
+`--use-polygon-mumbai`            
+:   Use the `Polygon Mumbai` chain profile to interact with the Autonolas Protocol registry contracts.
+
+`--use-polygon`                   
+:   Use the `Polygon` chain profile to interact with the Autonolas Protocol registry contracts.
+
 `--use-ethereum`
-: Use the Ethereum chain profile to interact with the Autonolas Protocol registry contracts. This option requires that you define the following environment variable:
+:   Use the `Ethereum` chain profile to interact with the Autonolas Protocol registry contracts.
 
-    - `ETHEREUM_CHAIN_RPC` : RPC endpoint for the Ethereum mainnet chain.
+`--use-goerli`                    
+:   Use the `Goerli` chain profile to interact with the Autonolas Protocol registry contracts.
 
-`--use-goerli`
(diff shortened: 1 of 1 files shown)


Adamantios (approved):

docs/advanced_reference/commands/autonomy_service.md
```diff
@@ -11,15 +11,46 @@ This command group consists of a number of functionalities to manage [the life c
...
+`--use-ethereum`
+:   Use the `Ethereum` chain profile to interact with the Autonolas Protocol registry contracts.
 
-`--use-goerli`
-: Use the Görli chain profile to interact with the Autonolas Protocol registry contracts. This option requires that you define the following environment variable:
+`--use-goerli`                    
+:   Use the `Goerli` chain profile to interact with the Autonolas Protocol registry contracts.
 
-    - `GOERLI_CHAIN_RPC` : RPC endpoint for the Görli testnet chain.
+To use these chain profile, you will have to export an environment variable for RPC in `<CHAIN_NAME>_CHAIN_RPC` format. For example if you want to use `ethereum`, you will have to export `ETHEREUM_CHAIN_RPC` and for `polygon-mumbai` it would be `POLYGON_MUMBAI_CHAIN_RPC`
```
jmoreira-valory, line 53:
```suggestion
To use these chain profiles, you have to export an environment variable for RPC in `<CHAIN_NAME>_CHAIN_RPC` format. For example if you want to use Ethereum (`--use-ethereum`), you have to export `ETHEREUM_CHAIN_RPC`. For Polygon Mumbai (`--use-polygon-mumbai`) you have to export `POLYGON_MUMBAI_CHAIN_RPC`, and so on.
```

docs/advanced_reference/commands/autonomy_service.md
```diff
@@ -11,15 +11,46 @@ This command group consists of a number of functionalities to manage [the life c
 `--dry-run`
 : Perform a dry run for the transaction.
 
-`--use-ethereum`
-: Use the Ethereum chain profile to interact with the Autonolas Protocol registry contracts. This option requires that you define the following environment variable:
+`--use-celo`                      
+:   Use the `Celo` chain profile to interact with the Autonolas Protocol registry contracts.
```
jmoreira-valory, line 15:
```suggestion
: Use the Celo chain profile to interact with the Autonolas Protocol registry contracts.
```
Please apply these changes on the rest of entries (leave only 1 space after :   and remove code quotes on chain names)

docs/advanced_reference/commands/autonomy_service.md
```diff
@@ -11,15 +11,46 @@ This command group consists of a number of functionalities to manage [the life c
...
+
+`--use-polygon`                   
+:   Use the `Polygon` chain profile to interact with the Autonolas Protocol registry contracts.
+
+`--use-ethereum`
+:   Use the `Ethereum` chain profile to interact with the Autonolas Protocol registry contracts.
 
-`--use-goerli`
-: Use the Görli chain profile to interact with the Autonolas Protocol registry contracts. This option requires that you define the following environment variable:
+`--use-goerli`
```
jmoreira-valory, line 50:
```suggestion
`--use-goerli`
```

docs/advanced_reference/commands/autonomy_service.md
```diff
@@ -49,7 +80,7 @@ This command group consists of a number of functionalities to manage [the life c
 
 !!! note
 
-    The options `--use-ethereum`, `--use-goerli`, `--use-custom-chain` and `--use-local` are mutually exclusive.
+    The chain profile flags are mutually exclusive.
```
jmoreira-valory, line 83:
```suggestion
    The chain profile flags (`--use-ethereum`, etc.) are mutually exclusive.
```

PR #2265, opened 2024-10-07
Fix: Add a small hash id in build_dir

### LINKED ISSUE #2264: When stopping a service using `autonomy deploy stop`, other services on the system also stops
### Subject of the issue
If there are two services running on the same machine, for example: a `hello-world` service and a `trader` service.
When we use `autonomy deploy stop --build-dir trader_service/abci_build` to stop just the `trader`, the `hello-world` service also stops.

### Your environment
- OS: Pop!_OS 22.04 LTS x86_64
- Python version: 3.11.10
- Package Version 0.14.12

### Steps to reproduce
1. Start the `hello-world` service following this: https://github.com/valory-xyz/hello-world?tab=readme-ov-file#prepare-the-environment
2. Start the `trader` service using [trader-quickstart](https://github.com/valory-xyz/trader)
3. Stop the `trader` service using `stop_service.sh`

### Expected behaviour
`hello-world` shouldn't stop

### Actual behaviour
`hello-world` also stops

OjusWiZard:
## Proposed changes

This is done to differentiate docker compose projects, and support running multiple services in parallel.

## Fixes

Closes #2264 
When we do `docker compose up` (happens inside `autonomy deploy run`), docker by default takes `abci_build` as the project name because that's the directory where `docker-compose.yaml` is present. So, any service we start, all of them are grouped by this common name. And when we stop, `docker compose down` stops all the container in this project name (which also included other services).

## Types of changes

## Checklist

### DIFF
--- a/tests/test_autonomy/test_cli/test_deploy/test_build/test_deployment.py
+++ b/tests/test_autonomy/test_cli/test_deploy/test_build/test_deployment.py
@@ -34,6 +34,7 @@ from aea_test_autonomy.configurations import (
     ETHEREUM_ENCRYPTION_PASSWORD,
 )
 
+from autonomy.cli.helpers.deployment import build_hash_id
 from autonomy.constants import DEFAULT_BUILD_FOLDER, DEFAULT_DOCKER_IMAGE_AUTHOR
 from autonomy.deploy.base import (
     DEFAULT_AGENT_CPU_LIMIT,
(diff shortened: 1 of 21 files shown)


dvilelaf (approved):

tests/test_autonomy/test_cli/test_deploy/test_build/test_deployment.py
```diff
@@ -157,13 +158,13 @@ def test_docker_compose_build(
     ) -> None:
         """Run tests."""
 
-        build_dir = self.t / DEFAULT_BUILD_FOLDER
+        build_dir = self.t / DEFAULT_BUILD_FOLDER.format(build_hash_id())
```
Adamantios, line 161: This is repeated a lot, might be better to define a new constant.

Adamantios (approved):

PR #2276, opened 2024-10-31
Feat: Strict checks in rounds' for properties

### LINKED ISSUE #2246: Enforce superclasses of base rounds to always define required attributes
**Is your feature request related to a problem? Please describe.**
Some base rounds require a number of base attributes to be defined by the superclasses. For example, [none_type](https://github.com/valory-xyz/open-autonomy/blob/89a778cd71bbb56ed4805fa95470cd4f160217e7/packages/valory/skills/abstract_round_abci/base.py#L1548) on `CollectSameUntilThresholdRound`. 

If the superclass fails to initialize these attributes, the business logic will proceed without a warning. However, if during the execution variable is required to launch an event, it will [crash the application](https://github.com/valory-xyz/open-autonomy/blob/v0.14.14.post1/packages/valory/skills/abstract_round_abci/base.py#L1603-L1606) when accessing that attribute. 

**Describe the solution you'd like**
In summary, the application should throw a warning or exception when these required attributes are not initialized. This applies to several other base round classes.

**Describe alternatives you've considered**

**Additional context**

OjusWiZard:
## Proposed changes

Strictly check if the required properties are defined in derived classes from `AbstractRound`

## Fixes

#2246

## Types of changes

## Checklist

## Further comments


packages/valory/skills/abstract_round_abci/base.py
```diff
@@ -1790,6 +1824,12 @@ class CollectDifferentUntilThresholdRound(CollectionRound, ABC):
     collection_key: str
     required_block_confirmations: int = 0
 
+    required_class_attributes: Tuple[str, ...] = (
+        "done_event",
+        "collection_key",
+        "required_block_confirmations",
```
jmoreira-valory, line 1830: This isn't usually defined by the developer, is it?

packages/valory/skills/transaction_settlement_abci/rounds.py
```diff
@@ -357,6 +360,12 @@ class SelectKeeperTransactionSubmissionARound(CollectSameUntilThresholdRound):
     no_majority_event = Event.NO_MAJORITY
     collection_key = get_name(SynchronizedData.participant_to_selection)
     selection_key = get_name(SynchronizedData.keepers)
+    required_class_attributes: Tuple[str, ...] = (
+        "done_event",
+        "no_majority_event",
+        "collection_key",
+        "selection_key",
+    )
```
jmoreira-valory, line 368: Why we need this here? doesn't it come from CollectSameUntilThresholdRound?

packages/valory/skills/abstract_round_abci/base.py
OjusWiZard: No, ideally if they are only inheriting it in a new class like `SomeNewRound(CollectDifferentUntilThresholdRound):` then they would just need to specify the attributes as usual, like `done_event = Event.DONE` etc.

Here, **we** set the `required_class_attributes` value internally, to check their present at runtime.

packages/valory/skills/transaction_settlement_abci/rounds.py
OjusWiZard: This is to override `required_class_attributes` in this class because we also override `end_block()` here. Since, there is no need for the `none_event` here, we set a new tuple in `required_class_attributes` with "none_event" removed.

jmoreira-valory (approved):

packages/valory/skills/transaction_settlement_abci/rounds.py
```diff
@@ -340,6 +342,7 @@ class RandomnessTransactionSubmissionRound(CollectSameUntilThresholdRound):
     payload_class = RandomnessPayload
     synchronized_data_class = SynchronizedData
     done_event = Event.DONE
+    none_event = Event.NONE
```
Adamantios, line 345: The none event cannot be triggered here:

https://github.com/valory-xyz/open-autonomy/blob/5c0b0a81160baf6eb58b710fb0bb9eb10e78bc21/packages/valory/skills/squads_transaction_settlement_abci/payloads.py#L31-L32

Let's do the same as in `SelectKeeperTransactionSubmissionARound`.

packages/valory/skills/abstract_round_abci/base.py
```diff
@@ -1549,6 +1561,14 @@ class CollectSameUntilThresholdRound(CollectionRound, ABC):
     collection_key: str
     selection_key: Union[str, Tuple[str, ...]]
 
+    required_class_attributes: Tuple[str, ...] = (
+        "done_event",
+        "no_majority_event",
+        "none_event",
+        "collection_key",
+        "selection_key",
```
Adamantios, line 1569: Let's define constans for these, here in base, so that we avoid repeating them everywhere else.

Adamantios (changes requested):

Adamantios (approved):

jmoreira-valory (approved):

PR #2357, opened 2025-09-09
fix for deployment app-tendermint logging issue with the same file op…

solarw:
…ened

## Proposed changes

## Fixes

## Types of changes

## Checklist

## Further comments


Adamantios (dismissed):

OjusWiZard (changes requested):
Also, I tried running with these changes and something is off...
Now the logs are some lines and then full of empty strings, could you check what's wrong here?

Attaching the logs here.

[node_0.txt](https://github.com/user-attachments/files/22465985/node_0.txt)

DavidMinarsch:
The underlying bug is real — a non-rotating `FileHandler` on the root logger (from `logging.basicConfig(filename=...)`) plus a `RotatingFileHandler` on `TendermintNode` both pointed at the same path, so after rotation the `FileHandler` kept writing to the already-rotated file. Worth fixing. But the current implementation has a few correctness issues I'd want addressed before merge:

1. **Double-writes via propagation.** `tendermint_logger = app.logger.getChild("tendermint")` propagates to its parent by default. The handler is added to `app.logger` and, when `WRITE_TO_LOG=true`, *also* to `tendermint_logger` — so every tendermint line hits the handler twice (once directly, once via propagation to `app.logger`). Either set `tendermint_logger.propagate = False` or only attach the handler to the parent.

2. **`WRITE_TO_LOG=false` no longer disables file logging.** `TendermintNode.log()` was changed from a guarded `if self.write_to_log: self.logger.info(...)` to an unconditional call. Because `tendermint_logger` propagates up to `app.logger` (which has the file handler unconditionally), tendermint output lands in the file even with `WRITE_TO_LOG=false`. The env var is effectively a no-op now.

3. **Likely cause of the empty-strings output @OjusWiZard reported.** With `self.logger.info(line.strip())` now called unconditionally, every blank/whitespace line from Tendermint stdout produces an empty log record (previously suppressed because the old formatter was `%(message)s` and the call was gated). Guarding with `if line.strip():` should fix it.

4. **Regression in `packages/valory/connections/abci/connection.py`.** The `self.node = TendermintNode(self.params, self.logger)` call was changed to `TendermintNode(self.params, None)`. On the agent side there's no Flask `app.logger` to attach to, so this drops the connection's logger entirely and routes tendermint output to the root logger. Looks like a copy-paste slip from the `app.py` changes — should be restored to `self.logger`.

5. **Stale `# pylint: disable=no-member` comments** in `app.py` — the comment got moved from the `.info(...)` line down to the closing `)` line in the `update_params()` edits, so it no longer suppresses the warning it was intended to.

6. **`too-many-locals` pylint disable** was added to `create_app` but only one new local (`log_file`) was introduced; the extra disable doesn't look needed.

Also worth noting: there are no unit tests for the reworked logging path. Codecov's "100% of modified lines covered" is misleading because `deployments/Dockerfiles/tendermint/app.py` is a container entrypoint and not exercised by the normal test suite — the empty-line regression had to be caught by eyeballing `node_0.txt`.

DIvyaNautiyal07:
The bug here is real and worth fixing, but I'd suggest closing this PR and redoing the fix in a smaller, separate one.

On the bug:

- Two file handlers end up open on the same log path: a non-rotating one from `basicConfig` and a rotating one from `TendermintNode`. Every line gets written twice via root propagation. After rotation, the non-rotating handler keeps writing to the renamed file, so disk usage grows silently while the visible `log.log` looks fine.
- Worse, with the default `WRITE_TO_LOG=false`, only the non-rotating handler is active, so there's no rotation at all. Log files grow unbounded forever despite the 50MB cap users probably assume is in effect.

On why this PR isn't the right vehicle:

- Stale: open since Sep 2025, last commit Oct 2025, currently 409 commits behind main, conflicting, e2e failing.
- Unaddressed review feedback. OjusWiZard reported empty-string lines flooding the logs (caused by `TendermintNode.log()` becoming unconditional). A later review listed six more concrete issues: child logger still propagating and double-writing, `WRITE_TO_LOG=false` no longer disabling file output, `connection.py` regression dropping the logger, stale `# pylint: disable=no-member` comments, an unneeded `too-many-locals` disable.
- Wrong scope: 44 files changed, ~40 of them just hash bumps from touching the packaged ABCI connection. For a ~30 line logging bug this is a lot of blast radius and re-locking effort to rebase.

Suggested fresh PR, much smaller:

- Drop `filename=` from `basicConfig`.
- Add a single `RotatingFileHandler` on `app.logger` inside `create_app`.
- Set `propagate=False` on the tendermint child logger.
- Keep `WRITE_TO_LOG` gating, add an `if line.strip():` guard in `TendermintNode.log()`.
- Leave `TendermintNode`'s signature, `connection.py`, and the packaged skill untouched.

Happy to open that one if useful.

PR #2369, opened 2025-11-03
Fix: Retry on TxSettler with unsynced RPC

OjusWiZard:
## Proposed changes

- Retry building and sending transaction with these two errors are returned by the RPC
    - `replacement transaction underpriced`
    - `Could not transact with/call contract function, is contract deployed correctly and chain synced?`

## Types of changes

## Checklist

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -67,6 +67,7 @@ base_deps = [
     "requests-toolbelt==1.0.0",  # Required for graphql client
     "aiohttp<4.0.0,>=3.8.5",
     "typing_extensions<=4.13.2,>=3.10.0.2",
+    "multiaddr==0.0.9"
 ]
 base_deps.extend(all_extras["cli"])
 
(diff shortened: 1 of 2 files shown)


autonomy/chain/tx.py
```diff
@@ -66,6 +66,8 @@
     "already_exists",
     "already known",
     "oldnonce",
+    "replacement transaction underpriced",
+    "is contract deployed correctly and chain synced?",
```
jmoreira-valory: Are we sure about this one here ? This can happen for several reasons:
1. Wrong contract address (real error, tx settler should fail)
2. Out of sinc RPC waiting a recent contract deployment (RPC issue, should retry)

OjusWiZard: Oh you mean if the contract address is wrong, yeah in that case the error will be "is contract deployed". I don't know yet how we can differentiate between the two, but I'd rather have it retry before eventually failing instead of not retrying at all. Because this way, it will be able to handle unsynced RPC with the cost of delay in the failure that is actually because of a "programming" error. Since, the wrong contract error is not flaky, we can avoid that as developers.

jmoreira-valory (approved):

PR #2393, opened 2026-01-31
Add polymarket checks from audit

jmoreira-valory:
## Proposed changes

Add polymarket checks from audit

## Types of changes

## Checklist

## Further comments


OjusWiZard (dismissed):

Adamantios (dismissed):

DavidMinarsch:
Small, focused, defensive PR — both audit-sourced checks are reasonable, but execution is uneven. A few concerns worth addressing before merge:

1. **Asymmetric hardening.** The 32-byte check is added to `get_poly_safe_create_transaction_hash` but not to `get_enable_module_transaction_hash`, even though the latter also returns a `bytes32` that is signed a few lines below in `get_service_manager_deploy_data`. If the audit recommended validating hash format, the same rationale applies to both hashes — an invalid `enable_module_hash` would corrupt `sig2_bytes` just as silently. Either justify the asymmetry in a comment or mirror the check.

2. **Cross-layer import couples the package to the framework.** `packages/valory/contracts/...` now imports `from autonomy.chain.config import ChainType`. AEA packages are designed to be portable and framework-independent — they can be `autonomy fetch`'d into other repos where `autonomy.chain.config` is not importable. This appears to be the only such import in the contract package tree. A safer pattern is to hard-code `POLYGON_CHAIN_ID = 137` locally in the contract module. Same correctness, no framework coupling. As a bonus, `ChainType.POLYGON.id` is typed `Optional[int]` and raises `ValueError` if the chain name isn't in the lookup table — a literal sidesteps that fragility too.

3. **Inconsistent exception types.** `RuntimeError` for the hash length mismatch vs `ValueError` for the chain mismatch. Both are contract-precondition failures — pick one (preferably `ValueError`).

4. **Minor fragility in the error message**: `f"... got {len(hash_bytes)} {type(hash_bytes)}"` will itself raise `TypeError` on an unsized return (e.g., if it came back as an int). Fine in practice given web3's `bytes32` ABI decoding, but you could make it bulletproof by only formatting `len(...)` after the `isinstance(..., bytes)` check has passed.

5. **No tests added.** The PR's own checkbox for tests is unchecked, and there is no existing test file under `packages/valory/contracts/poly_safe_creator_with_recovery_module/`. Audit-originated guards are exactly the kind of thing unit tests should cover — a mock/fake `LedgerApi` that returns the wrong chain_id, and a short `hash_bytes` return, each asserting the expected exception. Without these, a future refactor could quietly remove the guards.

6. **`# pylint: disable=too-many-locals`** for 2 new locals is a band-aid. Inlining `if ledger_api.api.eth.chain_id != POLYGON_CHAIN_ID:` avoids both the disable and the extra locals.

7. **Added RPC cost**: `ledger_api.api.eth.chain_id` is a live JSON-RPC call on every `get_service_manager_deploy_data` invocation (web3.py does not cache it). Negligible on the deploy path, but worth being conscious of.

Non-blocking / out-of-scope note: the `github-advanced-security` bot comment on this PR flags `autonomy/deploy/generators/localhost/tendermint/app.py`, which isn't touched by this diff — it's a base-branch code-scanning finding, not a critique of this change.

DavidMinarsch:
## All review comments addressed

Latest push (fbb9405f2) addresses all 7 points from the review:

1. **Symmetric hardening** — extracted `_validate_hash_bytes()` helper, now applied to both `get_poly_safe_create_transaction_hash` and `get_enable_module_transaction_hash`.

2. **Framework coupling removed** — replaced `from autonomy.chain.config import ChainType` with local `POLYGON_CHAIN_ID = 137`. No more cross-layer import. Note: two other packages also import from `autonomy` (`sign_message_lib` and `erc8004_identity_registry_bridger` both import `MultiSendOperation`) — same issue, trivial to fix (it's a 3-line enum), should be a follow-up.

3. **Consistent exceptions** — both checks now raise `ValueError`.

4. **Safe error message** — `isinstance(hash_bytes, bytes)` check runs first; `len()` only called after confirming it's bytes.

5. **14 unit tests added** — full coverage of `_validate_hash_bytes` (5 tests), `get_poly_safe_create_transaction_hash` (3 tests), `get_enable_module_transaction_hash` (3 tests), `get_service_manager_deploy_data` (3 tests including chain ID mismatch, happy path, and Polygon acceptance).

6. **Removed `pylint: disable=too-many-locals`** — inlined the chain_id check, no extra locals needed. Also removed unused imports (`eth_abi.encode`, `eth_utils.to_checksum_address`).

7. **RPC cost** — noted, no change needed (single call on deploy path).

All linters pass: black, isort, flake8, mypy, pylint, darglint, check-packages, check-api-docs.

DhairyaPatel7 (approved):
## LGTM — all 7 prior review items addressed

Verified at `fbb9405f2`:

| # | Prior finding | Fix |
|---|---|---|
| 1 | Asymmetric hardening | ✓ `_validate_hash_bytes()` applied to both `get_poly_safe_create_transaction_hash` and `get_enable_module_transaction_hash` |
| 2 | `ChainType` framework coupling | ✓ Replaced with local `POLYGON_CHAIN_ID = 137` |
| 3 | Inconsistent exceptions | ✓ Both checks raise `ValueError` |
| 4 | Fragile error message | ✓ `isinstance` guard before `len()` |
| 5 | No tests | ✓ 14 tests (5 validator + 3 create + 3 enable + 3 deploy-data) |
| 6 | `pylint: disable=too-many-locals` | ✓ Removed; chain_id check inlined; unused `eth_abi.encode` / `eth_utils.to_checksum_address` also dropped |
| 7 | RPC cost note | Acknowledged, no code change needed |

### Minor nits (non-blocking)

1. **`_validate_hash_bytes` return value unused at call sites** (`contract.py:74, 94`). The helper returns `bytes` but callers do `_validate_hash_bytes(hash_bytes)` then keep using the original name. Either assign back (`hash_bytes = _validate_hash_bytes(hash_bytes)`) or change the return to `None` — the side-effect (raise) is the real contract.
2. **Happy-path tests for `get_service_manager_deploy_data` under-assert** (`test_contract.py:192, 217`). Both only assert `"data_bytes" in result`; `codec.encode` is stubbed to return `b"\x00" * 64`, so the test effectively checks "call chain ran without exception." Worth pinning down (a) that both `crypto.sign_message` calls receive the correct hashes from the two contract calls, and (b) that `codec.encode` is called with `["(uint8,bytes32,bytes32)", "bytes"]` and the correct `(v, r, s)` decomposition (`r=sig[0:32]`, `s=sig[32:64]`, `v=sig[64]`). That pins the signature-packing layout, which is the audit-relevant invariant.
3. **Validator-propagation through `get_service_manager_deploy_data` not covered.** One test that returns a short hash from the inner call and asserts `ValueError` from `get_service_manager_deploy_data` would close that cheaply.
4. **PR description is still the raw template** — tests are now added, so please tick the box and drop the template fluff.
5. **Follow-up for `MultiSendOperation` cross-layer import** in `sign_message_lib` and `erc8004_identity_registry_bridger` — noted as out-of-scope in the response comment, please file a tracking issue so it doesn't drop.

OjusWiZard (approved):

PR #2402, opened 2026-02-12
Cap suggested sleep time

jmoreira-valory:
## Proposed changes

Suggested sleep time may grow exponentially, causing to pass out-of-bound values to the sleep method down the call stack.

## Types of changes

## Checklist

## Further comments


LOCKhart07 (approved):

Adamantios (changes requested):
This should be apparent to the dev setting the env vars for the sleep time. Instead, I suggest checking the backoff factor && retries in the model during the params init, warning the user, and failing early.

Adamantios:
> @Adamantios even if doing like that the problem will also happen. How do you suggest to proceed?

If we were to go for that solution, this message would let the dev understand that they are using irrational values for the backoff + retries, giving them a chance to revisit and reset them.

Another option would be to go for the solution you already implemented and warn the user, e.g.:

```sleep time capped to x sec```

DavidMinarsch:
Picked this up for review — the bug is real and the fix is worth landing, but it's blocked on a reasonable review comment that was never addressed, and there are a few gaps worth calling out before it moves.

## The bug is real

`RetriesInfo.suggested_sleep_time` returns `backoff_factor ** retries_attempted` uncapped. With the default backoff of 2.0, you don't need many retries before this explodes — after ~30 retries you're asking the agent to sleep ~34 years. `BaseBehaviour.sleep()` (`behaviour_utils.py:261`) does `datetime.now() + timedelta(0, seconds)`, which raises `OverflowError: date value out of range` once `seconds` pushes past year 9999 (~38 retries at default backoff). Even before the overflow, any huge value effectively hangs the agent. The branch name confirms this crashed a real service.

## What's good about this fix

Minimal, targeted, 2 substantive lines. 1 hour is a sensible upper bound for "suggested backoff sleep." The safety net applies regardless of why retries happen — transient API outage, config mistake, persistent remote failure, etc.

## What's missing

1. **No test for the cap.** `test_suggested_sleep_time` parametrizes `retries ∈ range(10)` → max 512s, well below the 3600s cap. The new behavior is completely untested. A parametrized case at e.g. `retries=20` would verify the cap kicks in.

2. **No warning/log when the cap fires.** This is exactly the compromise @Adamantios suggested in the review thread: *"go for the solution you already implemented and warn the user, e.g. 'sleep time capped to x sec'"*. Without it, operators see steady 1-hour sleeps and have no signal that something is misconfigured or a remote is persistently failing. A one-time `WARNING` log the first time the cap is hit would satisfy the change request and cost ~3 lines.

3. **The design debate was never closed.** @Adamantios wanted init-time validation (check `backoff_factor ** max_retries` at param init). @jmoreira-valory pushed back correctly that init validation can't prevent the runtime crash if real retries exceed the assumed max. These aren't mutually exclusive — the runtime cap is the actual safety net, the init check is an ergonomic nudge — but the discussion stalled. Landing the runtime cap first (with a warning) is the right call; init-time validation can be a follow-up.

4. **Post-fix behavior is better but not great.** An agent that's hit the cap keeps retrying forever at 1-hour intervals. That's *better than crashing* but probably still not what operators want. Out of scope for this PR, but worth flagging as a follow-up conversation.

5. **`MAX_SUGGESTED_SLEEP_TIME = 3600` is a hardcoded magic number.** A one-line comment explaining the rationale (1 hour, and/or "stay well below the datetime overflow threshold") would help.

6. **`Breaking fix` is mismarked.** Capping a sleep time isn't a breaking change — it's a semantic fix. This should be marked as a non-breaking fix.

## PR hygiene

- 2 months stale, unresolved `CHANGES_REQUESTED` review.
- `mergeStateStatus: DIRTY` — conflicts with main from hash drift since February.
- `lock_check (windows-latest)` failing on the stale run (likely pre-existing, but needs re-verification post-rebase).
- All PR checklist items are unchecked (contributing doc, local lint/tests, tests added, docs, upgrading notes).

## Recommended path forward

~30 minutes of work to unblock:

1. Add a parametrized test case at high retries that asserts the cap.
2. Add a one-time `self.context.logger.warning(...)` when the cap is first hit — this directly answers @Adamantios's change request.
3. Add a one-line comment justifying the 3600s constant.
4. Fix the "Breaking fix" marking.
5. Rebase on main, regenerate hashes (`autonomy packages lock`), update `ABSTRACT_ROUND_ABCI_SKILL_WITH_HASH` in `autonomy/constants.py`.
6. Mark the review comment resolved, re-request review.

If @jmoreira-valory isn't driving this anymore, it would be faster to open a fresh PR from main with the fix + test + warning than to keep this one rotting. The bug is a crash — someone should pick it up.

DIvyaNautiyal07:
The bug here is real and the fix is correct. This one's worth landing, just needs a small completion pass.

On the bug:

- `RetriesInfo.suggested_sleep_time` returns `backoff_factor ** retries_attempted` with no upper bound. With the default backoff of 2.0, around 30 retries gives ~34 years in seconds, and by ~38 retries `BaseBehaviour.sleep()` overflows when building `datetime.now() + timedelta(0, seconds)` and crashes the agent. The branch name confirms this hit a real service.
- Capping the value at 3600s is the right safety net. It's small, targeted, and applies regardless of why retries pile up (transient remote, config error, persistent failure).

On what's still open:

- @Adamantios's first ask (init-time validation of `backoff_factor ** max_retries`) and his second ask (a warning log when the cap fires) are complementary, not competing. The author was right that init-time validation alone can't prevent the runtime crash, but the warning log is a clean compromise that satisfies the change request without expanding scope.
- No test exercises the cap. The existing `test_suggested_sleep_time` parametrizes `retries ∈ range(10)`, max value 512s, well below the 3600s cap. The new behavior is currently untested.
- `MAX_SUGGESTED_SLEEP_TIME = 3600` could use a one-line comment, and the "Breaking fix" marking on the PR is incorrect (this is a non-breaking fix).

On the PR itself:

- Open since Feb 2026, last commit Feb 2026. About three months idle, much fresher than some of the other stale PRs but still sitting on a crash bug.
- Currently conflicting with main due to hash drift, lock_check failing on windows.
- The substantive code change is two lines. Everything else in the diff is package hash bumps from regenerating after the model edit.

The cleanest path is probably to open a fresh PR from main with the cap plus the missing pieces (test, warning log, constant comment, corrected marking) and let the hash regen happen once at the end. Re-running the lock + rebase on this branch and then layering the additional changes on top is more churn than redoing it cleanly, and it lets the full fix land in one coherent push instead of stacking commits on a stale base. Will reference this PR so the discussion isn't lost.

PR #2416, opened 2026-03-01
Feat/python3.14 compat

DavidMinarsch:
## Proposed changes

  Summary                                                                                                                                       

  - Bump Flask (>=3.1.0,<4.0.0) and Werkzeug (>=3.1.0,<4.0.0) — old versions used ast.Str which was removed in Python 3.14
  - Fix TypeCheckMixin.__post_init__: use get_type_hints(type(self)) instead of get_type_hints(self) — Python 3.14 no longer supports
  get_type_hints() on dataclass instances
  - Rename setup to setup_method in test classes — pytest no longer calls setup(self) on Python 3.14
  - Fix chain_selection_flag and abci_spec_format_flag click option defaults — the loop pattern caused the last option's default=False to win,
  resolving to the string "False" instead of the intended flag value
  - Fix tox py3.10-* environments: use explicit basepython = python3.10 instead of python3 to avoid conflicts with tox env name resolution

  Test plan

  - Package tests pass locally (Python 3.14) — only pre-existing fuzz test timeouts and IPFS port conflicts remain
  - Framework tests pass locally — remaining failures are Docker-dependent or pre-existing test logic issues
  - CI passes on all platforms (ubuntu, macos, windows) x Python versions (3.10-3.14)

## Types of changes

## Checklist

## Further comments

-

### DIFF
--- a/TENDERMINT_NODE_SYNC.md
+++ b/TENDERMINT_NODE_SYNC.md
@@ -21,8 +21,8 @@ and the test will catch drift.
 
 | File | Purpose | Key differences |
 |------|---------|-----------------|
-| `autonomy/deploy/generators/localhost/tendermint/tendermint.py` | Localhost dev deployment | Direct file I/O logging, optional monitoring param on `start()` |
-| `packages/valory/agents/register_reset/tests/helpers/slow_tendermint_server/tendermint.py` | Test fixture helper | Optional monitoring, conditional stdout piping, simplified logging |
+| `autonomy/deploy/generators/localhost/tendermint/tendermint.py` | Localhost dev deployment | Direct file I/O logging |
+| `packages/valory/agents/register_reset/tests/helpers/slow_tendermint_server/tendermint.py` | Test fixture helper | Optional monitoring, conditional stdout piping |
 
 Architectural differences (not accidental drift):
 - **Monitoring is optional** via `start(start_monitoring=False)` — Group 1 always monitors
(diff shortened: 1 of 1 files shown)


packages/valory/agents/register_reset/tests/helpers/slow_tendermint_server/tendermint.py
```diff
@@ -130,10 +130,7 @@ def get_node_command_kwargs(monitoring: bool = False) -> Dict:
             kwargs["stdout"] = subprocess.PIPE
             kwargs["stderr"] = subprocess.STDOUT
 
-        if platform.system() == "Windows":  # pragma: nocover
```
DavidMinarsch: double check

plugins/aea-test-autonomy/aea_test_autonomy/base_test_classes/contracts.py
```diff
@@ -44,6 +44,8 @@
 )
 from aea_test_autonomy.helpers.contracts import get_register_contract
 
+HARDHAT_CHAIN_ID = 31337
```
DavidMinarsch: Related tests didn't break because of Python 3.14 — they broke because of the web3 v7 upgrade.                                                               

The dependency is web3>=7,<8. In web3 v6, eth.chain_id queried the node directly via RPC, so it returned Hardhat's actual chain_id (31337) regardless of what was configured in the ledger API.
                                                                                                                                                  
In web3 v7, aea_ledger_ethereum added CachedChainIdMiddleware which intercepts eth_chainId calls and returns the configured value (DEFAULT_CHAIN_ID=1337) instead of asking the node. So eth.chain_id now returns 1337, which:

  1. Signs transactions with chain_id 1337 — Hardhat rejects them ("invalid chainId")
  2. Makes is_service_manager_token_compatible_chain() return False (1337 not in the list) — wrong ABI path, MismatchedABI error

The DEFAULT_CHAIN_ID=1337 was always wrong for Hardhat, but it was silently ignored when web3 actually queried the node. The middleware made the misconfiguration visible.

These failures were likely already present on main — you're just encountering them now because the CI suite runs Docker-based contract tests and the web3 v7 migration happened relatively recently.

packages/valory/contracts/service_registry/tests/test_contract.py
```diff
@@ -97,10 +97,16 @@ def test_verify_contract(self, valid_address: bool) -> None:
             contract_address = self.invalid_contract_address
             bytecode += "invalid"
 
-        result = self.contract.verify_contract(
-            self.ledger_api,
-            contract_address,
-        )
+        with mock.patch.object(
```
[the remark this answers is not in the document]
OjusWiZard: `self.ledger_api.api.eth.chain_id` is the [flaky bug on CI](https://github.com/valory-xyz/open-aea/actions/runs/21075576075/job/60628932098?pr=826#step:6:3017).
The chain_id is being returned as 1337, however it should be set as 31337.

PR #2417, opened 2026-03-01
Fix: handle missing skill packages in analyse service command (#2373)

### LINKED ISSUE #2373: Fix tox -e analyse-service
More details [here](https://github.com/valory-xyz/trader/issues/589).

DavidMinarsch:
Skip skills whose packages are not available locally (e.g. third-party packages that have not been synced) instead of crashing with FileNotFoundError. Also move the is_abstract override check before loading the skill package to avoid unnecessary loads.

## Proposed changes

 Summary                                                                                                                                         
                                                                                                                                                  
  - Fixes _get_chained_abci_skill to gracefully skip skills whose packages are not available locally (e.g. third-party packages that have not been
   synced) instead of crashing with FileNotFoundError                                                                                             
  - Moves the is_abstract override check before loading the skill package to avoid unnecessary loads                                              
  - Adds test for the new behavior with a non-ABCI skill whose package is missing                                                                 

  Test plan

  - All 21 verify_service tests pass (20 existing + 1 new)
  - Verify tox -e analyse-service works in downstream repos (trader, mech) with third-party skill dependencies

Fixes #2373

## Types of changes

## Checklist

## Further comments

-

### DIFF
--- a/autonomy/cli/helpers/analyse.py
+++ b/autonomy/cli/helpers/analyse.py
@@ -326,24 +326,29 @@ def _get_chained_abci_skill(
         if override is None:
             continue
 
+        # Check if the skill override has the `is_abstract` property set to true
+        if override.get("is_abstract", False):
+            continue
+
         package_id = PackageId(
             package_type=PackageType.SKILL,
             public_id=skill_id,
         )
 
-        skill_config = cast(
-            SkillConfig,
-            (
-                _load_from_ipfs(package_id=package_id)
-                if is_on_chain_check
-                else _load_from_local(
-                    package_id=package_id, package_manager=package_manager
-                )
-            ),
-        )
-
-        # Check if the skill override has the `is_abstract` property set to true
-        if override.get("is_abstract", False):
+        try:
+            skill_config = cast(
+                SkillConfig,
+                (
+                    _load_from_ipfs(package_id=package_id)
+                    if is_on_chain_check
+                    else _load_from_local(
+                        package_id=package_id, package_manager=package_manager
+                    )
+                ),
+            )
+        except FileNotFoundError:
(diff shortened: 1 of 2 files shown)


DavidMinarsch:
Closing this PR — on reflection I don't think this is the right fix, expanding on my earlier comment.

**The core change is wrong in shape.** Silently skipping a skill on `FileNotFoundError` turns a legitimate error condition ("package not synced locally" or "service definition references a skill that isn't available") into a no-op. The result is `analyse-service` happily reporting OK for a service that can't actually run. That's worse than a noisy failure — the whole point of `analyse-service` is to catch this class of problem before deployment.

**The `except FileNotFoundError` is also too broad.** `_load_from_local` can surface `FileNotFoundError` from causes other than "the skill package isn't present" — missing nested config, broken symlinks, etc. Those would be silently skipped too.

**The original pain point is already fixed at the correct layer.** The downstream trigger (valory/trader#589) was resolved on 2025-12-09 by valory/trader#585, which fixed the service definition rather than working around it in `open-autonomy`. That's the right shape of fix: if `analyse-service` complains about a missing skill, the service definition or the local package state is what should change, not the analyser's tolerance for missing packages. Issue #2373 was closed the same day this PR opened and hasn't been reopened — the upstream change isn't needed.

**The `is_abstract` reorder is a legitimate micro-optimization** (avoid loading a skill you're about to skip), but it's ~3 lines and not worth keeping a stale PR alive for. Happy to revisit it standalone if it ever matters.

If a future case emerges where a skill genuinely needs to be absent at analyse time (e.g. an optional pluggable skill), the right approach is an explicit "optional skill" marker with a clean error at the load site — not a broad swallow.

PR #2420, opened 2026-03-01
Fix: return dict from service registry get_service_information (#2330)

### LINKED ISSUE #2330: Service registry`get_service_information` does not return value as dict
**Describe the bug**
In order to access contract data from the behaviours there is a `get_contract_api_response` in a `BaseBehaviour`. Currently it's incompatible with `get_service_information` as it returns tuple and not a dict. Meaning that `_check_consistency` check would fail asserting that value is not a dict. 

https://github.com/valory-xyz/open-autonomy/blob/98daea90362d2717840002dea0229466494256d6/packages/valory/contracts/service_registry/contract.py#L205-L218

DavidMinarsch:
## Proposed changes

  Summary

  - get_service_information returned a raw tuple from the contract call, making it incompatible with get_contract_api_response which expects a
  dict via _check_consistency
  - Updated the method to return a JSONLike dict with named keys: security_deposit, multisig_address, config_hash, threshold,
  max_num_agent_instances, num_agent_instances, service_state, canonical_agents
  - Updated all callers across autonomy/chain/service.py, autonomy/analyse/service.py, autonomy/cli/helpers/chain.py, and
  autonomy/cli/helpers/deployment.py from tuple unpacking to dict key access

  Test plan

  - Verify contract tests pass (packages/valory/contracts/service_registry/tests/test_contract.py)
  - Verify mypy, pylint, black, flake8 all pass
  - Verify on-chain service operations (activate, register, deploy, terminate, unbond) work correctly with dict return type

## Fixes

#2330

## Types of changes

## Checklist

## Further comments

-


DavidMinarsch:
Closing this PR — after reviewing, I don't think the effort is worth the cost.

**The stated motivation doesn't hold up.** The commit message says the tuple return was "incompatible with `get_contract_api_response` which expects a dict." But grepping the entire repo, `get_service_information` has exactly one caller (`autonomy/chain/service.py`), which invokes it as a direct Python method, not via `get_contract_api_response`. No skill uses `contract_callable="get_service_information"`. The incompatibility this fix addresses is hypothetical — nothing in the codebase needed dict semantics.

**What the change delivers:** named-key access (`info["multisig_address"]`) instead of positional unpacking — a real readability/robustness win, and preparatory work if a future skill wants to call this via the contract API from a behaviour.

**What it costs:**
- Breaking change to an internal helper's return type.
- Churn across `autonomy/*` files and hash propagation through ~25 packages.
- Incomplete migration — several test callers still use tuple unpacking or mock tuples and are silently broken:
  - `tests/test_autonomy/test_cli/test_service/test_service_manager.py:814,829` — `_, multisig_address, *_ = get_service_info(...)`; iterating a dict yields keys, so `multisig_address` becomes the string `"multisig_address"`.
  - `tests/test_autonomy/test_cli/test_mint/test_mint_components.py:224,244` — same pattern with `threshold`; `assert threshold == 3` will fail.
  - `tests/test_autonomy/test_cli/test_helpers/test_chain_helpers.py:415,443` — `mock.patch(..., return_value=(1, None, state.value, None))`; production now does `info["service_state"]` → `TypeError` before the expected `ClickException`.
  - `tests/test_autonomy/test_cli/test_analyse/test_verify_service.py:827,915` — mocks return `[None, 0, None]` / `(None, 4, None)`; `autonomy/analyse/service.py` now does `info["service_state"]` → `TypeError`.
- Stale docstrings in `autonomy/chain/service.py:143-145` and `358-366` still describe the old tuple ordering.

**Verdict:** You don't do breaking API surgery on an internal helper for ergonomics alone — you do it when there's a concrete consumer. We can revisit when an actual skill needs to call this via `get_contract_api_response`; at that point the change is justified by a real requirement and the caller migration should be done together with it.

PR #2469, opened 2026-04-14
Deprecate tm for single-agent services

Adamantios:


DavidMinarsch:
## Review: comparison with #2104 (mock Tendermint at connection layer)

Both PRs solve the same problem — eliminating Tendermint overhead for single-agent services — but take different routes.

### This PR's approach (skill-layer bypass)

Bypasses TM entirely inside the skill: `send_a2a_transaction` calls `process_payload()` directly (no signing, no HTTP, no envelopes), and the behaviour loop drives `RoundSequence` via synthetic `begin_block/end_block/commit` calls.

**Strengths:** smaller diff (108 lines), faster execution (no network), no new infrastructure.

### Concerns

1. **Forks the skill code path permanently.** Every hot path now has `if no_tm:` branching (`send_a2a_transaction`, `async_act_wrapper`, `syncing_up`, `block_stall_deadline_expired`, `reset_pause_abci`). Any future change to the behaviour loop or transaction flow must be tested against both paths — ongoing maintenance tax.

2. **Skips transaction signing and verification.** `process_payload(payload)` bypasses `Transaction.encode()` → sign → verify. Signature validation bugs won't surface in single-agent mode. The Transaction class is completely untested in this path.

3. **Implicit flag.** `no_tm` is derived from `max_participants == 1` at runtime. Can't test a multi-agent skill in single-agent mode (for dev), can't run single-agent *with* TM (for debugging). Not configurable.

4. **`__async_act_no_tm` re-implements generator delegation.** Manually wraps `async_act()` with `next()/send()/throw()` — fragile if the generator protocol changes.

5. **No tests.**

### #2104 (connection-layer mock) — now complete

Mocks Tendermint as a black box at the connection layer. The skill code is completely untouched — the agent doesn't know it's running against a mock. Full ABCI envelope round-trip, transaction signing, HTTP submission all exercised normally.

- Zero new dependencies (stdlib `asyncio.start_server`)
- 16 unit tests + 2 e2e tests
- `UseMockTendermint` mixin for adding mock variants of any test
- All linters pass, all review comments addressed

### Recommendation

Closing in favour of #2104 which preserves protocol fidelity — any bug that would surface with real Tendermint also surfaces with the mock. This PR's approach could be revisited as an optional "fast mode" for local development if the ~1s block time proves too slow for iteration.

DavidMinarsch:
Closing in favour of #2104 which takes the connection-layer mock approach — preserves full protocol fidelity with no skill code changes. See review comment above for detailed comparison.

PR #2470, opened 2026-04-14
chore: migrate scripts to CLIs + consume open-aea 2.2.1 + remove flashbots plugin

DhairyaPatel7:
## Summary

This PR started as a `scripts/` → CLI migration and grew through review into a full dependency refresh to `open-aea 2.2.1`.

> **For re-reviewers:** skip `docs/api/**` — ~14.5k of additions are regenerated API markdown. Real code/CI changes are in `autonomy/`, `packages/valory/*/{*.yaml,*.py}`, `plugins/aea-helpers/`, `tox.ini`, `.github/workflows/`, `setup.py`, `pyproject.toml`, `deployments/Dockerfiles/`, `docs/*.md` (non-api).

### Four threads

1. **Scripts → CLIs.** Every Python script in `scripts/` is now a CLI command. `scripts/` contains only data files, git hooks, and docs.
2. **`open-aea` 2.1.0 → 2.2.1.** Bumps across `pyproject.toml`, `tox.ini`, `setup.py`, `poetry.lock`, inter-plugin pins, and every `open-aea*` version pin in `packages/valory/contracts/*/contract.yaml` and `packages/valory/agents/*/aea-config.yaml` (23 yamls). `autonomy packages lock` cascades the fingerprint changes through every downstream package.
3. **Upstream consolidation.** `check_third_party_hashes` and `generate_api_docs` were promoted from OA's `aea-helpers` to upstream `open-aea-ci-helpers` (valory-xyz/open-aea#876). This PR deletes the OA-local forks and wires `tox.ini` to call `aea-ci check-third-party-hashes --upstream valory-xyz/open-aea@2.2.1` and `aea-ci generate-api-docs --source-dir autonomy ...` directly.
4. **Flashbots plugin cleanup (minimal surface, intentional).** `open-aea-ledger-ethereum-flashbots` was removed upstream in 2.2.1 (valory-xyz/open-aea#877). This PR cleans up all dead-on-arrival call sites but **intentionally leaves the skill-level API intact.** See call-out below.

### Scripts migrated

**Tomte (0.6.5):**
- `check_copyright.py` → `tomte check-copyright --author "Valory AG"`
- `check_doc_links.py` → `tomte check-doc-links` (with URL skip flags)
- `freeze_dependencies.py` → `tomte freeze-dependencies --exclude-package open-autonomy`

**Upstream `aea-ci-helpers` (2.2.1):**
- `generate_package_list.py` → `aea-ci generate-pkg-list`
- `check_ipfs_hashes_pushed.py` → `aea-ci check-ipfs-pushed`
- `check_third_party_hashes.py` → `aea-ci check-third-party-hashes --upstream valory-xyz/open-aea@2.2.1`
- `generate_api_documentation.py` → `aea-ci generate-api-docs --source-dir autonomy --packages-dir packages --plugins-dir plugins --parallel --default-package ...`

**OA-local `aea-helpers`:**
- `generate_contract_list.py` → `aea-helpers generate-contract-list` (Olas registry — OA-specific)

### Flashbots cleanup (call-out)

Upstream `open-aea-ledger-ethereum-flashbots` is unmaintained and no longer shipped in open-aea 2.2.1. The 2.2.0 wheel remains frozen on PyPI.

**What this PR changes (dead-on-arrival call sites):**
- `autonomy/replay/agent.py` — dropped `cp ethereum_private_key.txt → ethereum_flashbots_private_key.txt` and `aea add-key ethereum-flashbots`.
- `deployments/Dockerfiles/development/run.sh` — dropped flashbots key-copy + `aea add-key ethereum_flashbots` (the container would have failed at entrypoint otherwise).
- `deployments/Dockerfiles/autonomy/scripts/start.sh` — removed `handleFlashbotsKey()` helper and its call site (dead code).
- `plugins/aea-helpers/aea_helpers/check_dependencies.py` — dropped `open-aea-flashbots` from Pipfile ignore list.
- `plugins/aea-helpers/aea_helpers/bump_dependencies.py` — dropped the `open-aea-ledger-ethereum-flashbots` version-bump entry.
- `docs/advanced_reference/commands/autonomy_build-image.md` — replaced stale `open-aea-ledger-flashbots==2.1.0` example with `open-aea-ledger-solana==2.2.1`.

**What this PR INTENTIONALLY DOES NOT CHANGE — skill-level `use_flashbots` API stays intact:**

`packages/valory/skills/*` is not touched. Specifically still present:
- `abstract_round_abci/behaviour_utils.py` — `FLASHBOTS_LEDGER_ID = "ethereum_flashbots"` constant, `use_flashbots` parameter on `send_raw_transaction()` / `_send_transaction_request()`, and the `if use_flashbots:` dispatch branch.
- `transaction_settlement_abci/behaviours.py` — `self.use_flashbots`, `_get_tx_data(use_flashbots=...)` parameter, payload plumbing.
- `transaction_settlement_abci/payload_tools.py` — `use_flashbots` + `raise_on_failed_simulation` 32-byte fields in the on-chain payload bytes. Removing them would shift offsets by 64 bytes and silently corrupt every in-flight transaction exchanged with agents still on 2.2.0.

**Runtime behaviour:**
- Agents with `use_flashbots=False` (the default): run **unchanged**.
- Agents with `use_flashbots=True`: fail at dispatch because the `ethereum_flashbots` ledger is no longer registered. This is the correct signal — those agents need to either (a) pin `open-aea<=2.2.0` + the frozen `open-aea-ledger-ethereum-flashbots==2.2.0` wheel, or (b) bundle flashbots support in their own fork/plugin.

A follow-up PR can fully remove the skill-level code paths and bump the payload wire format if/when the Valory agent portfolio is confirmed to have migrated off it. Out of scope here.

### Version bumps — consistency

All `open-aea*==2.1.0` pins in 23 package yamls bumped to `==2.2.1`:
- 12 `packages/valory/contracts/*/contract.yaml`
- 11 `packages/valory/agents/*/aea-config.yaml`

These are AEA package-frozen dep declarations. Bumping them cascades through `autonomy packages lock` to every package whose dep graph includes those yamls.

### Third-party hash realignment

Upstream open-aea 2.2.1 regenerated 7 third-party fingerprints (connection.yaml changes from the flashbots removal). `packages/packages.json` `third_party` block updated, packages re-synced via `autonomy packages sync --update-packages`, downstream package hashes re-locked.

`aea-ci check-third-party-hashes --upstream valory-xyz/open-aea@2.2.1` reports: *"All 8 third-party hashes are consistent with upstream."*

### Deferred to a follow-up release PR

- **OA version bump `0.21.16 → 0.21.17`** (and the attendant `packages lock` cascade).
- **`docs/upgrading.md` entry** covering the open-aea 2.1.0 → 2.2.1 jump, the flashbots plugin removal, the replay-flow behavior change, and the migration steps for downstream consumers.

The content of that upgrading note is drafted in the thread replies on this PR so reviewers have it on record, but it lives in the release PR.

## Test plan

Local runs on all CI envs — all green:

- [x] `tox -e black-check -e isort-check -e flake8`
- [x] `tox -e check-copyright`
- [x] `tox -e check-hash`
- [x] `tox -e check-packages`
- [x] `tox -e check-third-party-hashes` — *"All 8 third-party hashes are consistent with 1 reachable upstream(s)"*
- [x] `tox -e check-api-docs`
- [x] `tox -e check-dependencies`
- [x] `tox -e generate-api-documentation` — clean, output committed
- [x] `poetry run make generators` — abci-docstrings + fix-copyright + lock-packages + generate-api-documentation + fix-doc-hashes all pass
- [x] `poetry check --lock` — exit 0
- [x] `tomte format-copyright --author "Valory AG" --author "Fetch.AI Limited"` — only touches the 6 dummy test fixture files (matches main's CI behaviour)
- [ ] CI matrix (GitHub Actions) — Python 3.10-3.13 exercised; 3.14 declared-only per `pyproject.toml <3.15`

## Review addressed

All comments from @DavidMinarsch, @LOCKhart07, @OjusWiZard, and Copilot have been addressed in code or replied with rationale in the respective threads. One non-blocking typo in an auto-generated API doc is noted as a follow-up (upstream docstring fix, not in this PR's diff).


tox.ini
```diff
@@ -422,14 +422,15 @@ commands =
 usedevelop = True
 deps =
     {[deps-packages]deps}
+    ; requires open-aea-ci-helpers >= 2.2.0
+    open-aea-ci-helpers
```
OjusWiZard: would be good to pin them, so that we don't surprise ourselves in the future.

OjusWiZard (dismissed):

tox.ini
```diff
@@ -445,63 +446,67 @@ commands =
...
     {[deps-packages]deps}
+    tomte[cli]==0.6.5
+    ; requires open-aea-ci-helpers >= 2.2.0
+    open-aea-ci-helpers
 commands =
     python -m pip install --no-deps {toxinidir}{/}plugins{/}aea-helpers
     aea init --reset --author ci --remote --ipfs --ipfs-node "/dns/registry.autonolas.tech/tcp/443/https"
     aea packages sync
-    {toxinidir}/scripts/check_doc_links.py
+    tomte check-doc-links -n "http://pmg.csail.mit.edu/papers/osdi99.pdf" -u "https://github.com/valory-xyz/open-operator" -u "https://gateway.autonolas.tech/ipfs/<hash>"
```
LOCKhart07: **Blocking.** This drops 3 URL skips that the deleted `scripts/check_doc_links.py` had, and they all still appear in current docs:

- `http://host.docker.internal:8545` — `docs/guides/define_service.md`, `docs/configure_service/configure_access_external_chains.md`, `docs/configure_service/service_configuration_file.md`, `docs/advanced_reference/developer_tooling/debugging_in_the_cluster.md`
- `http://www.fipa.org/repository/ips.php3` — `docs/key_concepts/abci_app_async_behaviour.md`
- `https://github.com/valory-xyz/open-autonomy/trunk/infrastructure` — `docs/using_stack_deployment.md`, `docs/application_deployment.md`

`tox -e check-doc-links-hashes` is also missing from the PR test plan, so this likely wasn't exercised. Add the missing entries (`-n` for HTTP-allowlist, `-u` for status-code skips) before merge.

tox.ini
```diff
@@ -378,23 +378,23 @@ commands =
...
     isort --check-only --gitignore autonomy packages/valory scripts tests plugins
 
 [testenv:check-copyright]
 skipsdist = True
 skip_install = True
 deps =
-commands = {toxinidir}/scripts/check_copyright.py --check
-allowlist_externals = {toxinidir}/scripts/check_copyright.py
+    tomte[cli]==0.6.5
+commands = tomte check-copyright --author "Valory AG" --scan-path autonomy --scan-path tests --scan-path packages/valory/agents --scan-path packages/valory/connections/abci --scan-path packages/valory/contracts --scan-path packages/valory/skills --scan-path scripts --scan-path plugins -e protocols -e dummy_packages
```
LOCKhart07, line 390: Non-blocking. The deleted `scripts/check_copyright.py` excluded `_pb2.py`, `_pb2_grpc.py`, `build/`, `t_protocol/`, `t_protocol_no_ct/`. The new invocation only excludes `protocols` and `dummy_packages`, but generated `_pb2.py` files exist outside `packages/valory/protocols/` (e.g. `packages/valory/connections/abci/gogoproto/gogo_pb2.py`, which has no Valory header). This only passes if tomte's `check-copyright` auto-skips generated files — worth confirming explicitly, otherwise add the exclusions.

plugins/aea-helpers/aea_helpers/check_third_party_hashes.py
```diff
@@ -105,5 +104,8 @@ def main() -> None:
     )
 
 
-if __name__ == "__main__":
-    main()
+@click.command(name="check-third-party-hashes")
+def check_third_party_hashes() -> None:
+    """Check that third-party package hashes match the open-aea repository."""
+    root_dir = Path.cwd()
```
LOCKhart07: Non-blocking. Hardcoding `Path.cwd()` as `root_dir` makes the contract implicit — the command silently does the wrong thing if invoked from outside the OA repo root. Expose it as a click option (`--root-dir`, default `.`) so the dependency on cwd is documented.

plugins/aea-helpers/setup.py
```diff
@@ -32,7 +32,7 @@
 ]
 
 setup(
-    name="aea-helpers",
+    name="open-aea-helpers",
```
LOCKhart07, line 35: Non-blocking but worth calling out in the PR body / release notes: `open-aea-helpers` is not yet on PyPI (verified — `pip index versions open-aea-helpers` returns no matches), so the next release will be the first publish under the new name. Anyone with `aea-helpers` pinned externally will break, and the old PyPI name becomes orphaned (typosquat surface). Confirm someone will claim/reserve it.

plugins/aea-helpers/aea_helpers/check_third_party_hashes.py
```diff
@@ -25,28 +24,27 @@
 import sys
 from pathlib import Path
 
+import click
```
DavidMinarsch: why isn't this in open-aea? You can imagine agents building purely on open-aea wanting this

plugins/aea-helpers/aea_helpers/cli.py
```diff
@@ -24,7 +24,10 @@
 from aea_helpers.check_binary import check_binary
```
DavidMinarsch, line 24: So we're extending aea-helpers?

plugins/aea-helpers/aea_helpers/generate_api_docs.py
```diff
@@ -234,18 +234,22 @@ def install(package: str) -> int:
...
 
 
-if __name__ == "__main__":
-    parser = argparse.ArgumentParser("generate_api_docs")
-    parser.add_argument(
-        "--check-clean", action="store_true", help="Check if the working tree is clean."
-    )
-    arguments = parser.parse_args()
-
+def main(check_clean: bool = False) -> None:
```
DavidMinarsch: Why wasn't this backported to open-aea?

plugins/aea-helpers/setup.py
DavidMinarsch, line 35: why aea-helpers in autonomy?

pyproject.toml
```diff
@@ -7,15 +7,15 @@ license = "Apache-2.0 license"
 packages = [{include = "autonomy"}, {include = "packages"}]
```
DavidMinarsch, line 8: why are open-aea-ci-helpers and open-aea-dev-helpers not used here?

LOCKhart07 (dismissed):
Re-reviewed commits `aebd768` and `eff13d0b` since my previous review.

**Prior items — resolved:**
- Missing URL skips (`host.docker.internal`, `fipa.org`, `open-autonomy/trunk/infrastructure`) added to `check-doc-links-hashes` (blocker).
- `--root-dir` click option added to `check-third-party-hashes`.
- `open-aea-ci-helpers` now pinned to `==2.2.0`.

**Minor, non-blocking polish (optional):**
- `tox.ini` `check-third-party-hashes` still installs `aea-helpers` via `deps = {toxinidir}/plugins/aea-helpers`, while the other new envs (`check-api-docs`, `generate-api-documentation`, `check-doc-links-hashes`, `lock-packages`) switched to `python -m pip install --no-deps ...` in `commands` to avoid the circular `open-autonomy` resolution. Lower risk here because this env omits `{[deps-packages]deps}`, but aligning for consistency would be nice.
- `http://host.docker.internal:8545` appears in both `-n` and `-u` for `tomte check-doc-links` — intentional (non-HTTPS skip + unreachable status skip), just worth noting so it isn't "simplified" later.
- Still outstanding from prior pass (non-blocking): generated `_pb2.py` files outside `packages/valory/protocols/` and the `open-aea-helpers` PyPI name reservation.

LGTM.

plugins/aea-helpers/aea_helpers/check_third_party_hashes.py
DhairyaPatel7: This script checks third-party packages against the open-aea repo's hashes. We put it in OA's aea-helpers initially, but any downstream repo building on open-aea would want the same check. Should I migrate it to `open-aea-ci-helpers`?

plugins/aea-helpers/aea_helpers/cli.py
DhairyaPatel7: Yes — `aea-helpers` already existed in OA with commands like `run-agent`, `run-service`, `make-release`, `config-replace`, `check-binary`, etc. We added 3 commands that were previously in `scripts/`: `check-third-party-hashes`, `generate-contract-list`, and `generate-api-docs`. The first two are arguably OA-specific (contract list generation, third-party hash checking against open-aea). The API doc generator is OA-specific because it scans `autonomy/` rather than `aea/` — but per your other comment, that could be made configurable upstream instead.

plugins/aea-helpers/aea_helpers/generate_api_docs.py
DhairyaPatel7: The open-aea `aea-ci generate-api-docs` command is hardcoded to scan `aea/` and `plugins/` with open-aea's directory structure. OA's version scans `autonomy/` and `packages/valory/`. We couldn't reuse the open-aea version as-is, so we kept OA's version in aea-helpers.

The right fix would be to make `aea-ci generate-api-docs` configurable (e.g. `--source-dir`, `--packages-dir` flags) so both repos can use the same command. Should I create a follow-up PR on open-aea for this?

plugins/aea-helpers/setup.py
DhairyaPatel7: The other OA plugin follows the same pattern: `open-aea-test-autonomy` — also an `open-aea-*` name living in the OA repo. We're renaming this one from `aea-helpers` to `open-aea-helpers` on PyPI in this PR to match that convention. The commands it provides (`run-agent`, `run-service`, `make-release`, `config-replace`, `check-binary`) are OA-specific and depend on `autonomy`.

pyproject.toml
DhairyaPatel7: They're not runtime dependencies of open-autonomy — they're CI/dev tools. They're declared in tox.ini as deps for the specific tox envs that need them (e.g. `open-aea-ci-helpers==2.2.0` in `check-doc-links-hashes`). Adding them to pyproject.toml would make them install-time dependencies for every user of open-autonomy, which isn't right since most users don't need CI tooling.

Should we add them under `[tool.poetry.group.dev.dependencies]` instead so they're available during development but not for end users?

DhairyaPatel7:
Thanks @LOCKhart07. Aligned `check-third-party-hashes` with the other new envs in 26d57a6d1 — now uses `pip install --no-deps` in `commands` for consistency.

Leaving the `host.docker.internal:8545` entries in both `-n` and `-u` as-is per your note (intentional: non-HTTPS skip + unreachable status skip).

plugins/aea-helpers/aea_helpers/check_third_party_hashes.py
DavidMinarsch: yes please - migrate there and re-release for use here

plugins/aea-helpers/aea_helpers/cli.py
DavidMinarsch: third-party hash checking should be generic against any upstream repo (or list of repos)

DavidMinarsch: let's make api thing configurable upstream so we don't need to maintain two versions

DavidMinarsch: generate contract list can stay here. but what is it's purpose?

plugins/aea-helpers/aea_helpers/generate_api_docs.py
DavidMinarsch: yes

plugins/aea-helpers/setup.py
DavidMinarsch: ok

pyproject.toml
DavidMinarsch: yes dev dependencies

packages/packages.json
```diff
@@ -27,39 +27,39 @@
...
+        "agent/valory/registration_start_up/0.1.0": "bafybeidulmrwjvdnwaigkq4cwf2ttqyqibiju7v2r55fuqbwoo4rqcew2i",
+        "agent/valory/test_abci/0.1.0": "bafybeigesexiey6v3zrhub3hbhouxrqk53qrnkzychzvxizdedy4v7jg5u",
+        "agent/valory/register_reset_recovery/0.1.0": "bafybeieplaug7p2umzkfgdzokeowngs3jmxwwprqtxd4gzrqplfft5uuxe",
+        "agent/valory/offend_slash/0.1.0": "bafybeifzh4lia6yyyisrfwmzyxqtwno724bm7ggwqr6dbkaeko3pklddsu",
+        "agent/valory/solana_transfer_agent/0.1.0": "bafybeicdnndgujqxy5ou4iwqc7et5cjwua2dnnju6g2oi34qjxddgeviam",
         "service/valory/counter/0.1.0": "bafybeid5cbvapyjldzkfzetxziilrj3hr3cd7zqh35fzx4ggzs74crlac4",
-        "service/valory/register_reset/0.1.0": "bafybeiehewfjf4i3zvqj6utinvpymdimnfy46lddhmehsqgojr3nw25nsm"
+        "service/valory/register_reset/0.1.0": "bafybeieieajw6xbliw5xwxgppyidgclmwrjook2sw2dpsvzxw543tkmafa"
     },
     "third_party": {
```
DavidMinarsch, line 64: should there not have been more changes?

plugins/aea-helpers/aea_helpers/check_third_party_hashes.py
DhairyaPatel7: Closed — open-aea 2.2.1 is out, and this OA PR now consumes `aea-ci check-third-party-hashes` directly. The OA-local fork at `plugins/aea-helpers/aea_helpers/check_third_party_hashes.py` is deleted in this PR; the tox `check-third-party-hashes` env now runs:

```
aea-ci check-third-party-hashes --upstream valory-xyz/open-aea@2.2.1
```

Upstream command supports repeatable `--upstream owner/repo@version` for multiple upstreams and per-upstream fault tolerance.

plugins/aea-helpers/aea_helpers/cli.py
DhairyaPatel7: `generate-contract-list` fetches the Olas registry addresses JSON from [`valory-xyz/autonolas-registries`](https://raw.githubusercontent.com/valory-xyz/autonolas-registries/main/docs/configuration.json) and regenerates `docs/advanced_reference/on_chain_addresses.md` — a per-chain markdown table of canonical Olas contract addresses (ComponentRegistry, ServiceRegistry, Gnosis Safe multisigs, Multisend) with block-explorer links.

It's OA-specific to the Olas deployment (not useful for other open-aea consumers), so agreed — stays in `aea-helpers`.

DhairyaPatel7: Also closed on this thread:

- **Third-party-hash genericness** — upstream `aea-ci check-third-party-hashes` now accepts repeatable `--upstream owner/repo@version` and tolerates per-upstream flakiness. OA is on it in this PR.
- **API-docs configurable upstream** — done in open-aea#876 via `ApiDocsConfig` (`--source-dir`, `--packages-dir`, `--plugins-dir`, `--docs-dir`, `--default-package`, `--ignore-plugin`, `--ignore-prefix`, `--parallel`). OA's tox envs now call `aea-ci generate-api-docs` with the appropriate flags — see `[testenv:check-api-docs]` and `[testenv:generate-api-documentation]` in this PR's `tox.ini`.

plugins/aea-helpers/aea_helpers/generate_api_docs.py
DhairyaPatel7: Closed — done in open-aea#876. `aea-ci generate-api-docs` now accepts `--source-dir`, `--packages-dir`, `--plugins-dir`, `--docs-dir`, `--default-package` (repeatable), `--ignore-plugin`, `--ignore-prefix`, `--parallel` via an `ApiDocsConfig` dataclass. Released in open-aea 2.2.1.

This PR deletes OA's fork at `plugins/aea-helpers/aea_helpers/generate_api_docs.py` and wires the tox envs to call `aea-ci generate-api-docs` directly with OA-specific flags (`--source-dir autonomy`, `--packages-dir packages`, `--plugins-dir plugins`, 8 `--default-package` entries for the canonical OA packages, `--parallel`).

autonomy/chain/metadata.py
```diff
@@ -82,7 +82,7 @@ def publish_metadata(
         nft_image_hash=nft_image_hash,
     )
 
-    metadata_hash = ipfs_tool.client.add_str(metadata_string)
+    metadata_hash = ipfs_tool.client.add_bytes(metadata_string.encode())
```
LOCKhart07, line 85: Non-blocking. This is the on-chain metadata publish path. CIDs are deterministic over byte content, so UTF-8-encoded bytes should yield the same CID as the previous `add_str` (which also UTF-8-encoded internally). Worth a one-line note in the PR body (or commit message) confirming a spot-check that a known metadata string yields the same CID pre/post — a silent CID drift here would point minted services at the wrong IPFS object.

pyproject.toml
```diff
@@ -7,15 +7,15 @@ license = "Apache-2.0 license"
...
 
 [tool.poetry.dependencies]
-python = "^3.10"
+python = ">=3.10,<3.15"
```
LOCKhart07, line 10: Non-blocking. Range bumps to `<3.15` (declares 3.14 support), but the test plan lists only local tox runs and there's no indication 3.14 is actually exercised. Worth calling out in the description that CI will cover 3.10–3.13 and 3.14 is declared-only — or flag it as follow-up so consumers don't assume validated support.

Separately (meta): the PR title `feat: migrate all scripts to plugins and tomte CLI` no longer reflects scope — this is also an open-aea 2.2.1 bump + flashbots cleanup + upstream consolidation. Consider retitling. And the ~14.5k additions (mostly regenerated API docs) make re-review hard — worth telling reviewers to skip `docs/api/**`.

tox.ini
```diff
@@ -445,63 +446,77 @@ commands =
 skipsdist = True
 usedevelop = True
 deps =
-    requests
+    pip
+    open-aea-ci-helpers==2.2.1
```
LOCKhart07, line 453: Non-blocking nit. `deps = pip` (line 449) is redundant since pip is always available in a tox env, and `usedevelop = True` on this env is pointless because `aea-ci check-third-party-hashes` does not import the local package. Drop both to align with the other pure-`aea-ci` envs.

autonomy/deploy/generators/kubernetes/templates.py
```diff
@@ -1,7 +1,7 @@
 # -*- coding: utf-8 -*-
 # ------------------------------------------------------------------------------
 #
-#   Copyright 2021-2024 Valory AG
+#   Copyright 2021-2026 Valory AG
```
LOCKhart07, line 4: Non-blocking. Only a copyright year bump (2024→2026), presumably collateral from `tomte format-copyright`. Unrelated to the stated PR intent — a future reader will wonder why this file is in the diff.

tox.ini
LOCKhart07, line 390: Re-surfacing @LOCKhart07's still-open thread on exclusions: this invocation only excludes `protocols` and `dummy_packages`, but generated `_pb2.py` / `_pb2_grpc.py` files exist outside `packages/valory/protocols/` (e.g. `packages/valory/connections/abci/tendermint/abci/types_pb2_grpc.py`, `gogoproto/gogo_pb2.py`). A one-line confirmation that tomte's `check-copyright` auto-skips generated files would close the loop; otherwise the exclusions list needs the extra entries the deleted `scripts/check_copyright.py` had.

DavidMinarsch:
Did a careful consistency pass on the dep bumps and flashbots cleanup. Found a few real issues — two of them potential blockers.

## 🔴 Issues introduced by this PR

### 1. `setup.py` is pinned to `==2.2.0` while everything else is `==2.2.1`

```
setup.py:33    "open-aea-cli-ipfs==2.2.0",    ← stale
setup.py:54    "open-aea[all]==2.2.0",        ← stale
```

Meanwhile `pyproject.toml`, `tox.ini`, and `poetry.lock` are uniformly on `==2.2.1`. The diff against main shows the intent was to bump `setup.py` too — it went `2.1.0 → 2.2.0`, just stopped one version short.

This matters because `Makefile:140-142` still builds via `python setup.py sdist/bdist_wheel`, and `release.yml:34` invokes it via `poetry run make dist`. So the PyPI-published sdist/wheel would ship with `open-aea==2.2.0` constraints while the lockfile and CI run on `2.2.1`. The `tox -e check-dependencies` safety net doesn't catch this because `plugins/aea-helpers/aea_helpers/check_dependencies.py` only validates `pyproject.toml`/`Pipfile`/`tox.ini` — there's no `SetupPyConfig`.

Two fixes:
- **Minimum**: bump `setup.py` pins to `==2.2.1` to match.
- **Better**: rewrite `make dist` to use `poetry build`, so `setup.py` stops being release-authoritative. You're mid-Poetry-migration — this is a natural place to finish it.

### 2. Flashbots cleanup is incomplete — runtime break in the development image

The PR correctly cleaned up `autonomy/replay/agent.py` (removed the `ethereum_flashbots` `add-key` subprocess calls). But two deployment scripts still reference the removed plugin:

- **`deployments/Dockerfiles/development/run.sh:13,19`** — this is the bad one:
    ```bash
    cp /agent_key/ethereum_private_key.txt ethereum_flashbots_private_key.txt
    …
    aea add-key ethereum_flashbots
    ```
    `aea add-key ethereum_flashbots` will fail at container entrypoint because the ledger plugin is gone in 2.2.1. **The development Docker image will be broken after this PR merges.**

- **`deployments/Dockerfiles/autonomy/scripts/start.sh:42-47`** — `handleFlashbotsKey()` greps for `open-aea-ledger-ethereum-flashbots` in `aea-config.yaml`. Dead code now (the grep won't find anything), but inconsistent with the `replay/agent.py` cleanup.

The PR body is explicit that the *skill-level* flashbots API (`behaviour_utils.py`, `connection.yaml`) is intentionally preserved to avoid a wire-format break. That's fine — but `start.sh` and `run.sh` are runner-level, not skill-level. Either clean all three or none, and `development/run.sh` really must be cleaned because it actively breaks the container at startup.

### 3. No `docs/upgrading.md` entry for this dependency jump

Convention from commit `275ec2795` is to add a `## vX.Y.Z to vX.Y.Z+1` block documenting "Main dependency updates to account for" whenever `open-aea` or `tomte` moves. This PR bumps:

- `open-aea: ==2.1.0 → ==2.2.1` (the earlier `2.1.0 → 2.2.0` jump also missed the note, so this would catch up)
- `tomte: ==0.6.1 → ==0.6.5`

More importantly, **`open-aea-ledger-ethereum-flashbots` removal is a breaking change for downstream users**. Even with the skill-level wire format preserved, any agent config that references the plugin in `aea-config.yaml` or installs it as an extra dep will break. That warrants an explicit upgrading note so downstream repos (trader, mech, etc.) know to audit their configs.

### 4. Stale flashbots example in `docs/advanced_reference/commands/autonomy_build-image.md:87`

```bash
autonomy build-image ... -e open-aea-ledger-flashbots==2.1.0
```

Plugin no longer exists upstream. Minor, but since you're already touching flashbots-related cleanup, it's cheap to fix.

### 5. CI is red — `check-api-docs` failing

`aea-ci generate-api-docs --check` is exiting with code 1 on `ubuntu-latest, 3.14`. The log shows it processes everything and then fails the check — meaning at least one `.md` under `docs/api/` is out of sync with source. The PR did regenerate a bunch (9 new `plugins/aea_helpers/*.md` files) but something is still drifting. Fix: run `tox -e generate-api-documentation` locally and commit whatever delta falls out. This is blocking merge.

## 🟡 Pre-existing issue worth fixing while you're here

### 6. `.github/workflows/main_workflow.yml:94` calls a non-existent tox env

```yaml
- name: Check doc IPFS hashes
  run: tox -e check-doc-hashes
```

There is no `[testenv:check-doc-hashes]` in `tox.ini`. The actual env is `check-doc-links-hashes` (line 478). This bug predates this PR — it's already on main — but it's currently masked because the preceding `check-api-docs` step fails first, so CI never gets here. **Once you fix #5, this will start failing.** Best fixed in this PR since you're already touching the doc-hash pipeline. Either rename the workflow step to `tox -e check-doc-links-hashes` or add an alias env.

## 🟢 Consistency checks that passed

- `pyproject.toml`, `tox.ini`, `poetry.lock` are uniformly on `open-aea==2.2.1` (and ci-helpers, dev-helpers, ledger variants to match). No stragglers.
- `tomte` is uniformly `0.6.5` across `tox.ini` and `.github/workflows/main_workflow.yml`.
- Old `scripts/*.py` (`check_copyright`, `check_doc_links`, `check_ipfs_hashes_pushed`, `check_third_party_hashes`, `freeze_dependencies`, `generate_api_documentation`, `generate_package_list`, `generate_contract_list`) are all deleted; `scripts/` retains only `__init__.py`, `whitelist.py`, git hooks, and `RELEASE_PROCESS.md`. No dangling references in `Makefile` or workflows.
- `plugins/aea-test-autonomy/setup.py` uses `>=2.2.0,<3.0.0` — loose but accepts 2.2.1 correctly.
- The "aea-helpers" → "open-aea-helpers" PyPI rename is consistent across `plugins/aea-helpers/setup.py`, `pyproject.toml`, and `release.yml`.
- `autonomy/chain/metadata.py` `add_str` → `add_bytes(encode())` is a necessary runtime fix for the ipfshttpclient API change and is correctly applied.

## Ranked action list

1. **Fix `setup.py` pins to `==2.2.1`** — or, better, swap `make dist` to `poetry build`. Blocker.
2. **Clean up `deployments/Dockerfiles/development/run.sh`** — remove flashbots key copy and `aea add-key ethereum_flashbots`. Blocker — runs at container startup.
3. Clean up `handleFlashbotsKey()` in `start.sh` to match. Consistency.
4. Regenerate API docs to unblock CI.
5. Fix `main_workflow.yml:94` tox env name.
6. Add `upgrading.md` entry covering open-aea bump + flashbots plugin removal.
7. Update flashbots example in `autonomy_build-image.md`.

setup.py
```diff
@@ -30,7 +30,7 @@
 def get_all_extras() -> Dict:
     cli_deps = [
         "click>=8.1.0,<9",
-        "open-aea-cli-ipfs==2.1.0",
+        "open-aea-cli-ipfs==2.2.0",
```
OjusWiZard: The PR summary says this branch is moving to `open-aea 2.2.1`, but `setup.py` still pins `open-aea-cli-ipfs==2.2.0` here (and `open-aea[all]==2.2.0` below), while `pyproject.toml`/`tox.ini` use `2.2.1`. Could we align all packaging surfaces to the same exact version? Otherwise installs may behave differently depending on whether consumers come in through Poetry, pip/setuptools, or tox.

setup.py
```diff
@@ -51,7 +51,7 @@ def get_all_extras() -> Dict:
 
 base_deps = [
     "Flask>=3.1.0,<4.0.0",
-    "open-aea[all]==2.1.0",
+    "open-aea[all]==2.2.0",
```
OjusWiZard: Same version-skew concern here: `setup.py` is still on `open-aea[all]==2.2.0`, but the rest of the repo appears to have moved to `2.2.1`. This feels like a release/consumer-footgun for a packaging migration PR.

autonomy/replay/agent.py
```diff
@@ -101,13 +101,6 @@ def start(
...
             [*self.aea_cli, "add-key", "ethereum"], env=self.agent_env
         )
-        subprocess.run(  # nosec # pylint: disable=subprocess-run-check
-            ["cp", "ethereum_private_key.txt", "ethereum_flashbots_private_key.txt"],
-            env=self.agent_env,
-        )
-        subprocess.run(  # nosec # pylint: disable=subprocess-run-check
-            [*self.aea_cli, "add-key", "ethereum-flashbots"], env=self.agent_env
-        )
         subprocess.run(  # nosec # pylint: disable=subprocess-run-check
```
OjusWiZard, line 104: I see from the PR description that removing the flashbots key setup is intentional because the upstream plugin was dropped. That makes sense, but this is still a runtime behavior change for replay flows. Do we have a focused test or explicit release-note coverage for the new failure mode when a workflow expects `ethereum_flashbots` to be available?

docs/api/protocols/abci/tests/test_abci_dialogues.md
```diff
@@ -0,0 +1,32 @@
...
+
+<a id="packages.valory.protocols.abci.tests.test_abci_dialogues.TestDialoguesAbci.make_message_content"></a>
+
+#### make`_`message`_`content
+
+```python
+def make_message_content() -> dict
+```
+
+Make a dict with message contruction content for dialogues.create.
```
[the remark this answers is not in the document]
DhairyaPatel7: Out of scope for this PR — this file is auto-generated by `aea-ci generate-api-docs` from the docstring on `packages/valory/protocols/abci/tests/test_abci_dialogues.py`. The typo is in that Python docstring source. It's not touched by this PR's diff, so fixing it here would churn a package hash for a cosmetic doc fix. Noting as follow-up.

autonomy/chain/metadata.py
DhairyaPatel7: Confirmed — spot-checked locally that `ipfs.client.add_str(s)` and `ipfs.client.add_bytes(s.encode())` yield the same CID for a known metadata string. Both paths UTF-8-encode the input and upload raw bytes, and CIDs are deterministic over content so this is a no-op swap. No silent drift on minted services.

pyproject.toml
DhairyaPatel7: Noted — CI matrix exercises 3.10–3.13. Python 3.14 is declared-only. Will flag it as such in the PR body.

autonomy/deploy/generators/kubernetes/templates.py
DhairyaPatel7: Collateral from merging main in. When `main` merged, this file (which main had at `Copyright 2021-2024`) needed its end year bumped to pass `tomte check-copyright` against the new year. Not intentional scope.

tox.ini
DhairyaPatel7: Verified: `tomte check-copyright` auto-skips `*_pb2.py` and `*_pb2_grpc.py` via its built-in ignore-patterns list (they're always generated-code patterns). Confirmed locally — env passes clean with those files present, no additional exclusion needed. The other patterns the old script had (`build/`, `t_protocol/`, `t_protocol_no_ct/`) don't exist in OA's package tree anyway.

autonomy/replay/agent.py
DhairyaPatel7: Agreed — the replay-flow behavior change deserves an explicit note. Per PR discussion, the `docs/upgrading.md` entry for this open-aea 2.2.1 migration is deferred to a follow-up release PR (where OA version bumps `0.21.16 → 0.21.17`). That PR will specifically call out:

1. `autonomy replay` no longer provisions `ethereum_flashbots_private_key.txt` or registers the `ethereum-flashbots` key. Replays that depended on this will fail at `aea run` with "ledger not registered".
2. Migration: switch to plain `ethereum` in replay fixtures, or pin `open-aea<=2.2.0` + the frozen flashbots plugin.
3. The skill-level `use_flashbots` API is intentionally preserved (default `False`) to avoid a payload wire-format break.

No focused test added for the new failure mode — it's a generic "unknown ledger" AEA error path already covered by AEA's own tests.

DavidMinarsch:
Re-reviewed against `db85aa429`. Every blocker from my round-1 comment is now addressed, plus you root-caused a CI issue I mis-attributed. One new finding below.

## ✅ Round-1 blockers — all resolved

Commit `2b8dfb693` addresses everything:

| # | Finding | Status |
|---|---------|--------|
| 1 | `setup.py` stale `==2.2.0` pins | ✅ Bumped to `==2.2.1` |
| 2 | `development/run.sh` flashbots key ops would break container entrypoint | ✅ Removed |
| 3 | `start.sh` dead `handleFlashbotsKey()` | ✅ Removed along with its call site |
| 4 | `autonomy_build-image.md` stale flashbots example | ✅ Replaced with solana example |
| 5 | CI failing on `check-api-docs` | ✅ Root-caused and fixed (details below) |
| 6 | `main_workflow.yml:94` called non-existent `check-doc-hashes` env | ✅ Renamed to `check-doc-links-hashes` |
| 7 | `upgrading.md` entry missing | 🟡 Explicitly deferred to release PR (reasonable) |

## 🎯 Bonus: root-caused a CI issue I mis-attributed

My round-1 comment said `check-api-docs` was failing because "at least one `.md` under `docs/api/` is out of sync." That was wrong. The actual cause was in the *preceding* workflow step:

```yaml
run: tomte format-copyright --author valory --author fetchai
```

tomte 0.6.5 changed to match author strings **literally** against file headers. Real headers say `Valory AG` / `Fetch.AI Limited`, so the short slugs `valory` / `fetchai` caused tomte to treat every header as "missing the author" and prepend a second (wrong) copyright block. That dirtied the working tree right before `check-api-docs` ran, and the api-docs check caught the resulting diff. Clean root-cause fix in the same commit, with an explanatory comment in the workflow file — nicely done.

## 🎯 Bonus: package-level version consistency

The commit also bumps 23 package YAML files (`open-aea*==2.1.0 → ==2.2.1` across 11 agents and 12 contracts). These are AEA package-frozen dep declarations that cascade through `autonomy packages lock`. Plus third-party hash realignment for 7 upstream protocols/connections. After this, `aea-ci check-third-party-hashes --upstream valory-xyz/open-aea@2.2.1` reports *"All 8 third-party hashes are consistent with upstream."*

## ✅ CI now green

All previously-failing checks are now SUCCESS: `copyright_doc_and_dependencies_check`, `lock_check` on all three platforms, `install_check`, CodeQL, snyk. Only `linter_checks (ubuntu-latest, 3.13)` is still running at the time of this comment.

## 🔴 New finding — dev Docker image is internally inconsistent after this PR

`deployments/Dockerfiles/development/Dockerfile` and `Pipfile` weren't touched by this PR, but they need to be because the PR creates a version mismatch.

**`Dockerfile:1`**
```dockerfile
ARG AEA_VERSION=2.1.0
FROM valory/open-aea-user:${AEA_VERSION}
```

**`Pipfile:16`** (dev-packages)
```toml
open-aea-ledger-ethereum-flashbots = {editable = true, path = "/open-aea/plugins/aea-ledger-ethereum-flashbots"}
```

The `entrypoint.sh` runs `pipenv install --skip-lock --dev`, which installs from the base image's filesystem.

**On main today:** everything is consistent — `pyproject.toml` pins `open-aea==2.1.0`, the dev image bases on `valory/open-aea-user:2.1.0`, and the flashbots plugin exists at `/open-aea/plugins/aea-ledger-ethereum-flashbots` inside that base image.

**After this PR:** `pyproject.toml` pins `open-aea==2.2.1`, but the dev image still bases on 2.1.0. The `valory/open-autonomy-dev` image (built and pushed on every release per `release.yml:212-213`) will ship with:
- open-aea 2.1.0 preinstalled in the base image
- an editable flashbots install from 2.1.0 paths
- open-autonomy 0.21.17 layered on top, which depends on open-aea 2.2.1

At `pipenv install` time this will either hit a dep conflict (framework wants 2.2.1, base has 2.1.0 editable overriding it) or silently resolve to 2.1.0 and ship a dev image where open-autonomy thinks it has 2.2.1 APIs but actually doesn't. Either way, the published `valory/open-autonomy-dev:latest` on the next release will be broken or version-mismatched.

**Fix:**
1. `Dockerfile:1` — bump `ARG AEA_VERSION=2.1.0` → `2.2.1` (assuming `valory/open-aea-user:2.2.1` was published alongside the upstream 2.2.1 release, which it presumably was).
2. `Pipfile:16` — remove the `open-aea-ledger-ethereum-flashbots` editable entry. That path no longer exists in a 2.2.1 base image.
3. *Not a blocker but while you're there:* `Pipfile:8-9` has very stale pins (`watchdog==2.1.6`, `requests==2.27.1`) inconsistent with the main project's `>=2.28.1` requests constraint. Separate cleanup.

## Recommendation

This PR is in much better shape than 24 hours ago. With the dev-image fix above I'd consider it ready to land, modulo the `upgrading.md` entry deferred to the release PR. Thanks for the thorough follow-up.

OjusWiZard (commented):
~One important issue I found: `deployments/Dockerfiles/development/Pipfile` still references `open-aea-ledger-ethereum-flashbots`, while the rest of this PR removes Flashbots support for the `open-aea 2.2.1` migration. Keeping this editable dev dependency looks like an incomplete cleanup and could break the development image if `/open-aea/plugins/aea-ledger-ethereum-flashbots` no longer exists upstream. I think that line should be removed as well.~

Just fixed in the above commit. Rest LGTM.

OjusWiZard (approved):

DavidMinarsch (approved):
Round-3 re-review against `e23be7d50`. All round-2 findings are fixed, and you caught several stragglers I had missed.

## ✅ Round-2 blocker resolved

**Dev Docker image:**
- `deployments/Dockerfiles/development/Dockerfile:1` — `ARG AEA_VERSION=2.1.0` → `2.2.1` ✅
- `deployments/Dockerfiles/development/Pipfile:16` — flashbots editable install removed ✅

## 🎯 Your comprehensive sweep caught files I missed

In round-2 I only flagged the dev image. Running my own full sweep on `e23be7d50`, your commit also bumps:

- `deployments/Dockerfiles/autonomy/Dockerfile` — same `ARG AEA_VERSION` pattern
- `deployments/Dockerfiles/autonomy-user/requirements.txt` — **three** `open-aea*==2.1.0` lines I hadn't flagged
- `skaffold.yaml` — `AEA_VERSION: "2.1.0"` → `"2.2.1"`
- `packages/valory/connections/ipfs/connection.yaml` — `open-aea-cli-ipfs: version: ==2.1.0`
- `packages/valory/skills/abstract_round_abci/skill.yaml` — same
- `docs/guides/define_agent.md` — example config had `open-aea-ledger-ethereum: version: ==2.1.0`
- `autonomy/constants.py` — `ABSTRACT_ROUND_ABCI_SKILL_WITH_HASH` bumped to reflect the hash cascade

The commit message candidly documents the methodology gap (grepping `==2.2.0` missed files still on `2.1.0` with no intermediate step; `--include` filename filters excluded extensionless `Pipfile`). Good postmortem.

## ✅ My own verification sweep on `e23be7d50`

- `grep -rn "==2\.1\.0" --exclude-dir=docs/api --exclude=HISTORY.md --exclude=poetry.lock` finds **only** `docs/upgrading.md:47` (`open-aea: ==2.0.8 -> ==2.1.0`) — a correct historical upgrading entry, not a stale pin.
- `grep -rn "AEA_VERSION"` confirms all three occurrences (skaffold, `autonomy/Dockerfile`, `development/Dockerfile`) are on `2.2.1`.
- `grep -rn "flashbots"` outside `packages/valory/skills/...`, `packages/valory/connections/ledger/connection.yaml`, and `docs/api/...` returns nothing. All remaining flashbots references are either the skill-level API intentionally preserved per the PR body, or auto-generated API docs mirroring that code. Consistent with stated intent.

## 🟡 One remaining tiny nit — not a blocker

`CLAUDE.md:156` still says:

```
- `open-aea[all]==2.1.0` - Core AEA framework
```

Documentation drift in the "Key Dependencies" section — should be `==2.2.1` to match reality. Nothing breaks; easy one-line bump if you want to squash it in, otherwise fine as a follow-up.

## ✅ CI

All static/lint/build/install/lock checks green. Unit tests + e2e running at time of this comment.

## Approving

This PR is ready to land. Both review rounds' blockers addressed, your sweep caught more than mine did, CI is clean. The `upgrading.md` entry for the open-aea bump and flashbots plugin removal is explicitly deferred to the release PR, which is reasonable given this PR's scope. Nice work on the thorough follow-through.

PR #2474, opened 2026-04-16
fix: switch release build from setup.py to python -m build

DhairyaPatel7:
## Summary

The v0.21.17 release workflow failed on Python 3.14 because `python setup.py sdist bdist_wheel` requires `setuptools`, which is no longer bundled with Python 3.12+.

```
python setup.py sdist
ModuleNotFoundError: No module named 'setuptools'
```

Switches to `python -m build --sdist --wheel` (PEP 517), which reads `[build-system]` from `pyproject.toml` and manages its own build dependencies. This matches how the upstream `open-aea` release workflow already builds.

## Changes

- **`Makefile` `dist` target:** `python setup.py sdist` + `python setup.py bdist_wheel` → `python -m build --sdist --wheel`
- **`release.yml`:** plugin builds now loop with `python -m build` instead of `poetry run python setup.py sdist bdist_wheel`. Added `build` to the `pip install` step.

## Test plan

- [ ] Merge → re-tag v0.21.17 → release workflow succeeds

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -14,8 +14,8 @@ jobs:
         os: [ubuntu-latest]
         python-version: ["3.14"]
     steps:
-      - uses: actions/checkout@master
-      - uses: actions/setup-python@v3
+      - uses: actions/checkout@v4
+      - uses: actions/setup-python@v5
         with:
           python-version: ${{ matrix.python-version }}
       - name: Install dependencies
@@ -23,25 +23,22 @@ jobs:
           sudo apt-get update --fix-missing
           sudo apt-get autoremove
           sudo apt-get autoclean
-          python -m pip install --upgrade pip
-          pip install poetry twine
+          python -m pip install --upgrade pip setuptools
+          pip install poetry twine build
           python --version
 
       - name: Build wheels and source tarball
         run: |
           echo Processing open-autonomy
-          poetry install --all-extras --no-interaction
-          poetry run make dist
+          # python -m build performs PEP 517 isolated builds — it does
+          # not need the project's runtime deps, only the `build` package.
+          # Run outside poetry's venv so system-installed `build` is used.
+          make dist
 
-          echo Processing aea-test-autonomy
-          cd plugins/aea-test-autonomy
-          poetry run python setup.py sdist bdist_wheel
-          cd ../..
-
-          echo Processing aea-helpers
-          cd plugins/aea-helpers
-          poetry run python setup.py sdist bdist_wheel
-          cd ../..
+          for plugin in aea-test-autonomy aea-helpers; do
+            echo "Processing $plugin"
+            (cd "plugins/$plugin" && python -m build --sdist --wheel)
+          done
 
       - name: Publish open-autonomy Framework to PyPI
         uses: pypa/gh-action-pypi-publish@release/v1
@@ -78,8 +75,8 @@ jobs:
         python-version: ["3.14"]
 
     steps:
-      - uses: actions/checkout@master
-      - uses: actions/setup-python@v3
+      - uses: actions/checkout@v4
+      - uses: actions/setup-python@v5
         with:
           python-version: ${{ matrix.python-version }}
       - name: Install dependencies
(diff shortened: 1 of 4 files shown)


OjusWiZard (dismissed):

.github/workflows/release.yml
```diff
@@ -33,15 +33,10 @@ jobs:
           poetry install --all-extras --no-interaction
           poetry run make dist
```
LOCKhart07: **Blocking:** this will fail with `ModuleNotFoundError: No module named 'build'`.

`build` is installed into the runner's system python on line 27, not into the poetry venv. The Makefile's `dist` target now calls `python -m build`, and `poetry run make dist` executes it inside poetry's isolated venv, where `build` is not a declared dep.

Verified locally on a minimal poetry project: `poetry run python -m build` → `No module named build`.

Fix options:
- Drop `poetry run` — the Makefile only needs a python that has `build` installed, which the runner's system python already does. `python -m build` performs PEP 517 isolated builds, so it doesn't need the project's runtime deps present.
- Or install `build` into the poetry env (`poetry run pip install build`, or add to poetry dev deps).

Note: since release.yml only runs on `release: published`, this won't be caught by PR CI — it'll fail again when v0.21.17 is re-tagged.

.github/workflows/release.yml
```diff
@@ -33,15 +33,10 @@ jobs:
...
-          poetry run python setup.py sdist bdist_wheel
-          cd ../..
-
-          echo Processing aea-helpers
-          cd plugins/aea-helpers
-          poetry run python setup.py sdist bdist_wheel
-          cd ../..
+          for plugin in aea-test-autonomy aea-helpers; do
+            echo "Processing $plugin"
+            (cd "plugins/$plugin" && python -m build --sdist --wheel)
```
LOCKhart07, line 40: Nit (pre-existing, but relevant here): `plugins/aea-test-autonomy/pyproject.toml:3` and `plugins/aea-helpers/pyproject.toml:3` use `build_backend` (underscore) instead of `build-backend` (hyphen). Per PEP 517 the misspelled key is silently ignored, so tools fall back to the legacy `setup.py` backend. It happens to work via build isolation, but since this PR switches plugin builds to `python -m build`, fixing the key makes the declared `setuptools.build_meta` backend actually take effect.

.github/workflows/release.yml
```diff
@@ -78,16 +73,16 @@ jobs:
...
+      - uses: actions/setup-python@v5
         with:
           python-version: ${{ matrix.python-version }}
       - name: Install dependencies
         run: |
           sudo apt-get update --fix-missing
           sudo apt-get autoremove
           sudo apt-get autoclean
-          python -m pip install --upgrade pip
+          python -m pip install --upgrade pip setuptools
```
LOCKhart07, line 87: Nit: `setuptools` is added to this job's pip upgrade but not to the first job's (`pip install poetry twine build` on line 27). If this is defensive against 3.14's unbundled setuptools (e.g. for legacy-sdist transitive deps that pip builds without isolation), it should be applied symmetrically. If it's needed only for `pip install .[all]` here, a short comment explaining why would help.

.github/workflows/release.yml
DhairyaPatel7: Fixed — dropped `poetry run` from `make dist`. `python -m build` does PEP 517 isolated builds so it only needs the `build` package on the system python, not in the poetry venv. Added a comment in the workflow explaining why.

.github/workflows/release.yml
DhairyaPatel7: Fixed — both `plugins/aea-test-autonomy/pyproject.toml` and `plugins/aea-helpers/pyproject.toml` corrected from `build_backend` to `build-backend`.

.github/workflows/release.yml
DhairyaPatel7: Fixed — added `setuptools` to the first job's pip install line (line 26) for symmetry.

LOCKhart07 (approved):
Re-reviewed at `0c60c67b`. All three issues from the prior pass are addressed:
- `poetry run make dist` → `make dist` (system `build` now used) ✅
- `build_backend` → `build-backend` in both plugin pyproject.tomls ✅
- `setuptools` added symmetrically to both jobs ✅

LGTM.

OjusWiZard (approved):

DavidMinarsch (approved):

PR #2476, opened 2026-04-16
fix: release v0.21.18 — move HWI to optional extra (fixes armv7 Docker builds)

DhairyaPatel7:
## Problem

v0.21.17 Docker image builds fail on `linux/arm/v7`:

```
ERROR: Failed to build 'hidapi'
ERROR: Failed building wheel for Pillow
```

### Root cause

`open-aea-ledger-ethereum-hwi` was added to `pyproject.toml` in Aug 2025 but was **never shipped in the published wheel** until v0.21.17 — because the old build system (`setup.py`) didn't include it in `base_deps`.

v0.21.17 switched to `python -m build` (PEP 517 / Poetry backend), which reads all `pyproject.toml` deps. For the first time, HWI appeared in the wheel metadata, pulling in `ledgerwallet → hidapi + Pillow` — neither has armv7 wheels anywhere.

### Why HWI belongs as an optional extra

HWI is only used when the user passes `--hwi` on the CLI. The import is already lazy with a try/except that tells users to pip install the plugin. The code handles its absence gracefully — it was designed to be optional.

### Why HWI is excluded from [all]

The Dockerfile at `deployments/Dockerfiles/autonomy-user/requirements.txt` installs `open-autonomy[all]`. If HWI were in `[all]`, the armv7 build would still pull `ledgerwallet → hidapi + Pillow` and fail. HWI is therefore intentionally excluded from `[all]` and only available via explicit `pip install open-autonomy[hwi]`.

### Local armv7 verification

Tested locally via `docker buildx build --platform linux/arm/v7` with HWI excluded from the install — all deps compiled and installed successfully on armv7 under QEMU (~20 min build). The key packages that previously failed (hidapi, Pillow) are no longer in the dependency tree.

## Commits

1. `a1dbd4037` — move HWI from base deps to `[hwi]` optional extra
2. `4a6901cf1` — version bump `0.21.17` → `0.21.18` + hash cascade
3. `01ed6bc9a` — exclude HWI from `[all]` (per LOCKhart07 review) + poetry lock regen

## Test plan

- [x] All 10 CI tox envs pass locally
- [x] 60 tests pass (`test_constants.py` + `test_base.py`)
- [x] `docker buildx --platform linux/arm/v7` simulation completes successfully
- [ ] CI matrix
- [ ] After merge: tag `v0.21.18` → release workflow builds Docker images on all 3 platforms


OjusWiZard (dismissed):

LOCKhart07 (approved):

DavidMinarsch (approved):
Reviewed for root-cause diagnosis, scope, and version-bump consistency. Ship it.

## Root cause diagnosis — spot-on

`open-aea-ledger-ethereum-hwi` was in `pyproject.toml` since Aug 2025 but the old `setup.py sdist/bdist_wheel` path never included it in `base_deps`, so it was silently absent from published wheels. v0.21.17's switch to `python -m build` (PEP 517 via poetry-core) finally reads pyproject.toml, which first exposed HWI in the wheel metadata → `ledgerwallet → hidapi + Pillow`, which has no armv7 wheels. Classic "the fix revealed a latent bug."

## Fix — well-scoped

- `pyproject.toml:21` — HWI now `{version = "==2.2.1", optional = true}`, registered as `[hwi]` extra.
- `setup.py:40-53` — mirror structure with `hwi_deps`; `[all]` intentionally excludes `[hwi]`, commented.
- `autonomy/chain/config.py:194-204` already guards the HWI import with try/except raising a ClickException that tells users to `pip install open-aea-ledger-ethereum-hwi`. Making it optional is safe.

## `[all]` exclusion is correct

If HWI were in `[all]`, the `autonomy-user` Docker image (which installs `open-autonomy[all]`) would still pull `ledgerwallet → hidapi + Pillow` and fail armv7. Intentional exclusion + explicit `[hwi]` extra is the right shape, clearly commented in both files. Good catch by @LOCKhart07.

## ✅ Version bump consistency

All `0.21.17 → 0.21.18` locations verified:
- `autonomy/__version__.py`, `pyproject.toml`, both plugin `setup.py` files, `configurations.py` (`MATCHING_FRAMEWORK_VERSION`), `tests/test_base.py`, `autonomy-user/requirements.txt`.
- Package yaml cascade: `open-aea-test-autonomy: ==0.21.17 → ==0.21.18` propagated through all skill/contract/connection/agent yamls; `autonomy packages lock` regenerated downstream hashes consistently.
- `ABSTRACT_ROUND_ABCI_SKILL_WITH_HASH` matches `packages/packages.json`.
- No stale `0.21.17` refs outside HISTORY/upgrading/RELEASE_PROCESS.

## ✅ Poetry lock

`poetry.lock` has `open-aea-ledger-ethereum-hwi` at 2.2.1 with `hwi = ["open-aea-ledger-ethereum-hwi"]` extras declaration (not in the default install set). Consistent with pyproject.toml.

## ✅ Docs

HISTORY.md and upgrading.md both have concise 0.21.18 entries. Upgrading note clearly tells `--hwi` users to install with `pip install open-autonomy[hwi]`.

## ✅ CI

All static/lint/lock/install checks green.

Approving.

OjusWiZard (approved):

PR #2487, opened 2026-04-22
fix(aea-helpers): resolve bin-template-path without importing bin_template

DhairyaPatel7:
## Problem

`aea-helpers bin-template-path` crashes on every invocation outside a PyInstaller-frozen process:

```
$ uv run aea-helpers bin-template-path
  File ".../aea_helpers/pyinstaller_deps.py", line 132, in bin_template_path
    from aea_helpers import bin_template
  File ".../aea_helpers/bin_template.py", line 41, in <module>
    validation_module._CUR_DIR = Path(sys._MEIPASS) / validation_module._CUR_DIR
AttributeError: module 'sys' has no attribute '_MEIPASS'
```

`bin_template.py` is the PyInstaller entry-point template — it patches AEA's config paths for the frozen bundle environment. `sys._MEIPASS` is only set by the PyInstaller bootloader at runtime, so importing the module outside a frozen process errors at module load. The file's own docstring explicitly says _"This file is NOT meant to be executed or linted in the aea-helpers context."_

`bin_template_path()` nevertheless did `from aea_helpers import bin_template` just to read `bin_template.__file__`, turning every build-time invocation into the crash above.

## Reach

Latent in **0.21.17, 0.21.18, and 0.21.19**. The command was introduced in 0.21.17 with this bug and has never worked.

Net effect on downstream agent repos: every `make build-agent-runner` that uses `$(shell uv run aea-helpers bin-template-path)` to locate the template — which is the canonical pattern in the trader / meme-ooorr / optimus Makefiles — gets an empty shell expansion, and pyinstaller fails with ``pyinstaller: error: the following arguments are required: scriptname``.

Reproduced locally on trader's `chore/bump-deps-cleanup` branch against both 0.21.18 and 0.21.19.

## Fix

Resolve the template path via `Path(aea_helpers.__file__).parent / \"bin_template.py\"` instead of importing the template module. `aea_helpers/__init__.py` is a normal package module and safe to import outside PyInstaller.

```diff
-    from aea_helpers import bin_template
-    click.echo(bin_template.__file__, nl=False)
+    import aea_helpers
+    path = Path(aea_helpers.__file__).parent / \"bin_template.py\"
+    click.echo(str(path), nl=False)
```

Added a comment at the call site so the non-obvious reason stays visible.

## Test plan

- [x] Reproduced the crash on trader `chore/bump-deps-cleanup` against both `open-aea-helpers==0.21.18` and the `0.21.19`-era source in this repo (editable install).
- [x] With this patch applied (editable install into trader's venv):
  - `uv run aea-helpers bin-template-path` → prints the template path, exit 0.
  - `make build-agent-runner` → 60 MB Mach-O arm64 at `dist/agent_runner_bin`, `--version` succeeds.
  - `make check-agent-runner` → `[OK] Found 'Starting AEA' within 80s`.
- [ ] CI (relying on repo test suite; the failing path was never covered because it only triggers outside frozen bundles).

## Downstream impact

Agent repos currently carry no-op `bin-template-path` invocations in their build targets. Once this lands in a release, they can unwind any local workarounds.


plugins/aea-helpers/aea_helpers/pyinstaller_deps.py
```diff
@@ -129,6 +129,13 @@ def bin_template_path() -> None:
...
-    click.echo(bin_template.__file__, nl=False)
+    # Do NOT `import aea_helpers.bin_template`: the module runs
+    # `Path(sys._MEIPASS)` at top level, which is only defined inside a
+    # PyInstaller-frozen process. Outside a frozen bundle the import
+    # crashes with `AttributeError: module 'sys' has no attribute
+    # '_MEIPASS'`, so `make build-agent-runner` in any downstream agent
+    # repo would fail before pyinstaller even starts.
+    import aea_helpers  # pylint: disable=import-outside-toplevel
+
+    path = Path(aea_helpers.__file__).parent / "bin_template.py"
```
DIvyaNautiyal07, line 140: Nit: `pyinstaller_deps.py` lives in the same directory as `bin_template.py`, so `Path(__file__).parent / "bin_template.py"` resolves to the same path without the self-import (and the `import-outside-toplevel` pragma becomes unnecessary). Strictly cosmetic — current code is correct.

DIvyaNautiyal07 (approved):
LGTM. Fix is correct and minimal: `bin_template.py` runs `Path(sys._MEIPASS)` at module top level, so importing it outside a frozen process crashes — resolving the path via `Path(aea_helpers.__file__).parent` avoids the import. Repro and downstream impact in the PR description match what I see in the code.

One stylistic nit inline.

LOCKhart07 (approved):

PR #2493, opened 2026-05-01
fix: bundle all ejected contracts in wheel/sdist

DhairyaPatel7:
## Summary

Restores wheel/sdist bundling for the 11 contracts under `autonomy/data/contracts/` that have been silently dropped from PyPI artifacts since `v0.21.17`.

## Background

`make eject-contracts` (Makefile:131-137) copies 15 contracts from `packages/valory/contracts/` into `autonomy/data/contracts/<name>/` so the framework can resolve them at runtime via `autonomy.data.contracts.<name>` when installed without the `packages/` tree.

Until v0.21.16, `make dist` ran `python setup.py sdist bdist_wheel`. setuptools honoured `setup.py`'s `package_data` and `find_packages()` and the ejected copies were shipped. Commit `0c60c67b7` (v0.21.17) switched to `python -m build --sdist --wheel` because Python 3.14 dropped bundled setuptools and broke `python setup.py`. `python -m build` is PEP 517-compliant and uses the backend declared in `[build-system]` — `poetry.core.masonry.api`. Poetry-core ignores `setup.py` entirely and excludes anything matching `.gitignore`.

`.gitignore` lists 11 of the ejected contract dirs (to prevent accidentally committing duplicate copies), so poetry-core dropped them. The other 4 contracts were missing from `.gitignore` by oversight, which is the only reason any contracts shipped at all.

### Evidence

Contracts in published PyPI wheels under `autonomy/data/contracts/`:

| Version | Count |
|---|---|
| 0.21.14, 0.21.15, **0.21.16** | **15** (all expected) |
| **0.21.17**, 0.21.18, 0.21.19 | **4** (`erc8004_identity_registry`, `erc8004_identity_registry_bridger`, `poly_safe_creator_with_recovery_module`, `sign_message_lib`) |

## Fix

- `pyproject.toml`: add explicit `[tool.poetry] include` directive — poetry-core's `include` overrides VCS exclusions, so ejected copies ship regardless of `.gitignore`.
- `.gitignore`: add the 4 missing entries (`erc8004_identity_registry`, `erc8004_identity_registry_bridger`, `poly_safe_creator_with_recovery_module`, `sign_message_lib`) so the working tree stays clean after `make eject-contracts`. Bundling is now guaranteed by the explicit `include`, not by `.gitignore` accidentally omitting them.

## Test plan

- [x] `make dist` succeeds and the resulting wheel ships all 15 contract dirs (verified locally)
- [x] Built wheel ships 17 `build/*.json` ABI files — matches v0.21.16 baseline exactly
- [x] Working tree stays clean (no untracked files) after `make eject-contracts`
- [ ] CI passes
- [ ] Smoke-test installing the resulting wheel and importing each of the 11 newly-bundled contracts

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -407,14 +407,18 @@ packages/valory/protocols/ledger_api
 autonomy/data/contracts/agent_registry
 autonomy/data/contracts/component_registry
 autonomy/data/contracts/erc20
+autonomy/data/contracts/erc8004_identity_registry
+autonomy/data/contracts/erc8004_identity_registry_bridger
 autonomy/data/contracts/gnosis_safe
 autonomy/data/contracts/gnosis_safe_proxy_factory
 autonomy/data/contracts/multisend
+autonomy/data/contracts/poly_safe_creator_with_recovery_module
+autonomy/data/contracts/recovery_module
 autonomy/data/contracts/registries_manager
 autonomy/data/contracts/service_manager
 autonomy/data/contracts/service_registry
 autonomy/data/contracts/service_registry_token_utility
-autonomy/data/contracts/recovery_module
+autonomy/data/contracts/sign_message_lib
 
 .vscode/
 
--- a/.gitignore
+++ b/.gitignore
@@ -413,12 +413,12 @@ autonomy/data/contracts/gnosis_safe
 autonomy/data/contracts/gnosis_safe_proxy_factory
 autonomy/data/contracts/multisend
 autonomy/data/contracts/poly_safe_creator_with_recovery_module
-autonomy/data/contracts/recovery_module
 autonomy/data/contracts/registries_manager
 autonomy/data/contracts/service_manager
 autonomy/data/contracts/service_registry
 autonomy/data/contracts/service_registry_token_utility
 autonomy/data/contracts/sign_message_lib
+autonomy/data/contracts/recovery_module
 
 .vscode/
 
--- a/tests/test_autonomy/test_packaging.py
+++ b/tests/test_autonomy/test_packaging.py
@@ -21,9 +21,7 @@
 
 import re
 import sys
-from pathlib import Path
 from typing import List
-
 
 if sys.version_info >= (3, 11):
     import tomllib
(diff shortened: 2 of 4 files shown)


.gitignore
```diff
@@ -407,14 +407,18 @@ packages/valory/protocols/ledger_api
...
 autonomy/data/contracts/gnosis_safe
 autonomy/data/contracts/gnosis_safe_proxy_factory
 autonomy/data/contracts/multisend
+autonomy/data/contracts/poly_safe_creator_with_recovery_module
+autonomy/data/contracts/recovery_module
 autonomy/data/contracts/registries_manager
 autonomy/data/contracts/service_manager
 autonomy/data/contracts/service_registry
 autonomy/data/contracts/service_registry_token_utility
-autonomy/data/contracts/recovery_module
```
OjusWiZard, line 417: why remove this?

OjusWiZard (commented):
can we please add a test also, that avoids this from happening again?

tests/test_autonomy/test_packaging.py
OjusWiZard, line 1: no I mean it's just matching regex from the code, but we're not really "verifying" that the final bundle has it. I'm not sure how complex it is but would be good to have if possible. A test something like:
1. package the wheel/sdist
2. verify len(expected_contract) > 0
3. test the expected_contracts exists by importing (and optionally using) them

DhairyaPatel7: Added in `6bc9b032c`. The new `test_built_wheel_ships_ejected_contracts`:

1. Replicates `make eject-contracts` in Python (cross-platform — no `make` dep on Windows runners).
2. Runs `python -m build --wheel` via subprocess.
3. Opens the wheel zip and asserts both `contract.py` and `contract.yaml` are present for every contract in the eject list.
4. Has a `len(contracts) >= 15` sanity floor.

Gated to `linux + py3.14` so the ~30s build runs once per push instead of 9 times across the matrix — wheel content is platform-independent, so single-cell coverage is sufficient. Added `build` to the shared test deps so the targeted job actually exercises it.

I skipped the *importing* step deliberately: a real import requires either a fresh-venv install (~60s+ extra) or sys.path tricks that get shadowed by the test env's editable install. The zip-content assertion proves the file is in the artifact and non-empty, which is what catches the v0.21.17–v0.21.19 regression class. Happy to add the install-and-import variant if you'd like the stronger guarantee despite the cost — let me know.

OjusWiZard (approved):

LOCKhart07 (approved):

PR #2494, opened 2026-05-01
Surface named env-var placeholders from service overrides

DhairyaPatel7:
## Summary

- `Service.process_component_overrides` previously emitted env-var entries keyed only by the YAML traversal path (e.g. `SKILL_DUMMY_SKILL_MODELS_PARAMS_ARGS_MESSAGE`). Named placeholders like `${HELLO_WORLD_MESSAGE:str:hi}` left no `HELLO_WORLD_MESSAGE` entry in the agent's env dict, so downstream callers (e.g. middleware injecting from `service.yaml`) could not override values by their declared name.
- Added `Service._extract_named_env_var_defaults`, a parallel walk over the raw override config (run *before* `apply_env_variables` substitutes the placeholders away) that collects `{NAME: default}` for every `${NAME:type:default}` placeholder. The collected entries are merged with `setdefault` so existing path-keyed entries always win on conflict.
- Path-based env-var output is unchanged. For mixed configs, both keys are emitted side-by-side.

## Why

The prior behavior meant named env vars declared in `service.yaml` were silently dropped from the generated agent env. Middleware workarounds had to inject them manually based on the service template, which only covered template-declared names — anything added directly in `service.yaml` was missed.

## Test plan

- [x] `tests/test_autonomy/test_configurations.py` — three new cases (`test_named_env_var_placeholders_are_emitted`, `test_anonymous_placeholders_emit_only_path_keys`, `test_named_and_anonymous_placeholders_coexist`) covering named-only, anonymous-only, and mixed placeholders. All 16 tests in the file pass.
- [x] `tests/test_autonomy/test_deploy/test_service_specification.py` — 21 existing service-builder tests still pass.
- [x] Local lint: black-check, isort-check, flake8, mypy, pylint (10/10), bandit, safety, vulture, liccheck all green.

## Notes

- `check-packages` / `check-hash` / `check-dependencies` fail locally on `connection/valory/ledger:0.19.0` due to pre-existing main-branch hash drift unrelated to this PR (no `packages/` files touched here).
- Marking draft for review.

### DIFF
--- a/autonomy/configurations/base.py
+++ b/autonomy/configurations/base.py
@@ -47,7 +47,11 @@ from aea.configurations.data_types import (
 )
 from aea.exceptions import AEAValidationError
 from aea.helpers.base import SimpleIdOrStr
-from aea.helpers.env_vars import apply_env_variables, generate_env_vars_recursively
+from aea.helpers.env_vars import (
+    ENV_VARIABLE_RE,
+    apply_env_variables,
+    generate_env_vars_recursively,
+)
 
 from autonomy.configurations.constants import DEFAULT_SERVICE_CONFIG_FILE, SCHEMAS_DIR
 from autonomy.configurations.validation import ConfigValidator
@@ -291,6 +295,8 @@ class Service(PackageConfiguration):  # pylint: disable=too-many-instance-attrib
                 raise ValueError(
                     f"Overrides not provided for agent {agent_idx}; component={component_id}"
                 )
+        named_env_vars = self._extract_named_env_var_defaults(configuration)
+
         configuration = apply_env_variables(
             data=configuration, env_variables=os.environ.copy()
         )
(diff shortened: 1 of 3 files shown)


autonomy/configurations/base.py
```diff
@@ -291,16 +295,49 @@ def process_component_overrides(
...
+
         configuration = apply_env_variables(
             data=configuration, env_variables=os.environ.copy()
         )
         env_var_dict = self.generate_environment_variables(
             component_id=component_id,
             component_configuration_json=configuration,
         )
+        for name, default in named_env_vars.items():
+            env_var_dict.setdefault(name, default)
```
OjusWiZard: it's setting the default, but it should also prefer the values from os.environ no?

OjusWiZard (approved):

LOCKhart07 (approved):

PR #2498, opened 2026-05-07
chore(claude-skills): cover gaps from trader/optimus/meme-ooorr audits

### LINKED ISSUE #2495: claude-skills: gap analysis from a real audit run on valory-xyz/trader
# claude-skills: gap analysis from a real audit run on `valory-xyz/trader`

Ran `audit-fsm` and `audit-resilience` against the [`trader`](https://github.com/valory-xyz/trader) repo and recorded what each skill *did not* catch but plausibly should. The skills are good — the most impactful findings (an unprotected `json.loads` that wedges a single-thread connection, two third-party clients with no timeout, a 12-week-stale fallback rate) were caught. This issue lists what was missed, ordered by practical impact.

References to specific findings from those audit runs:
- FSM: [valory-xyz/trader#937](https://github.com/valory-xyz/trader/issues/937)
- Resilience: [valory-xyz/trader#938](https://github.com/valory-xyz/trader/issues/938)

---

### 1. Determinism / consensus correctness in `end_block()`
`AbstractRound.end_block()` must be a **pure function of consensus state** — but the skill never asks about it. Concrete example from the trader repo: `CheckBenchmarkingModeRound.end_block()` (`states/check_benchmarking.py:64-112`) reads a **local JSON file** to choose its event. In a multi-agent service that breaks consensus (different agents see different files); in a single-agent service it is fine. The skill flagged nothing.

**Proposed check (new C-level):** in any `end_block()` body, flag:
- File I/O (`open(`, `Path.read_*`, `json.load(open`, etc.)
- `time.time()`, `datetime.now()` (use `last_round_transition_timestamp` instead)
- `random.*` (use the consensus randomness)
- `os.environ.get(`
- Calls to non-deterministic external services

If the surrounding service is single-agent / sovereign (e.g. Pearl-shipped), demote to "informational" but still surface — the call may break if the service is later promoted to multi-agent.

### 2. SynchronizedData type round-trip safety
The skill checks `collection_key`/`selection_key` ↔ property-name mapping (M5) but not whether the values written survive serialization:
- `set`s, `Decimal`s, `dataclass` instances are JSON-non-serializable.
- Custom serializer hooks (e.g. `EGreedyPolicy.deserialize`) are easy to forget on the read side.
- Multiple inheritance of `SynchronizedData` (in trader: `MechInteractSyncedData`, `MarketManagerSyncedData`, `TxSettlementSyncedData`, `ChatuiSyncedData` — `decision_maker_abci/states/base.py:92-97`) can silently shadow properties when two parents define the same name.

**Proposed check:** for each `SynchronizedData` property, walk up the MRO and warn if any parent defines a property with the same name (silent shadowing). For each round's writes, infer the value type and flag JSON-incompatible types.

### 3. `cross_period_persisted_keys` completeness
The skill's M3 says "post-conditions of predecessor should be a superset of pre-conditions of successor" but does not verify that **every db key read by the post-reset initial round is in `cross_period_persisted_keys`**. A missing key works in the first period (it gets written then) but disappears after the reset.

**Proposed check:** statically collect every `self.db.get` / `get_strict` in the initial round chain after a reset, and intersect with `cross_period_persisted_keys`. Diff is the finding.

### 4. Payload schema / module-path stability
`_MetaPayload` keys payloads by `{module}.{ClassName}`. Renaming a payload class or moving its module breaks replay of persisted Tendermint state and breaks rolling deployments where some agents are on the new version.

**Proposed check:** flag payload class moves/renames against `git log`. Hard to do statically without history; could be done as an optional "pre-merge" mode that diffs against `main`.

### 5. Cross-skill imports
`chain()` deliberately allows shared events (e.g. `decision_maker_abci` imports `MarketManagerEvent.FETCH_ERROR`), but other forms of cross-skill coupling are silent footguns: importing a behaviour class, monkey-patching a parent's `synchronized_data_class`, etc.

**Proposed check:** enumerate cross-skill imports in the audit scope; classify each as data-class import (OK), event import (OK if intentional), behaviour import (suspicious), monkey-patch (always bad).

### 6. `customs/` strategies
The skill scope ends at `skills/`. But `customs/` (e.g. `kelly_criterion`, `fixed_bet`) is loaded by IPFS hash at runtime and decides bet sizing — i.e. directly drives the FSM's financial actions. They get zero coverage today.

**Proposed extension:** when the agent config references `FILE_HASH_TO_STRATEGIES`, also enter `customs/<strategy>/` with a reduced checklist (no FSM, but type safety, numeric precision, error handling).

### 7. Logging hygiene
Behaviours log payloads, API responses, signed orders. No automated check for secret/PII leakage. Patterns to flag:
- `logger.*(f"... {private_key|api_key|token|secret} ...")` (regex on variable names)
- Logging entire signed order objects
- Logging full HTTP response bodies that might contain auth tokens

**Proposed check (new L-level):** name-based regex on variables interpolated into log lines. Low false-positive rate if scoped to common secret-name tokens.

### 8. Numeric precision in financial logic
No check for:
- `float` arithmetic on token amounts (silent rounding)
- Mixing 6-decimal (USDC) and 18-decimal (most ERC-20) without unit tracking
- Integer overflow in older Solidity-call result handling
- `==` on floats

The trader repo's `FALLBACK_POL_TO_USD_RATE = 0.089935` (a `float` literal) flows through `int(pol_amount * (10**18))` (`trader_abci/handlers.py:643`) — fine here, but the pattern is fragile.

### 9. Service-vs-agent config drift
The trader repo has separate `trader_pearl/service.yaml` (Omenstrat) and `polymarket_trader/service.yaml` (Polystrat) overrides on top of one `aea-config.yaml`. Local dev defaults differ from deployed service. CLAUDE.md explicitly warns about this. The skill never reads service YAML — it audits package source only.

**Proposed extension:** if `services/*/service.yaml` exists, parse param overrides and verify they don't shadow types/defaults in incompatible ways with what behaviours actually expect.

### 10. `is_abstract: true` semantics
The agent_performance dialogues finding worked out, but only because the CLI tool happened to fail. In general the skill has no notion that an `is_abstract` skill is composed-only and doesn't need standalone wiring. Other audits (e.g. is the skill self-contained for testing?) should branch on this flag.

---

### 1. Idempotency on retry of financial actions
**The biggest gap.** The skill flags blocking timeouts (BP8, BP9) but never asks "if this retries, can it duplicate a financial action?"

In trader: `polymarket_client` retries Gamma reads (idempotent — fine) but the bet-placement / set-approval / redeem paths don't retry at the connection layer — yet **the FSM retries the entire round** on its own timeout. A partial settlement (signed order accepted upstream, response lost) plus an FSM retry can re-place the same order. The CLOB nonce protects some of this, but the audit should flag the *pattern* and verify the dedup mechanism per call site.

**Proposed check (new BP):** for every POST to a non-idempotent endpoint, trace whether the FSM round it sits inside can re-enter it after timeout. If yes, document whether there's an idempotency key (nonce, request-id, on-chain check) and flag if not.

### 2. TLS / auth handling
No check for:
- `verify=False` on `requests.*` / `httpx.*`
- Hardcoded API keys/tokens (regex)
- Secrets in URLs vs headers (e.g. `?apikey=...`)
- Token expiry / refresh handling
- `set_api_creds` failure paths (in the trader repo, if `client.create_or_derive_api_key()` fails the connection is unusable but no explicit handling exists)

### 3. Pagination correctness and partial failure
The skill's failure matrix asks about "200 but `[]`" but not "halfway through a paginated read, page 5 returns 5xx." In trader: `_fetch_markets_by_tag_slug` (`polymarket_client/connection.py:557+`) is a `while True` paginator — a mid-pagination failure returns `None, error` and the partial list is discarded. If callers wanted partial data, they don't get it; if they wanted all-or-nothing, that's enforced. Either way, the *contract* should be explicit.

**Proposed check:** every paginated loop should be flagged with a question: "what does the caller see on partial failure — empty list, partial list, or error?" Mismatch with caller expectation is a bug.

### 4. Response size bounds
No audit of:
- Subgraph queries with no `limit:` clause
- Iterating `mech_responses` from a degenerate API (memory blowup)
- Streaming binary responses through `response.body.decode()` (already partly covered by BP5 but only in the malformed-bytes sense, not the size sense)

### 5. Clock-window-signed requests
Signed orders / relayer transactions typically include a timestamp + expiry window. Clock drift vs the upstream causes signature failures. Not in the skill today.

**Proposed check:** find calls to `relayer_client.execute(...)`, `client.post_order(...)` etc. and verify whether the calling code uses `time.time()` (drift-prone) or `last_round_transition_timestamp` (consensus-safe — wrong here, it's an *agent-local* timestamp). Note this also intersects with FSM determinism (gap #1 above).

### 6. Observability gaps
The skill has CC3 ("inconsistent error reporting") but does not catalog. A useful output is a table:

| Call site | Logs success? | Logs failure? | Emits metric? | Bubbles to UI? |

Without this, a stale `FALLBACK_POL_TO_USD_RATE` is invisible until P&L diverges weeks later.

### 7. Reconcile worst-case blocking with actual round timeouts
The skill computes "this can block N seconds" but never compares against the FSM round timeout for the round that calls it. The trader audit found 60 s of polymarket retry blocking — the round timeout for `BetPlacementRound` is 30 s. So the round times out *before* the connection responds, leaving an orphaned worker thread that keeps running after the FSM has moved on. That's a resource-leak pattern.

**Proposed check:** for each external call site, look up the round_timeout of the calling round, and warn when worst-case_blocking > round_timeout.

### 8. Dependency CVE / pinning
`py_clob_client_v2`, `py_builder_relayer_client`, `httpx`, `requests`, `web3` — no check that pins are current or free of known CVEs. Out of scope for a static audit, but a pointer to "run `safety check` / `pip-audit` after this audit" in the report would help.

---

## Cross-cutting gaps in both skills

1. **No `service.yaml` override coverage.** Both skills read package source; neither parses service-level overrides that change runtime behaviour materially (in trader: Omenstrat vs Polystrat).

2. **No data-flow-across-composition trace.** Both skills audit one skill at a time. The trader_abci composition chains 11 apps — a bug in cross-period data flow (e.g. `cross_period_persisted_keys` missing a key the next period needs) wouldn't surface in either pass.

3. **No security overlay.** Neither catches private-key handling, env-var injection in `skill.yaml`, `eval`/`exec` patterns, unsafe deserialization (`pickle.loads`).

4. **`is_abstract: true` not consulted.** Many of the agent_performance "missing dialogues" / "missing fsm_specification" findings would benefit from knowing whether the skill is meant to run standalone.

5. **Library-skill convention is partly captured but not parameterized.** The audit-fsm skill has a hardcoded "library skills" list (`registration_abci`, `reset_pause_abci`, ...). When a downstream repo defines its own library skills, the convention guidance breaks.

6. **No "what does this audit *not* cover" section in the report.** Helpful to call out e.g. "this audit did not check security / numeric precision / dependency CVEs" so the consumer knows what they still need to do.

---

## Suggested prioritization for skill updates

Highest leverage (likely to catch real bugs in real services):

1. **Resilience #1 — idempotency on retry of financial actions.** Direct financial impact; easy class of bug to introduce; not currently audited.
2. **FSM #1 — determinism in `end_block()`.** Will catch bugs the moment a service becomes multi-agent. Cheap to implement.
3. **Resilience #7 — worst-case blocking vs round timeout.** Catches the orphaned-thread pattern and re-entrancy issues. Mostly a join across data the skill already gathers.
4. **FSM #6 — `customs/` strategies.** Today's coverage is zero; the actual bet-sizing logic is here.
5. **Both — `service.yaml` override coverage.** Common deployment footgun.

Happy to test follow-up versions on the trader repo if helpful.

DavidMinarsch:
Closes #2495.

Extends both audit skills with the gaps documented in #2495 from real audit runs against `valory-xyz/trader`, `valory-xyz/optimus`, and `valory-xyz/meme-ooorr`.

### `audit-fsm`
- **C5** non-determinism in `end_block()` (file I/O, `time.time`, `random.*`, `os.environ`, network)
- **M6** `SynchronizedData` JSON round-trip safety + MRO shadowing across composed data classes
- **M7** `cross_period_persisted_keys` completeness for post-reset initial-round reads
- **M8** misleading `Event`-valued class attributes on `CollectSameUntilThresholdRound` subclasses
- **L4** logging hygiene (secret / PII leakage)
- **L5** numeric precision in financial logic
- **C3** carve-out for framework-scheduled events (`ROUND_TIMEOUT` from `event_to_timeout`) — fixes the meme-ooorr false positive
- Scope extensions for `customs/` strategies (when `FILE_HASH_TO_STRATEGIES` is set) and `services/*/service.yaml` overrides
- Methodology guardrails: data-type trace before flagging type mismatches
- `is_abstract: true` semantics; library-skill list now marked as project-specific
- CLI-tool-unavailable handling in Step 0 (do not report missing library skills as a finding)
- Coverage Limitations + report-template "what this audit did not cover"

### `audit-resilience`
- **BP15** idempotency on retry of financial actions — the highest-leverage gap
- **BP16** stringified-exception retry conditions (`if "rate limit" in str(e)`)
- **BP17** defined-but-unenforced security allowlists
- **BP18** inverted fail-safe on unknown state (RPC down → optimistic default)
- **BP19** `verify=False` / hardcoded credentials / secrets in URL query strings
- **BP20** pagination partial-failure semantics
- **BP21** response size bounds
- **BP22** clock-window-signed requests (intersects with FSM determinism)
- **CC8** worst-case blocking vs `round_timeout` (orphaned thread pattern)
- **CC9** per-call-site observability table
- `services/*/service.yaml` override coverage
- Methodology guardrails: data-type trace + verbatim BP1 pre-flag reminder (the `JSONDecodeError`-vs-`ValueError` false positive)
- Coverage Limitations + report-template observability table, override findings, and "what this audit did not cover"

Two issue items deferred deliberately: audit-fsm gap #4 (payload-class git-history diffing — not statically derivable) and #5 (cross-skill import classification — partial coverage exists via existing checks). Everything else from the issue body and the optimus / meme-ooorr follow-up comments is wired in.

## Test plan

- [ ] Re-run `/audit-fsm` against `valory-xyz/trader` and confirm the C5 / M6-M8 / L4-L5 findings surface
- [ ] Re-run `/audit-resilience` against `valory-xyz/trader` and confirm BP15 (idempotency) is now flagged
- [ ] Re-run `/audit-fsm` against `valory-xyz/meme-ooorr` and confirm `ROUND_TIMEOUT`-on-`end_block` false positive is gone
- [ ] Re-run `/audit-resilience` against `valory-xyz/optimus` and confirm `JSONDecodeError`-on-`except ValueError` false positives are gone
- [ ] Both reports include a "What this audit did not cover" section

### DIFF
--- a/claude-skills/audit-fsm/SKILL.md
+++ b/claude-skills/audit-fsm/SKILL.md
@@ -377,13 +377,14 @@ Different agents may have different files → different events → consensus bre
 
 **Severity by service shape:**
 - **Multi-agent service:** Critical — consensus break.
-- **Single-agent / sovereign service** (e.g. some Pearl-shipped configurations): demote to **informational** but still surface — the bug appears the moment the service is promoted to multi-agent.
+- **Single-agent / sovereign service** (e.g. some Pearl-shipped configurations): demote to **Low (L)** with a note: *"severity escalates to Critical when the service runs with N>1 agents."* This keeps the finding in the standard L section of the report rather than vanishing — Pearl-shipped services regularly get promoted to multi-agent, so silent loss is the worst outcome.
 
 **How to check:**
 1. Walk every `end_block()` body and any function it calls.
-2. Grep the imports of the round module for `time`, `datetime`, `random`, `os`, `requests`, `urllib`, `pathlib`.
-3. For each non-deterministic call found, verify whether the surrounding code path can reach `end_block()`.
-4. Check whether the agent's service config (look at `services/*/service.yaml` `number_of_agents`) is 1 or N to set severity.
+2. **Also walk every `process_payload()` body in the same round.** `process_payload()` populates `self.synchronized_data` / round-local state that `end_block()` then reads — non-deterministic writes in `process_payload()` produce non-deterministic `end_block()` reads even when the `end_block()` body itself is clean. An auditor who walks only `end_block()` will report a clean C5 on a real consensus break.
+3. Grep the imports of the round module for `time`, `datetime`, `random`, `os`, `requests`, `urllib`, `pathlib`.
+4. For each non-deterministic call found, verify whether the surrounding code path can reach `end_block()` (directly or via `process_payload()` → `synchronized_data` propagation).
+5. Check whether the agent's service config (look at `services/*/service.yaml` `number_of_agents`) is 1 or N to set severity.
 
 **Fix:** All non-deterministic data must be sourced from `synchronized_data` — i.e. agreed via consensus by the predecessor round's payload.
 
@@ -561,11 +562,16 @@ Plus: when a `SynchronizedData` subclass uses **multiple inheritance** to combin
 
 **Bug examples:**
 ```python
-# BUG: set is not JSON-serializable — TypeError at write time
+# BUG (write-time crash): set is not JSON-serializable — TypeError at write time
 self.synchronized_data.update(participants={agent1, agent2})
 
-# BUG: Decimal becomes float after JSON round-trip — silent precision loss
+# BUG (write-time crash): Decimal is not JSON-serializable — TypeError at write time
+# (same failure mode as set; do NOT mistake this for silent rounding to float)
 self.synchronized_data.update(price=Decimal("0.123456789012345678"))
+
+# BUG (silent precision loss): float arithmetic drifts BEFORE serialization;
+# JSON round-trip then preserves the wrong value
+self.synchronized_data.update(price=0.1 + 0.2)  # writes 0.30000000000000004
 
 # BUG: same property name in two parents
 class MechInteractSyncedData:
@@ -585,9 +591,9 @@ class SynchronizedData(MechInteractSyncedData, MarketManagerSyncedData):
 **What:** After a period reset (`ResetAndPauseRound`), the next period restarts at the chain's initial round. Any DB key that the initial round chain (or its behaviours) reads via `self.db.get` / `self.db.get_strict` must be in `cross_period_persisted_keys` — otherwise it's missing after the first reset and the read returns `None` / raises `KeyError`. The bug is invisible in the first period (the key was just written) and surfaces only after the reset.
 
 **How to check:**
-1. Identify the initial round chain after a reset (typically `RegistrationRound` → ... → first business round). Use the composed app's `transition_function` to follow `Event.RESET_TIMEOUT` / `Event.DONE` from `ResetAndPauseRound` to the next round.
+1. Identify the initial round chain after a reset (typically `RegistrationRound` → ... → first business round). Use the composed app's `transition_function` to follow **`Event.DONE`** from `ResetAndPauseRound` to the next round — this is the normal post-reset re-entry path. `Event.RESET_AND_PAUSE_TIMEOUT` and `Event.NO_MAJORITY` are failure paths that route to `FinishedResetAndPauseErrorRound` and do not represent the post-reset initial-read set (see `packages/valory/skills/reset_pause_abci/rounds.py:94-98`). Tracing the failure path will surface bogus keys or miss the real ones.
 2. Statically collect every `self.db.get(...)` / `self.db.get_strict(...)` and every `self.synchronized_data.<property>` access in those rounds and their behaviours. Resolve property accesses to the underlying DB key.
-3. Subtract `cross_period_persisted_keys`. The diff is the set of keys that will be missing after the first reset.
+3. Subtract `cross_period_persisted_keys` **AND `AbciAppDB.default_cross_period_keys`**. The framework auto-injects the defaults at every period boundary (`packages/valory/skills/abstract_round_abci/base.py:516-523` then `:549`); a skill that reads any of `{all_participants, participants, consensus_threshold, safe_contract_address}` after a reset does NOT need to redeclare them — they survive every period automatically. Subtracting only `cross_period_persisted_keys` produces false positives on every audit. The remaining diff is the set of keys that will genuinely be missing after the first reset.
 
 **Why it matters:** First-period success masks the bug. Production crashes appear only after the first natural reset (often hours or days into a deployment).
 
--- a/claude-skills/audit-fsm/SKILL.md
+++ b/claude-skills/audit-fsm/SKILL.md
@@ -176,6 +176,24 @@ Downstream projects compose these library skills with their own custom skills vi
 
 **Audit implication:** Do NOT flag `ROUND_TIMEOUT` (or other standard event names) as M2 findings in library skills when the enum member exists but isn't used in the skill's own transition function. This is a convention, not dead code.
 
+**Project-specific library skills.** The list above is the *framework default*. Downstream projects often introduce their own library skills (skills meant to be composed only). When auditing a downstream repo:
+
+- Skills with `is_abstract: true` in `skill.yaml` are typically library skills.
+- Check the project's `CLAUDE.md` / `AGENTS.md` for an explicit library-skills list.
+- Read `aea-config.yaml` of the agent under audit — skills that appear only as composition imports (and are never directly wired to the FSM by the agent) are library skills.
+- Apply the M2 / C4 carve-outs to those project-specific library skills as well.
+
+### `is_abstract: true` Semantics
+
+A skill with `is_abstract: true` in its `skill.yaml` is composition-only — it is never instantiated as a top-level skill, only consumed by another `chain()` call. Audit checks branch on this flag:
+
+- **Skip** "missing `fsm_specification.yaml`" findings — abstract skills are composed-only and don't need standalone wiring.
+- **Skip** "missing dialogue X" findings if the dialogue is provided by the composing skill.
+- **Demote** "missing handler" findings to informational unless the handler is required for composition (e.g. an `ABCIHandler` is mandatory).
+- **Apply** the same M2 / C4 carve-outs as library skills (unused events / dead timeouts may be intentional extension points).
+
+The `agent_performance` "missing dialogues" / "missing fsm_specification" findings from prior audits worked out only because the CLI tool happened to fail. With the `is_abstract` branch, those findings are correctly suppressed at the source.
+
 ### Test Infrastructure
 
 The framework provides base test classes in `packages/valory/skills/abstract_round_abci/test_tools/` and `plugins/aea-test-autonomy/`.
@@ -316,6 +334,8 @@ if (
 2. Verify each such event appears in the round's entry in `transition_function`
 3. For rounds extending `CollectSameUntilThresholdRound` etc., check that `done_event`, `none_event`, `no_majority_event` (and `fail_event` for keeper rounds) are all wired in the transition function
 
+**Carve-out — framework-scheduled events:** Events scheduled by the framework via `event_to_timeout` (most commonly `ROUND_TIMEOUT`) are NOT emitted by `end_block()` — the framework dispatches them via Tendermint's `update_time()`. An event is `end_block`-emittable only if it is part of the round's base class event set (`done_event`, `none_event`, `no_majority_event`, `fail_event`) or returned explicitly by a custom `end_block` override. Do NOT flag a transition-function event as "missing from `end_block()`" if it is wired into `event_to_timeout` — that is the framework's job.
+
 **Why it matters:** A missing transition means the round never completes, hanging the entire agent service.
 
 #### C4: Dead Timeouts
@@ -330,6 +350,42 @@ if (
 **Why it matters:** Developers may believe a timeout is protecting against hangs, when in fact it never fires. This creates a false sense of safety.
 
 **Caveat:** Skills whose directory name starts with `test_` (e.g. `test_abci`, `test_solana_tx_abci`) are **test/scaffold skills** that exist to exercise the framework, not production FSM apps. Dead timeouts in these skills are often intentional test fixtures (e.g. testing that `SharedState.setup()` can wire `event_to_timeout` from params). Do NOT flag these as C4 findings. See also the False Positive Guidance section.
+
+#### C5: Non-Determinism in `end_block()`
+
+**What:** `AbstractRound.end_block()` must be a **pure function of consensus state**. Reading wall-clock time, the file system, environment variables, randomness, or external services produces a different event on different agents — silently breaking consensus. The Tendermint replica picks one of the divergent state transitions; the others are wedged or marked Byzantine.
+
+**Search patterns inside `end_block()` bodies (and any helper called from `end_block`):**
+- File I/O: `open(`, `Path.read_*`, `json.load(open(`, `os.path.exists(`, `pathlib.*`
+- Wall clock: `time.time()`, `datetime.now()`, `datetime.utcnow()`, `time.monotonic()` — use `synchronized_data.last_round_transition_timestamp` instead (set by the framework from agreed block time)
+- Randomness: `random.*`, `secrets.*`, `os.urandom(` — use `synchronized_data.most_voted_randomness` (consensus randomness)
+- Environment: `os.environ.get(`, `os.getenv(`
+- Network / external services: `requests.*`, `urlopen(`, `socket.*`, `subprocess.*`, contract calls, IPFS reads
+- Locale-dependent operations: `locale.*`, `re` patterns that depend on system locale
+
+**Bug example:**
+```python
+def end_block(self) -> Optional[Tuple[BaseSynchronizedData, Enum]]:
+    # BUG: each agent reads its own local file
+    with open("/tmp/benchmark_mode.json") as f:
+        config = json.load(f)
+    if config["mode"] == "benchmark":
+        return self.synchronized_data, Event.BENCHMARK
+    return self.synchronized_data, Event.NORMAL
+```
+Different agents may have different files → different events → consensus break.
+
+**Severity by service shape:**
+- **Multi-agent service:** Critical — consensus break.
+- **Single-agent / sovereign service** (e.g. some Pearl-shipped configurations): demote to **informational** but still surface — the bug appears the moment the service is promoted to multi-agent.
+
+**How to check:**
+1. Walk every `end_block()` body and any function it calls.
+2. Grep the imports of the round module for `time`, `datetime`, `random`, `os`, `requests`, `urllib`, `pathlib`.
+3. For each non-deterministic call found, verify whether the surrounding code path can reach `end_block()`.
+4. Check whether the agent's service config (look at `services/*/service.yaml` `number_of_agents`) is 1 or N to set severity.
+
+**Fix:** All non-deterministic data must be sourced from `synchronized_data` — i.e. agreed via consensus by the predecessor round's payload.
 
 ### High (H) — Issues with significant impact
 
@@ -422,6 +478,62 @@ class SynchronizedData(BaseSynchronizedData):
 ```
 
 **Why it matters:** The property silently returns data from a different round, causing logic errors or `KeyError` crashes.
+
+#### M6: SynchronizedData JSON Round-Trip Safety
+
+**What:** Writes to `synchronized_data` go through `AbciAppDB.update()` which JSON-serializes values. Values that are not JSON-native (`set`, `Decimal`, `dataclass` instances, `datetime`, `tuple` of mixed types, custom enum members) silently lose fidelity (e.g. `tuple` → `list`) or raise `TypeError` at write time. Custom serializer hooks (e.g. `EGreedyPolicy.serialize` / `deserialize`) must be applied symmetrically on the read side.
+
+Plus: when a `SynchronizedData` subclass uses **multiple inheritance** to combine data classes from composed skills (e.g. `class SynchronizedData(MechInteractSyncedData, MarketManagerSyncedData, TxSettlementSyncedData)`), two parents defining the same property name silently shadow according to MRO order. The subclass author may believe they are reading from skill A's data when they're actually reading from skill B's.
+
+**How to check:**
+1. For each round's `end_block()`, list keys + value-expressions written via `update(...)` or `update(synchronized_data_class=..., ...)`.
+2. For each value, infer the type (walk back from the call site, or look at the payload field type, or trace through `json.loads`). Flag values that are not JSON-native unless a custom serializer is documented at both write and read sides.
+3. For each `SynchronizedData` subclass with multiple base classes, walk the MRO and flag any property name defined in more than one parent — explicitly call out which parent wins.
+
+**Bug examples:**
+```python
+# BUG: set is not JSON-serializable — TypeError at write time
+self.synchronized_data.update(participants={agent1, agent2})
+
+# BUG: Decimal becomes float after JSON round-trip — silent precision loss
+self.synchronized_data.update(price=Decimal("0.123456789012345678"))
+
+# BUG: same property name in two parents
+class MechInteractSyncedData:
+    @property
+    def latest_event(self) -> str: return self.db.get_strict("mech_event")
+class MarketManagerSyncedData:
+    @property
+    def latest_event(self) -> str: return self.db.get_strict("market_event")
+class SynchronizedData(MechInteractSyncedData, MarketManagerSyncedData):
+    pass  # reads mech_event — likely surprising for someone reading the file top-down
+```
+
+**Fix:** Use lists/dicts/primitives or apply explicit serialize/deserialize hooks. Pick distinct property names across composed data classes, or override explicitly with documented intent.
+
+#### M7: `cross_period_persisted_keys` Completeness for Post-Reset Initial Reads
+
+**What:** After a period reset (`ResetAndPauseRound`), the next period restarts at the chain's initial round. Any DB key that the initial round chain (or its behaviours) reads via `self.db.get` / `self.db.get_strict` must be in `cross_period_persisted_keys` — otherwise it's missing after the first reset and the read returns `None` / raises `KeyError`. The bug is invisible in the first period (the key was just written) and surfaces only after the reset.
+
+**How to check:**
+1. Identify the initial round chain after a reset (typically `RegistrationRound` → ... → first business round). Use the composed app's `transition_function` to follow `Event.RESET_TIMEOUT` / `Event.DONE` from `ResetAndPauseRound` to the next round.
+2. Statically collect every `self.db.get(...)` / `self.db.get_strict(...)` and every `self.synchronized_data.<property>` access in those rounds and their behaviours. Resolve property accesses to the underlying DB key.
+3. Subtract `cross_period_persisted_keys`. The diff is the set of keys that will be missing after the first reset.
+
+**Why it matters:** First-period success masks the bug. Production crashes appear only after the first natural reset (often hours or days into a deployment).
+
+#### M8: Misleading Event-Shaped Class Attributes on Round Subclasses
+
+**What:** On `CollectSameUntilThresholdRound` and siblings, the framework's `end_block()` consults a fixed set of class attributes: `done_event`, `none_event`, `no_majority_event`, `collection_key`, `selection_key` (and `fail_event` for keeper rounds — see `extended_requirements`). Defining additional class attributes whose RHS is an `Event` enum value (e.g. `withdrawal_initiated: Event = Event.WITHDRAWAL_INITIATED`) does NOT wire that event into the round — the parent's `end_block()` never reads it.
+
+Combined with a `transition_function` entry for that event, this produces a "looks-like-config-but-isn't" footgun. The current C3 check catches the dead transition. M8 catches the misleading attribute that gives the false impression the round will emit it.
+
+**How to check:**
+1. For each `CollectSameUntilThresholdRound` (or sibling) subclass, list class-level attributes whose RHS is an `Event` enum value.
+2. Subtract the names actually consulted by the parent class (`done_event`, `none_event`, `no_majority_event`, `fail_event`, `collection_key`, `selection_key`) and any name read by a custom `end_block` override in the subclass.
+3. The remainder is dead config. Suggested fix: either implement a custom `end_block()` that uses the attribute, or remove it.
+
+**Why it matters:** Combined with a wired transition, the dead attribute makes a reader assume mid-round side transitions exist (e.g. "withdrawal can interrupt this round") when they don't — leading to design decisions based on capabilities the FSM doesn't actually have.
 
 ### Test (T) — Test correctness and coverage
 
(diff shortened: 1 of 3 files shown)


claude-skills/audit-fsm/SKILL.md
```diff
@@ -331,6 +351,42 @@ if (
...
 
+#### C5: Non-Determinism in `end_block()`
+
+**What:** `AbstractRound.end_block()` must be a **pure function of consensus state**. Reading wall-clock time, the file system, environment variables, randomness, or external services produces a different event on different agents — silently breaking consensus. The Tendermint replica picks one of the divergent state transitions; the others are wedged or marked Byzantine.
+
+**Search patterns inside `end_block()` bodies (and any helper called from `end_block`):**
+- File I/O: `open(`, `Path.read_*`, `json.load(open(`, `os.path.exists(`, `pathlib.*`
+- Wall clock: `time.time()`, `datetime.now()`, `datetime.utcnow()`, `time.monotonic()` — use `synchronized_data.last_round_transition_timestamp` instead (set by the framework from agreed block time)
```
LOCKhart07: Factual error: `synchronized_data.last_round_transition_timestamp` does not exist on `BaseSynchronizedData`. The property `last_round_transition_timestamp` is defined on `RoundSequence` (`packages/valory/skills/abstract_round_abci/base.py:3487`) and is the agreed Tendermint block time of the *previous* transition (set from `self._blockchain.last_block.timestamp`). From inside `end_block()` it is reachable via `self.context.state.round_sequence.last_round_transition_timestamp`, not via `synchronized_data`.

The "Fix" line at the end of C5 already states the right pattern (consensus-agreed via predecessor round's payload). But this inline reference will get quoted verbatim in findings and produce fix recommendations that don't compile.

Suggested rewording: ```
- Wall clock: `time.time()`, `datetime.now()`, `datetime.utcnow()`, `time.monotonic()` — use `self.context.state.round_sequence.last_round_transition_timestamp` (agreed Tendermint block time of the last transition) or read a timestamp written into a payload field by a predecessor round (consensus-agreed, then surfaced via `synchronized_data`).
```

File: claude-skills/audit-fsm/SKILL.md [hunk 2]
```diff
@@ -176,6 +176,24 @@ Downstream projects compose these library skills with their own custom skills vi
...
+### `is_abstract: true` Semantics
+
+A skill with `is_abstract: true` in its `skill.yaml` is composition-only — it is never instantiated as a top-level skill, only consumed by another `chain()` call. Audit checks branch on this flag:
+
+- **Skip** "missing `fsm_specification.yaml`" findings — abstract skills are composed-only and don't need standalone wiring.
+- **Skip** "missing dialogue X" findings if the dialogue is provided by the composing skill.
+- **Demote** "missing handler" findings to informational unless the handler is required for composition (e.g. an `ABCIHandler` is mandatory).
+- **Apply** the same M2 / C4 carve-outs as library skills (unused events / dead timeouts may be intentional extension points).
+
+The `agent_performance` "missing dialogues" / "missing fsm_specification" findings from prior audits worked out only because the CLI tool happened to fail. With the `is_abstract` branch, those findings are correctly suppressed at the source.
```
LOCKhart07: Minor wording: "worked out only because the CLI tool happened to fail" reads ambiguously — sounds like the findings were the desired outcome. Suggested rephrase:

> Prior audits avoided false positives here only by accident — the CLI tool failed before reaching these checks. The `is_abstract` branch now makes the suppression deliberate.

claude-skills/audit-fsm/SKILL.md
```diff
@@ -331,6 +351,42 @@ if (
...
+        config = json.load(f)
+    if config["mode"] == "benchmark":
+        return self.synchronized_data, Event.BENCHMARK
+    return self.synchronized_data, Event.NORMAL
+```
+Different agents may have different files → different events → consensus break.
+
+**Severity by service shape:**
+- **Multi-agent service:** Critical — consensus break.
+- **Single-agent / sovereign service** (e.g. some Pearl-shipped configurations): demote to **informational** but still surface — the bug appears the moment the service is promoted to multi-agent.
```
OjusWiZard: **Severity bucket undefined.** "Informational" is not in the C/H/M/L taxonomy used by the report template (`Critical Findings` / `High Findings` / `Medium Findings` / `Low Findings`), so an auditor has nowhere to put this finding and will silently drop it. Pearl-shipped services get promoted to multi-agent, so silent loss is the worst outcome here.

Suggest:

> **Single-agent / sovereign service** (e.g. some Pearl-shipped configurations): demote to **Low (L)** with a note: "severity escalates to Critical when the service runs with N>1 agents."

This keeps the finding in the Low section of the standard report instead of vanishing.

claude-skills/audit-fsm/SKILL.md
```diff
@@ -331,6 +351,42 @@ if (
...
+
+**How to check:**
+1. Walk every `end_block()` body and any function it calls.
+2. Grep the imports of the round module for `time`, `datetime`, `random`, `os`, `requests`, `urllib`, `pathlib`.
+3. For each non-deterministic call found, verify whether the surrounding code path can reach `end_block()`.
+4. Check whether the agent's service config (look at `services/*/service.yaml` `number_of_agents`) is 1 or N to set severity.
```
OjusWiZard: **C5 misses non-determinism in `process_payload()`.** `end_block()` reads `self.synchronized_data`, which is populated by `process_payload()`. A `time.time()` / `os.environ` write inside `process_payload()` produces a non-deterministic `end_block()` *read* even though the `end_block()` body itself is clean. Auditors following step 1 literally will report a clean C5 on a real consensus break.

Suggest adding a step (or extending step 1):

> Also walk every `process_payload()` body in the same round and flag non-deterministic writes to `self.synchronized_data` — those propagate into `end_block()` reads and break consensus equally.

claude-skills/audit-fsm/SKILL.md
```diff
@@ -423,6 +479,62 @@ class SynchronizedData(BaseSynchronizedData):
...
+2. For each value, infer the type (walk back from the call site, or look at the payload field type, or trace through `json.loads`). Flag values that are not JSON-native unless a custom serializer is documented at both write and read sides.
+3. For each `SynchronizedData` subclass with multiple base classes, walk the MRO and flag any property name defined in more than one parent — explicitly call out which parent wins.
+
+**Bug examples:**
+```python
+# BUG: set is not JSON-serializable — TypeError at write time
+self.synchronized_data.update(participants={agent1, agent2})
+
+# BUG: Decimal becomes float after JSON round-trip — silent precision loss
+self.synchronized_data.update(price=Decimal("0.123456789012345678"))
```
OjusWiZard: **The `Decimal` example mislabels the failure mode.** `json.dumps(Decimal("0.1"))` raises `TypeError` — it does not silently round-trip to a float. So this example fails the same way as the `set` example one line above (write-time crash), not via "silent precision loss."

Verified:
```
>>> import json; from decimal import Decimal
>>> json.dumps(Decimal("0.1"))
TypeError: Object of type Decimal is not JSON serializable
```

An auditor following this rule will mis-classify every `Decimal`-write as silent precision loss and may suggest "convert to `float` before storing" — exactly the wrong direction.

Fix one of:

```python
# BUG: Decimal is not JSON-serializable — TypeError raised at write time
self.synchronized_data.update(price=Decimal("0.123456789012345678"))
```

or, if you want a true silent-precision-loss example, use `float`:

```python
# BUG: float arithmetic loses precision; survives JSON but the value already drifted
self.synchronized_data.update(price=0.1 + 0.2)  # 0.30000000000000004
```

claude-skills/audit-fsm/SKILL.md
```diff
@@ -423,6 +479,62 @@ class SynchronizedData(BaseSynchronizedData):
...
+```
+
+**Fix:** Use lists/dicts/primitives or apply explicit serialize/deserialize hooks. Pick distinct property names across composed data classes, or override explicitly with documented intent.
+
+#### M7: `cross_period_persisted_keys` Completeness for Post-Reset Initial Reads
+
+**What:** After a period reset (`ResetAndPauseRound`), the next period restarts at the chain's initial round. Any DB key that the initial round chain (or its behaviours) reads via `self.db.get` / `self.db.get_strict` must be in `cross_period_persisted_keys` — otherwise it's missing after the first reset and the read returns `None` / raises `KeyError`. The bug is invisible in the first period (the key was just written) and surfaces only after the reset.
+
+**How to check:**
+1. Identify the initial round chain after a reset (typically `RegistrationRound` → ... → first business round). Use the composed app's `transition_function` to follow `Event.RESET_TIMEOUT` / `Event.DONE` from `ResetAndPauseRound` to the next round.
```
OjusWiZard: **Wrong event name — `RESET_TIMEOUT` is the failure path, not the normal post-reset re-entry.** In `reset_pause_abci`, `Event.RESET_TIMEOUT` from `ResetAndPauseRound` routes to `FinishedResetAndPauseErrorRound` (error recovery). Normal re-entry into the chain's initial business round is on `Event.DONE` only.

An agent following this literally will trace the error-recovery path, miss the actual post-reset initial-read set, and either flag bogus keys or miss the real ones.

Suggest:

> Use the composed app's `transition_function` to follow `Event.DONE` from `ResetAndPauseRound` to the next round.

If the failure path also matters for cross-period analysis, describe it separately and label it as the failure case.

claude-skills/audit-fsm/SKILL.md
```diff
@@ -423,6 +479,62 @@ class SynchronizedData(BaseSynchronizedData):
...
+2. Statically collect every `self.db.get(...)` / `self.db.get_strict(...)` and every `self.synchronized_data.<property>` access in those rounds and their behaviours. Resolve property accesses to the underlying DB key.
+3. Subtract `cross_period_persisted_keys`. The diff is the set of keys that will be missing after the first reset.
```
OjusWiZard: **Will produce false positives in every audit — `default_cross_period_keys` not subtracted.** The framework auto-injects defaults regardless of what the skill declares. From `packages/valory/skills/abstract_round_abci/base.py:549`:

```python
self._cross_period_persisted_keys = self.default_cross_period_keys.union(
    cross_period_persisted_keys or frozenset()
)
```

Defaults include `all_participants`, `participants`, `consensus_threshold`, `safe_contract_address`. Any skill that reads e.g. `safe_contract_address` after a reset without listing it in `cross_period_persisted_keys` will be flagged — but it works in production.

Suggest extending step 3:

> 3. Subtract `cross_period_persisted_keys` **and `AbciAppDB.default_cross_period_keys`** (`all_participants`, `participants`, `consensus_threshold`, `safe_contract_address`). The latter are injected by the DB layer unconditionally and don't need to appear in the skill's own declaration.

OjusWiZard (commented):
Inline comments on the new C5/M6/M7 sections — five issues that will cause auditors to either misclassify findings or trace the wrong code path. Other findings will follow in a separate pass.

DavidMinarsch:
Pushed 5 follow-up commits (`1e65485d0..70940771a`) addressing review feedback and additional findings from continued audit-skill use. Diff is markdown-only; CI was previously cancelled mid-run by mistake (pylint had already passed at 10.00/10).

## What's new

### Review-comment fixes (`1e65485d0`)

@LOCKhart07's two inline comments:

- **Factual error in C5** — `last_round_transition_timestamp` does not exist on `BaseSynchronizedData`. From inside `AbstractRound.end_block()` it is reached via `self.context.state.round_sequence.last_round_transition_timestamp` (a `RoundSequence` property at `packages/valory/skills/abstract_round_abci/base.py:3487`, fed from `self._blockchain.last_block.timestamp` at line 3801). The framework dogfoods this exact access path — `PendingOffencesRound.offence_status` returns `self.context.state.round_sequence.offence_status` (`base.py:3889`) from inside its own `end_block`. Same factual fix applied to audit-resilience BP22.
- **Wording nit** in the `is_abstract: true` semantics section — replaced ambiguous "worked out only because the CLI tool happened to fail" with explicit phrasing per the suggestion.

### Sub-agent verification guardrail (`11c5b6b9b`)

Real audit runs surfaced an over-classification pattern: discovery sub-agents return candidate findings with candidate severities, and the spawning agent was treating those classifications as final. Concrete failures:

- `initial_behaviour_cls` flagged as Critical (should be L)
- "dead comment" flagged as Critical (should be L)
- One AgentDB accessor bug split into 3 separate Criticals (should consolidate to 1 High)

Adds an explicit guardrail at the top of the Methodology Guardrails section in both skills: the spawning agent MUST hand-verify each sub-agent finding against source before accepting severity. Five steps — read the cited line, confirm the bug, re-evaluate severity, consolidate split findings, demote liberally. *Sub-agent classification is input, not output.*

### Dev-package filtering for unscoped CLI tools (`bb7253a16`)

`autonomy analyse handlers` / `dialogues` / `docstrings` walk every package under `packages/` including framework-vendored ones in the `third_party` section of `packages.json`. Findings against `third_party` belong to the upstream repo, not this audit, so they were producing noise.

Surgical fix: filter only the unscoped CLI tools' **output** against the dev set built from `jq -r '.dev | keys[]' packages.json`. Manual checks, composition analysis (`chain()`), and cross-skill data-flow checks (M3 / M7 / M8) still need full visibility into library skills, so the discovery step and rest of the audit are deliberately unchanged.

### C6 unmapped finals + H4 `extended_requirements = ()` footgun (`b4e759b05`)

Two findings from real runs that the existing checklist either buried or didn't surface:

- **C6 — Unmapped Reachable Finals in Composition.** Splits out the "every reachable inner final has a mapping entry" half of the old H2. The other half ("every mapping key/value valid") is framework-validated by `chain()` at construction time and produces zero new findings, which was masking the still-valuable unmapped-finals check. Real audits found 2 unmapped finals in one downstream `composition.py` — highest-impact FSM finding from that run, almost slipped through under H2's bundled framing. C-level because the blast radius matches C3 (round hang = whole agent hang).
- **H4 — `extended_requirements = ()` Override Without Compensating `end_block`.** `_MetaAbstractRound` (`base.py:1066-1068`) uses `extended_requirements` to verify required class attributes exist at class-creation time. Setting it to `()` disables the check — legitimate when the subclass overrides `end_block`, footgun when the subclass inherits the parent's `end_block` AND omits an attribute it reads. Result: AttributeError on first `end_block` call → behaviour crash → agent crash. AgentDBRound was the canonical real-world instance.
- **H2 trimmed** to a redirect/index entry pointing at C6 / M7 / metaclass. Bullets 1-2 are framework-validated; do-not-duplicate notice added explicitly.

### BP23 MagicMock-without-spec + DRY `MAX_WORKER_THREADS` (`70940771a`)

- **BP23 — `MagicMock` without `spec=` hides production-only bugs.** The motivating case: an AgentDB `response.text` access bug went undetected in tests because `MagicMock(text="...")` auto-vivifies any attribute access. Tests pass, prod crashes. The check applies to every `MagicMock` / `Mock` / `AsyncMock` substituting for an HTTP response, SDK client, or connection object — flag if no `spec=` argument is set. Severity rule: pull through to P0/P1 if the underlying masked bug is high-impact; otherwise P3 (test hygiene).
- **DRY the MAX_WORKER_THREADS=1 + no-timeout warning.** Same content was duplicated across the threading-model architecture section, BP8, the Step 3 stuck-identification checklist, the priority rules, and the quality checklist. Made the architecture section the canonical source under a `Canonical: MAX_WORKER_THREADS=1 + no-timeout call = subsystem outage` anchor; rewrote the other locations to reference it. BP8 keeps its actionable bits but no longer repeats the httpx/requests examples.

## Final layout

- audit-fsm: C1–C6, H1–H4, M1–M8, T1–T6, L1–L5
- audit-resilience: BP1–BP23, CC1–CC9
- Both skills now have the sub-agent verification guardrail at the top of their Methodology Guardrails sections.
- audit-fsm CLI step now filters output to dev packages only.

PR #2499 (security-review skill) is stacked on this branch; once this merges, that one rebases automatically.

DavidMinarsch:
Thanks @OjusWiZard — your five inline comments were missed in the first review pass (I'd only addressed @LOCKhart07's two). Pushed `eed80af4f` to fix all five. Each one was a real correctness issue that would have caused auditors to either misclassify findings or trace the wrong code path; framework claims verified against source before editing.

### C5 — "Informational" not in C/H/M/L taxonomy (line 380)
Demoted single-agent / sovereign to **L** with an explicit escalation note: *"severity escalates to Critical when the service runs with N>1 agents."* Pearl-shipped services regularly get promoted to multi-agent, so silent loss is the worst outcome here.

### C5 — misses non-determinism in `process_payload()` (line 386)
Extended the check procedure with a new step: walk every `process_payload()` body in the same round. `process_payload()` writes to `self.synchronized_data`; `end_block()` reads it. A non-deterministic write in `process_payload()` produces a non-deterministic `end_block()` read even when the `end_block()` body itself is clean — auditors who walked only `end_block()` literally would have reported clean C5 on a real consensus break.

### M6 — Decimal example mislabeled (line 568)
Verified: `json.dumps(Decimal("0.1"))` raises `TypeError`, same failure mode as `set`. Relabeled the Decimal example as a write-time crash and added a **separate** genuine silent-precision-loss example using float arithmetic (`0.1 + 0.2` → `0.30000000000000004`). The audit now has a correct example of each failure mode, not a conflated one that would push auditors toward "convert to float" — exactly the wrong direction.

### M7 — wrong event name (line 588)
Verified against `packages/valory/skills/reset_pause_abci/rounds.py:94-98`:
```python
ResetAndPauseRound: {
    Event.DONE: FinishedResetAndPauseRound,                  # ← normal exit
    Event.RESET_AND_PAUSE_TIMEOUT: FinishedResetAndPauseErrorRound,  # failure
    Event.NO_MAJORITY: FinishedResetAndPauseErrorRound,             # failure
}
```
Fixed step 1 to follow `Event.DONE` for the normal post-reset re-entry path, with an explicit note that `Event.RESET_AND_PAUSE_TIMEOUT` and `Event.NO_MAJORITY` are failure paths and tracing them surfaces bogus keys / misses real ones. Also corrected the event name (`RESET_AND_PAUSE_TIMEOUT`, not `RESET_TIMEOUT`).

### M7 — `default_cross_period_keys` not subtracted (line 590)
Verified against `packages/valory/skills/abstract_round_abci/base.py:516-523` then `:549`:
```python
default_cross_period_keys: FrozenSet[str] = frozenset({
    "all_participants",
    "participants",
    "consensus_threshold",
    "safe_contract_address",
})
# ...
self._cross_period_persisted_keys = self.default_cross_period_keys.union(
    cross_period_persisted_keys or frozenset()
)
```
Updated step 3 to subtract **both** `cross_period_persisted_keys` AND `AbciAppDB.default_cross_period_keys`. Without this, every audit would flag false positives on any skill that reads `safe_contract_address` / `all_participants` / `participants` / `consensus_threshold` after a reset — those are auto-injected by the DB layer regardless of declaration.

---

**Note on CI:** the `gitleaks` failure on the latest run is from commit `9a7d839` on the stacked PR #2499 (security-review skill) — gitleaks fetches all branches with `fetch-depth: 0` and matched a literal `xoxb-` example in the search-pattern documentation of the new security-review skill. That false positive will be cleaned up on the #2499 branch (the literal prefixes were redundant with gitleaks defaults anyway). Doesn't affect the content of this PR.

LOCKhart07 (dismissed):

claude-skills/audit-fsm/SKILL.md
```diff
@@ -423,6 +549,67 @@ class SynchronizedData(BaseSynchronizedData):
...
+1. For each round's `end_block()`, list keys + value-expressions written via `update(...)` or `update(synchronized_data_class=..., ...)`.
+2. For each value, infer the type (walk back from the call site, or look at the payload field type, or trace through `json.loads`). Flag values that are not JSON-native unless a custom serializer is documented at both write and read sides.
+3. For each `SynchronizedData` subclass with multiple base classes, walk the MRO and flag any property name defined in more than one parent — explicitly call out which parent wins.
+
+**Bug examples:**
+```python
+# BUG (write-time crash): set is not JSON-serializable — TypeError at write time
+self.synchronized_data.update(participants={agent1, agent2})
+
+# BUG (write-time crash): Decimal is not JSON-serializable — TypeError at write time
```
DhairyaPatel7: Both examples say `TypeError at write time`, but the actual flow is:

`AbciAppDB.update()` (`base.py:658`) → `self.validate(kwargs)` (`base.py:645–655`) → `is_json_serializable(...)` (`utils.py:408–417`, only accepts primitives + tuple/list/dict recursively) → returns `False` for both `set` and `Decimal` → `validate` raises **`ABCIAppInternalError("`AbciAppDB` data must be json-serializable …")`**, not Python's `TypeError`.

The eed80af fix corrected the failure-mode label (silent precision loss → write-time crash) but left the exception class wrong. Auditors using these examples as a reference will write findings citing the wrong exception, and reviewers won't be able to confirm them from a traceback.

Suggest replacing both occurrences of `TypeError at write time` with `ABCIAppInternalError at write time (raised by AbciAppDB.validate, base.py:649–655)`.

claude-skills/audit-fsm/SKILL.md
```diff
@@ -331,6 +351,78 @@ if (
...
+```python
+class InnerAppA(AbciApp):
+    final_states = {DoneRound, FailedRound, ImpossibleRound}
+
+# In composition.py:
+abci_app_transition_mapping = {
+    FinishedAppARound: SomeInitialRound,
+    # BUG: FailedRound and ImpossibleRound are reachable finals of InnerAppA
+    #      but absent from this mapping AND absent from composed_app.final_states
+}
```
DhairyaPatel7, line 418: The bug example is internally inconsistent: `InnerAppA.final_states = {DoneRound, FailedRound, ImpossibleRound}` but the mapping key is `FinishedAppARound`, which is not in that set. Per L383 of this skill — *"the framework validates that mapping keys ARE final states of inner apps and `chain()` raises `ABCIAppInternalError` at construction time on violations"* — this mapping would not survive construction, so the unmapped-finals bug being demonstrated never reaches runtime.

Either rename the mapping key to one of `InnerAppA`'s actual finals:

```python
abci_app_transition_mapping = {
    DoneRound: SomeInitialRound,
    # BUG: FailedRound and ImpossibleRound are reachable finals of InnerAppA
    #      but absent from this mapping AND absent from composed_app.final_states
}
```

…or add a one-liner noting that `FinishedAppARound` is a final of a *different* inner app in the chain, so the reader doesn't have to guess. As-is, the highest-leverage check in the diff ships with a reference example that doesn't parse against its own rules.

claude-skills/audit-resilience/SKILL.md
```diff
@@ -486,6 +495,194 @@ response, error = self._route_request(payload=payload)
...
+
+**Fix:** invert to the conservative outcome (`sufficient_funds = False`, `proceed = False`); OR model an explicit third state (`enum: Sufficient | Insufficient | Unknown`) so callers can distinguish "unable to check" from "checked and OK".
+
+**Why it matters:** This pattern silently turns an upstream observability gap into a downstream financial error. The agent appears healthy; the failure surfaces at the counterparty side, with no audit trail upstream.
+
+### BP19: TLS Verification Disabled / Hardcoded Credentials
+
+Grep-level checks. Each finding is independent; flag every match.
+
+- **`verify=False`** on `requests.*` / `httpx.*` / `aiohttp.*` — disables TLS certificate verification, opens MITM. Search: `verify=False`, `verify = False`, `ssl=False`, `ssl_context=None` in HTTP-call sites.
```
DhairyaPatel7: `ssl_context=None` is the **default** for `httpx.Client(...)` / `httpx.AsyncClient(...)` and is equivalent to omitting the argument — it means "use the default SSL context," which validates certificates against the system trust store. Grepping for it will surface every explicit-default usage and flag them all as TLS-disabled.

Drop `ssl_context=None` from the search list; the genuinely-dangerous forms are `verify=False` / `ssl=False` (and for aiohttp, `ssl=False` or an `ssl.SSLContext(...)` with `check_hostname = False` / `verify_mode = ssl.CERT_NONE`). If a stronger `ssl_context` signal is wanted, search for `ssl.create_default_context()` followed by attribute mutations that disable verification — but the raw `=None` literal is a false positive generator.

OjusWiZard (dismissed):
All five inline comments from my 2026-05-07 pass are addressed in HEAD — severity bucket re-routed to L, C5 walk extended to process_payload, Decimal example correctly labeled as TypeError with a separate float example for silent precision loss, M7 post-reset re-entry now follows Event.DONE with the failure-path contrast called out, and step 3 subtracts both cross_period_persisted_keys and AbciAppDB.default_cross_period_keys with source pointers. LGTM.

DavidMinarsch:
Thanks @DhairyaPatel7. Pushed `c6a91852b` addressing all three. Each was source-verified against framework code before editing.

### M6 — wrong exception class (audit-fsm:568)

You're right — my previous fix verified Python-level `json.dumps(Decimal)` → `TypeError` but missed the framework's validation layer that intercepts first. Traced the actual flow:

```
AbciAppDB.update()       base.py:658-660  → self.validate(kwargs)
AbciAppDB.validate()     base.py:645-655  → is_json_serializable(...)
is_json_serializable()   utils.py:408-417 → False for set, Decimal
                                            (only primitives + tuple/list/dict
                                             recursively are accepted)
validate raises          ABCIAppInternalError(
                           "`AbciAppDB` data must be json-serializable..."
                         )
```

So auditors see `ABCIAppInternalError`, not `TypeError`. Both example labels updated; the M6 "What" paragraph rewritten to spell out the validation chain with explicit `file:line` citations so future auditors can confirm from a traceback.

### C6 — self-inconsistent example (audit-fsm:415)

Confirmed: `InnerAppA.final_states = {DoneRound, FailedRound, ImpossibleRound}` did not include `FinishedAppARound`, so the mapping `{FinishedAppARound: SomeInitialRound}` would have failed `chain()`'s own validation at construction time — meaning the C6 bug being demonstrated could never reach runtime. Self-contradiction against my own rule.

Renamed the mapping key to `DoneRound` (one of the actual finals) and added an inline comment clarifying what `chain()` does and doesn't enforce — so the example now demonstrates exactly C6's gap (a mapping that survives construction but leaves `FailedRound`/`ImpossibleRound` reachable and unmapped).

### BP19 — `ssl_context=None` false positive (audit-resilience:587)

Confirmed: `ssl_context=None` is the `httpx.Client` / `httpx.AsyncClient` default, equivalent to omitting the argument — it uses the system default SSL context which validates against the trust store. Grepping for it would have surfaced every explicit-default usage as a finding.

Dropped from the BP19 search list and replaced with the concrete dangerous forms per stack:
- `requests` / `httpx`: `verify=False` (also `verify = False`)
- `aiohttp`: `ssl=False`, OR a custom `ssl.SSLContext(...)` with `check_hostname=False` / `verify_mode=ssl.CERT_NONE`
- `ssl.create_default_context()` followed by attribute mutations that disable verification

Explicit note added that `ssl_context=None` is NOT a finding.

---

Self-critique on M6: this is the second time a reviewer caught a framework-claim error in my work that source-tracing would have caught. The methodology guardrail I added in `11c5b6b9b` says "verify against source before flagging" — I applied it to the original C5/M7/H4 work but on `eed80af4f` I only Python-verified the failure mode without tracing through `AbciAppDB.validate`. Will be more disciplined about following the chain through to the framework layer on the next pass.

PR #2500, opened 2026-05-11
chore: add bump-versions Claude skill

DhairyaPatel7:
## Summary

Adds a new Claude Code skill, `bump-versions`, under `claude-skills/bump-versions/SKILL.md`.

End-to-end procedure for bumping `open-autonomy` and `open-aea` (and their plugins / upstream-pin tags) to the latest released versions across any Valory downstream repo (mech, mech-interact, mech-server, mech-predict, mech-agents-fun, trader, optimus, market-creator, market-resolver, meme-ooorr, kv-store, funds-manager, genai, etc.).

## What the skill covers

1. **Phase 0 — Ground truth**: reads current OA/OAEA pins from `pyproject.toml` and the `[tool.tomte] upstream_pins` list, so the same skill works for OA-only repos and multi-upstream repos without per-repo branching.
2. **Phase 1 — Resolve targets**: latest releases via `gh release list`; verifies OA-at-target pins OAEA-at-target; resolves compatible release tags for each upstream pin with transitive-consistency check; explicit plugin-coupling table (`open-aea-test-autonomy` and `open-aea-helpers` track **OA**, not OAEA).
3. **Phase 2 — Breaking changes**: fetches `HISTORY.md` + `docs/upgrading.md` for both frameworks; classifies each change (pin-only / YAML / API / CI / removed / no-op).
4. **Phase 3 — Bump pins everywhere**: `pyproject.toml` (PEP 621), `tox.ini`, all workflow files (incl. `release.yaml`), dev YAML `dependencies:` blocks, plus easy-miss sites (`deployments/Dockerfiles/**/requirements.txt`, docs, Pipfile, scripts).
5. **Phase 4 — Code changes (OA/OAEA only)**: scoped strictly to dev-package paths from `packages.json` + top-level non-package code; explicit "do NOT edit third-party files".
6. **Phase 5 — Hash sync**: pre-audit third-party CID drift, then `autonomy packages sync --source` against OAEA, OA, and every upstream-pin entry; documents the cache/recovery error modes.
7. **Phase 6 — Lock + verify**: full local CI per every workflow job, not a sampled subset. Documents recurring failure modes (specifier conflict, `check-dependencies` parity, `liccheck` UNKNOWN, pytest-plugin autoload crash, etc.).
8. **Phase 7 — Stop**: hands off to the user; does not commit, push, or open a PR.

## Hard rules baked in

- OA + OAEA drive code changes; everything else in `upstream_pins` is version+hash-only.
- Edit `.py` only under dev-package paths; third-party files are refreshed by sync.
- `open-aea-test-autonomy` and `open-aea-helpers` track **OA** (not OAEA), despite the name.
- Run `autonomy packages lock` after every YAML edit; hashes cascade.

## Changed files

- `claude-skills/bump-versions/SKILL.md` — new skill (579 lines)

## Test plan

- [ ] Skill renders correctly in Claude Code (`/bump-versions` invocation)
- [ ] Dry-run the skill against a real downstream bump and confirm the phases match the artifacts produced

Co-Authored-By: Claude Opus 4.7 (1M context) <noreply@anthropic.com>


claude-skills/bump-versions/SKILL.md
```diff
@@ -0,0 +1,579 @@
...
+from pathlib import Path
+
+local = json.load(open('packages/packages.json'))['third_party']
+
+# Upstreams to diff against. Read from [tool.tomte] upstream_pins + the two frameworks.
+# (Substitute real local checkout paths or `gh api` fetches as available.)
+UPSTREAMS = [
+    # (local_checkout, ref)
+    ('/Users/dhairya/Desktop/Work/Valory/Github/open-aea',       'v<NEW_OAEA>'),
+    ('/Users/dhairya/Desktop/Work/Valory/Github/open-autonomy',  'v<NEW_OA>'),
```
jmoreira-valory: **Personal absolute paths leak into the skill — fails on copy-paste for any user other than @DhairyaPatel7.**

Lines 394-395 hardcode the author's local machine layout:

```python
UPSTREAMS = [
    ('/Users/dhairya/Desktop/Work/Valory/Github/open-aea',       'v<NEW_OAEA>'),
    ('/Users/dhairya/Desktop/Work/Valory/Github/open-autonomy',  'v<NEW_OA>'),
    ...
]
```

Anyone running the skill from this snippet will get `WARN: ... not present locally, skipping` for both upstreams and silently skip the drift audit — defeating Phase 5.1's whole point.

Same issue at line 17 in the intro:

> Reference: a Poetry-era predecessor of this skill lives at `~/Desktop/Work/Valory/Github/oa-bump/SKILL.md`.

Replace both with placeholders that don't reference a specific user's filesystem:

```python
import os
UPSTREAMS = [
    (os.environ.get("LOCAL_OAEA_CHECKOUT", os.path.expanduser("~/git/open-aea")),       'v<NEW_OAEA>'),
    (os.environ.get("LOCAL_OA_CHECKOUT",   os.path.expanduser("~/git/open-autonomy")),  'v<NEW_OA>'),
    ...
]
```

…and drop the absolute-path reference in line 17 (or generalize to *"a Poetry-era predecessor exists in the legacy `oa-bump` skill"* without naming a filesystem path). The skill ships under `claude-skills/` in the open-autonomy repo — leaking the author's `~/Desktop/Work/Valory/Github/` layout publicly is unnecessary.

claude-skills/bump-versions/SKILL.md
```diff
@@ -0,0 +1,579 @@
+---
+name: bump-versions
+description: Bump `open-autonomy` and `open-aea` (and their plugins / upstream-pin tags) to the latest released versions across any Valory downstream repo. Reads HISTORY.md + upgrading.md for both frameworks, applies code changes to dev packages only for OA/OAEA breaking changes, bumps version pins everywhere, then syncs third-party hashes from all upstream repos.
+argument-hint: "[oa=<version>] [oaea=<version>]  # both optional; defaults to latest GitHub release of each"
+disable-model-invocation: false
```
jmoreira-valory: **`disable-model-invocation: false` is risky for a 7-phase destructive operation.**

This frontmatter setting lets Claude auto-invoke the skill based on conversation context. Given Phase 3-6 mutate `pyproject.toml`, every workflow file, every dev YAML, every `packages/<author>/*.yaml`, run `autonomy packages lock`, sync from external sources, and clear the local AEA cache (Phase 5.3: `find ~/.aea/cache/packages -mindepth 1 -maxdepth 1 -type d -empty -exec rmdir`), an accidental auto-invocation against the wrong repo or wrong branch is a high-blast-radius outcome.

The `argument-hint` indicates this is designed for explicit invocation anyway (`[oa=<version>] [oaea=<version>]`). Other heavyweight valory skills follow the explicit-only pattern — see `oa-propel`'s deploy guidance which is reference-only for the agent.

Suggest flipping to `disable-model-invocation: true` so users have to type `/bump-versions` explicitly. Documented in the user-facing help, Claude doesn't accidentally trigger it when a user says *"check if this repo is on the latest open-autonomy"* in an unrelated conversation.

claude-skills/bump-versions/SKILL.md
```diff
@@ -0,0 +1,579 @@
...
+```bash
+# 1. Regenerate the lockfile and install
+uv lock
+uv sync --all-groups
+
+# 2. Re-lock packages (third-party hashes cascade into dev fingerprints)
+autonomy packages lock
+
+# 3. Auto-format (some pin changes trip isort/black)
+tox -e black
```
jmoreira-valory: **Phase 6 verify commands use bare `tox -e` form throughout, inconsistent with Phase 3.2's `tomte tox -e` acknowledgment.**

Phase 3.2 correctly calls out:

> Repos using `tomte tox -e <env>` instead of bare `tox -e <env>` inherit dep blocks from tomte's canonical config — many `[deps-*]` blocks may be thin or absent. Check the Makefile to see which form this repo uses.

But every Phase 6 command (lines 479-516) uses uniform `tox -e <env>`. A user running this skill on a wave-2 fleet repo (market-creator, market-resolver, kv-store, funds-manager, genai, mech, mech-interact, etc. — most of the fleet at this point) would hit cryptic `ERROR: env 'black' not defined in tox config` failures on every step because tomte's canonical `tox.ini` ships the envs and the bare-tox install doesn't see them.

Two clean ways to fix:

1. **One-liner at the top of Phase 6**:
   ```bash
   # If the repo's Makefile uses `tomte tox`, prefix every command below with `tomte`.
   TOX=$(grep -l "tomte tox" Makefile 2>/dev/null && echo "tomte tox" || echo "tox")
   ```
   …then use `$TOX -e black` etc.

2. **Use `tomte tox` uniformly** with a note that on legacy repos still on bare tox, the user should strip the `tomte` prefix.

Option 1 is closer to the skill's "repo-agnostic, key off the existing layout" framing.

claude-skills/bump-versions/SKILL.md
DhairyaPatel7: Good catch — fixed in two passes. `edcd42af8` did an initial scrub (env-var-driven `LOCAL_OAEA_CHECKOUT` / `LOCAL_OA_CHECKOUT` defaults), and `2e8125229` then did the right thing: the Phase 5.1 drift-audit snippet now fetches each upstream's `packages.json` directly via `gh api` — same path Phase 1.2 and Phase 2.1 already use for `pyproject.toml` / `HISTORY.md` / `upgrading.md`. No local checkouts required, no env-var ceremony, no per-user filesystem assumption. The line-17 intro reference is also gone — now just "the legacy `oa-bump` skill" with no filesystem path.

claude-skills/bump-versions/SKILL.md
DhairyaPatel7: Agreed — flipped to `disable-model-invocation: true` in `edcd42af8`. The skill now only fires on explicit `/bump-versions`. Reasoning matches yours: Phase 3–5 mutate `pyproject.toml`, every workflow file, every dev YAML, run `autonomy packages lock`, and may clean `~/.aea/cache/packages`. Too much blast-radius to auto-trigger on a conversational hint like "check if this repo is on latest open-autonomy".

claude-skills/bump-versions/SKILL.md
DhairyaPatel7: Fixed in `edcd42af8`. Phase 6 now opens with a one-line probe that picks the right invocation form based on the repo's Makefile:

```bash
TOX=$(grep -q "tomte tox" Makefile 2>/dev/null && echo "tomte tox" || echo "tox")
```

…and every Phase 6 command switched from bare `tox -e <env>` to `$TOX -e <env>` (format, verify, full CI suite). That matches the "repo-agnostic, key off the existing layout" framing — went with your option 1 because it stays correct on legacy bare-tox repos without manual prefix stripping.

The remaining `tox -e ...` references in the doc (Phase 3.2 / 3.4 / 6.1) all reproduce CI error strings or name envs conceptually, not commands you run verbatim — left those alone.

claude-skills/bump-versions/SKILL.md
```diff
@@ -0,0 +1,592 @@
...
+disable-model-invocation: true
+---
+
+# Bump `open-autonomy` + `open-aea` (and upstream pins) in any downstream repo
+
+End-to-end procedure for bumping a Valory downstream repo (mech, mech-interact, mech-client, mech-server, mech-predict, mech-agents-fun, trader, optimus, market-creator, market-resolver, meme-ooorr, kv-store, funds-manager, genai, etc.) to the latest released `open-autonomy` and `open-aea`. Repo-agnostic — keys off `[project] dependencies` and `[tool.tomte] upstream_pins` in `pyproject.toml`, not a hardcoded repo list.
+
+The two frameworks that drive **code changes** are `open-autonomy` (OA) and `open-aea` (OAEA). Everything else listed in `[tool.tomte] upstream_pins` (e.g. `valory-xyz/mech-interact`, `valory-xyz/genai`, `valory-xyz/mech`) is bumped **tag + third-party-hash only** — never apply code rewrites for those.
+
+> Reference: a Poetry-era predecessor of this procedure exists in the legacy `oa-bump` skill. The pitfalls and tox/CI hardening notes there still apply; everything below assumes the modern **uv + tox + PEP 621 `[project]`** layout the fleet now uses.
```
OjusWiZard: **Stale cross-reference.** No `oa-bump` skill exists in this repo. `claude-skills/` contains only `audit-fsm`, `audit-resilience`, and `bump-versions`; `git log --all --full-history -- 'claude-skills/oa-bump*'` returns nothing. Operator following the pointer finds dead air.

Fix: delete the paragraph, or inline the specific pitfalls it claimed to carry (the plugin-coupling and `tomte tox` vs bare `tox` warnings are already covered inline).

OjusWiZard (commented):
Multi-perspective review (code-reviewer, comment-analyzer, silent-failure-hunter, code-simplifier). Seven inline findings below — all are operator-procedure correctness/silent-failure issues, no critical/security defects. Already-addressed items from prior rounds (personal paths, `disable-model-invocation`, `$TOX` probe replacing bare `tox`) are not re-flagged.

claude-skills/bump-versions/SKILL.md
```diff
@@ -0,0 +1,592 @@
...
+| `open-aea-cli-ipfs` | OAEA | `plugins/aea-cli-ipfs` |
+| `open-aea-cli-benchmark` | OAEA | `plugins/aea-cli-benchmark` |
+| `open-aea-ci-helpers` / `open-aea-dev-helpers` | OAEA | `plugins/aea-{ci,dev}-helpers` |
+
+**Verify the OA-side plugins are published for the target OA version** (PyPI sometimes lags by a few minutes after a release):
+
+```bash
+# Anchor the version on a non-digit boundary so 0.21.2 doesn't false-match inside 0.21.20.
+pip index versions open-aea-test-autonomy 2>/dev/null | grep -E "(^|[^0-9])$NEW_OA([^0-9]|$)"
+pip index versions open-aea-helpers       2>/dev/null | grep -E "(^|[^0-9])$NEW_OA([^0-9]|$)"
```
OjusWiZard: **Silent-failure: plugin-not-yet-published proceeds anyway.** `pip index versions … | grep …` exits 1 on no match. The prose below says "hold the bump and tell the user", but the command itself doesn't enforce that — easy to miss in a Claude-driven run.

Fix:

```bash
pip index versions open-aea-test-autonomy 2>/dev/null \
  | grep -E "(^|[^0-9])$NEW_OA([^0-9]|$)" \
  || { echo "ERROR: open-aea-test-autonomy==$NEW_OA not on PyPI — hold the bump"; exit 1; }
```

Same fix on the `open-aea-helpers` line.

claude-skills/bump-versions/SKILL.md
```diff
@@ -0,0 +1,592 @@
...
+
+If missing, hold the bump and tell the user — they should wait for the publish job to finish.
+
+### 1.2 Verify OA at `NEW_OA` pins OAEA at `NEW_OAEA`
+
+The bump only makes sense if the new OA itself pins the new OAEA. Pull OA's own `pyproject.toml` at `v$NEW_OA` and read the `open-aea` line(s) — OA still ships Poetry-style (`[tool.poetry.dependencies]`), so the pin shape is e.g. `open-aea = {version = "==2.2.3", extras = ["all"]}`:
+
+```bash
+gh api "repos/valory-xyz/open-autonomy/contents/pyproject.toml?ref=v$NEW_OA" \
+  --jq '.content' | base64 -d | grep -E '^[[:space:]]*open-aea[[:space:]]*='
```
OjusWiZard: **Silent-failure: `gh api` failure indistinguishable from "OA doesn't pin OAEA".** If `gh api` errors (bad ref, rate limit, auth), `base64 -d` gets empty input, `grep` exits 1, operator sees no output — but the **remediation differs**: a fetch failure is a network/setup issue, an empty grep result is a Phase 1.2 trust-OA's-pin branch.

Fix: materialize the fetch first.

```bash
gh api "repos/valory-xyz/open-autonomy/contents/pyproject.toml?ref=v$NEW_OA" \
  --jq '.content' > /tmp/oa.b64 || { echo "ERROR: gh api failed for OA pyproject.toml@v$NEW_OA"; exit 1; }
base64 -d /tmp/oa.b64 | grep -E '^[[:space:]]*open-aea[[:space:]]*='
```

claude-skills/bump-versions/SKILL.md
```diff
@@ -0,0 +1,592 @@
...
+gh api "repos/valory-xyz/open-autonomy/contents/HISTORY.md?ref=v$NEW_OA" \
+  --jq '.content' | base64 -d > "$TMP/oa_history.md"
+gh api "repos/valory-xyz/open-autonomy/contents/docs/upgrading.md?ref=v$NEW_OA" \
+  --jq '.content' | base64 -d > "$TMP/oa_upgrading.md"
+
+# Open-aea — HISTORY.md and docs/upgrading.md
+gh api "repos/valory-xyz/open-aea/contents/HISTORY.md?ref=v$NEW_OAEA" \
+  --jq '.content' | base64 -d > "$TMP/oaea_history.md"
+gh api "repos/valory-xyz/open-aea/contents/docs/upgrading.md?ref=v$NEW_OAEA" \
+  --jq '.content' | base64 -d > "$TMP/oaea_upgrading.md"
```
OjusWiZard: **Highest-impact silent failure in the doc.** Four `gh api … | base64 -d > "$TMP/…"` chains (lines 137–146) can each fail (bad tag, rate limit, auth) and silently leave an empty file — the redirect succeeds even if `gh api` errors. Phase 2.2 then reads empty files, finds zero breaking changes, and reports "no action required". Operator applies pin bumps without the code changes upgrading.md actually required → broken downstream after merge.

Fix: add `set -euo pipefail` at the top of the bash block, or guard each fetch:

```bash
for f in oa_history oa_upgrading oaea_history oaea_upgrading; do
  [[ -s "$TMP/${f}.md" ]] || { echo "ERROR: $TMP/${f}.md empty — fetch failed"; exit 1; }
done
```

claude-skills/bump-versions/SKILL.md
```diff
@@ -0,0 +1,592 @@
...
+
+Two known failure modes (from real bumps):
+
+- **`ValueError: not enough values to unpack (expected 1, got 0)` from `base.py:valid()`** — leftover empty cache dir from an interrupted fetch:
+  ```bash
+  find ~/.aea/cache/packages -mindepth 1 -maxdepth 1 -type d -empty -exec rmdir {} \;
+  ```
+- **`Destination path '...' already exists`** — a prior sync wrote a package to the wrong on-disk path:
+  ```bash
+  git clean -fdx packages/
```
OjusWiZard: **Destructive: silently nukes unstaged operator edits.** Phase 5.3 frames `git clean -fdx packages/` as routine sync recovery, but `-fdx` removes untracked files, untracked dirs, **and gitignored files**. If the operator has applied Phase 3.4 YAML edits or Phase 4 `.py` rewrites that aren't staged yet (and the doc never tells them to stage between phases), this command erases them — the next sync "works" because the unfinished edits are gone.

Fix: use `git stash -u -- packages/` for the retry, or add an explicit precondition: *"Only safe if all Phase 3–4 edits under `packages/` are staged or committed. If unsure, `git stash -u -- packages/` first."*

claude-skills/bump-versions/SKILL.md
```diff
@@ -0,0 +1,592 @@
...
+
+## Phase 6 — Lock + format + verify
+
+Run in this exact order. Each step gates the next.
+
+**Pick the right `tox` invocation form first.** Most of the modern fleet (market-creator, market-resolver, kv-store, funds-manager, genai, mech, mech-interact, …) runs envs through `tomte tox -e <env>` so the testenvs inherit dep blocks from tomte's canonical config. Bare `tox -e <env>` on those repos fails with `ERROR: env '<name>' not defined in tox config`. Detect once and use `$TOX` throughout:
+
+```bash
+# Use `tomte tox` if the Makefile references it; otherwise fall back to bare `tox`.
+TOX=$(grep -q "tomte tox" Makefile 2>/dev/null && echo "tomte tox" || echo "tox")
```
OjusWiZard: **Silent fallback breaks the "repo-agnostic" claim for no-Makefile repos.** Some uv-only fleets don't ship a Makefile. `grep -q … Makefile 2>/dev/null` swallows the file-not-found error and `$TOX` silently defaults to bare `tox`. Every subsequent `$TOX -e <env>` then fails confusingly with `env not defined` for repos that actually need `tomte tox`.

Fix:

```bash
if [[ ! -f Makefile ]]; then
  echo "WARNING: no Makefile; defaulting TOX=tox — verify this is correct for this repo"
  TOX=tox
elif grep -q "tomte tox" Makefile; then
  TOX="tomte tox"
else
  TOX=tox
fi
```

claude-skills/bump-versions/SKILL.md
```diff
@@ -0,0 +1,592 @@
...
+---
+
+## Appendix: quick-reference grep patterns
+
+```bash
+# Every site that still references the old OA version
+grep -rln "$CUR_OA" --exclude-dir=.git --exclude-dir=.venv --exclude=uv.lock --exclude=poetry.lock .
+
+# Every site that still references the old OAEA version
+grep -rln "$CUR_OAEA" --exclude-dir=.git --exclude-dir=.venv --exclude=uv.lock --exclude=poetry.lock .
```
OjusWiZard: **Inconsistent with Phase 3's own correctness rule.** Phase 3 (line 195) explicitly requires `-E "==${CUR_OA}([^0-9]|$)"` to prevent `0.21.2` matching `0.21.20`. The Appendix quick-reference patterns on lines 582 and 585 drop the boundary and the `==` prefix — with `CUR_OA=0.21.2`, they false-match `0.21.20` everywhere, exactly the failure mode Phase 3 calls out.

Fix: either apply Phase 3's boundary pattern here too, or relabel the section as *"loose triage greps (false positives expected — use Phase 3 patterns for the canonical sweep)"*.

claude-skills/bump-versions/SKILL.md
DhairyaPatel7: Fixed in `0a3d2bba5`. Deleted the paragraph — the plugin-coupling and `tomte tox` vs bare `tox` warnings are already inlined in Phase 1.1 and Phase 6 respectively, so there's nothing being kept out-of-tree. Replaced with a one-liner noting the uv + PEP 621 assumption and that Poetry repos need the equivalent edits applied to `[tool.poetry.dependencies]`.

claude-skills/bump-versions/SKILL.md
DhairyaPatel7: Fixed in `0a3d2bba5`. Both plugin checks now run inside a `for pkg in ...; do ... done` loop with `|| { echo ERROR; exit 1; }` so a no-match exits the skill with a clear `ERROR: $pkg==$NEW_OA not yet on PyPI — hold the bump until the publish job completes` rather than silently letting the bump proceed.

claude-skills/bump-versions/SKILL.md
DhairyaPatel7: Fixed in `0a3d2bba5`. The `gh api` fetch now writes to `$TMP/oa_pyproject.b64` first with `|| { echo ERROR: gh api failed for OA pyproject.toml@v$NEW_OA — check tag exists, gh auth, rate limit; exit 1; }`. The subsequent `base64 -d | grep` then either prints the OAEA pin line or exits cleanly with no output (the trust-OA's-pin branch). Two failure modes, two distinguishable signals.

claude-skills/bump-versions/SKILL.md
DhairyaPatel7: Fixed in `0a3d2bba5`. The four fetches now run via a `_fetch()` shell helper that pipes `gh api … | base64 -d` to the output file and then asserts `[[ -s $out ]]` — a zero-byte file kills the skill with `ERROR: empty fetch — $repo:$path@$ref`. Added `set -euo pipefail` at the top of the block too, so any failure in the `gh api | base64 -d` pipe propagates instead of being masked by the redirect's success. This was the highest-impact silent failure in the doc — appreciate the catch.

claude-skills/bump-versions/SKILL.md
DhairyaPatel7: Fixed in `0a3d2bba5`. Replaced `git clean -fdx packages/` with `git stash push -u -- packages/` → retry sync → `git stash pop`, so any in-progress Phase 3.4 YAML / Phase 4 .py edits round-trip safely. Also added an explicit precondition: `git clean -fdx packages/` is only safe after verifying `git status -- packages/` is clean AND `git ls-files --others --exclude-standard -- packages/` is empty. Operator can still reach for the clean form when they've confirmed nothing's pending, but the default path no longer eats their work.

claude-skills/bump-versions/SKILL.md
DhairyaPatel7: Fixed in `0a3d2bba5`. Replaced the one-liner with an explicit three-branch conditional:

```bash
if [[ ! -f Makefile ]]; then
  echo "WARNING: no Makefile — defaulting TOX=tox. Verify this repo doesn't require 'tomte tox'."
  TOX="tox"
elif grep -q "tomte tox" Makefile; then
  TOX="tomte tox"
else
  TOX="tox"
fi
```

No-Makefile repos now get a loud warning instead of a silent fallback, so the operator sees the assumption being made before the first `$TOX -e <env>` fails confusingly downstream.

claude-skills/bump-versions/SKILL.md
DhairyaPatel7: Fixed in `0a3d2bba5`. The two version-targeting Appendix greps now use the same `==${VER}([^0-9]|$)` boundary pattern Phase 3 enforces — no more false-matching `==0.21.2` inside `==0.21.20`. Added a one-line preamble noting the discipline so the choice is visible: anyone wanting the looser triage view can drop the boundary deliberately. The YAML package-name grep keeps the loose form because it's matching package names, not version literals.

claude-skills/bump-versions/SKILL.md
```diff
@@ -0,0 +1,620 @@
...
+
+- `deployments/Dockerfiles/**/requirements.txt` and `deployments/Dockerfiles/**/Dockerfile` — `open-autonomy[all]==X` baked into the Docker user image
+- `docs/**/*.md` — tutorial snippets with `pip install open-autonomy==<old>`
+- `Pipfile` / `Pipfile.lock` — some repos still ship pipenv alongside uv
+- `scripts/*.sh` and repo-root bootstrap shells with `pip install ... ==<old>`
+- `Makefile` targets that hardcode the version (rare but real)
+- `mkdocs.yml` env_vars and template strings
+- `setup.py` if the repo still ships one alongside `pyproject.toml`
+
+If `grep -rn "$CUR_OA\b"` (no `==` prefix) returns hits in non-code paths (READMEs, badges, CHANGELOGs), update those too **when** they're describing the supported version. Don't touch historical changelog entries.
```
LOCKhart07: **Uses `\b` — the doc's own rule (lines 207 and 299) explicitly bans it.**

Line 207: *"`\b` is GNU-grep-only and trips on BSD grep / macOS."* Line 299: *"`\b/\|` are GNU-only."* This line then writes:

```bash
grep -rn "$CUR_OA\b"
```

On macOS BSD grep this either silently no-matches or false-matches `0.21.20` inside `==0.21.2` — exactly the failure mode Phase 3 calls out at length.

Fix: use the same `([^0-9]|$)` boundary the rest of the doc enforces:

```bash
grep -rnE "${CUR_OA}([^0-9]|$)"
```

claude-skills/bump-versions/SKILL.md
```diff
@@ -0,0 +1,620 @@
...
+echo "using TOX=$TOX"
+```
+
+```bash
+# 1. Regenerate the lockfile and install
+uv lock
+uv sync --all-groups
+
+# 2. Re-lock packages (third-party hashes cascade into dev fingerprints)
+autonomy packages lock
```
LOCKhart07: **`autonomy packages lock` runs BEFORE the formatters here, contradicting this repo's own published rule.**

The order in steps 2 → 3 is `autonomy packages lock` → `$TOX -e black` / `isort`. But the open-autonomy `CLAUDE.md` states:

> **Hashing rule:** always run `autonomy packages lock` … at the end of a work cycle, after every other edit/lint/format step has settled. **Locking earlier is wasted work**: any subsequent edit (including linter-triggered ones) will dirty the files again and require a re-lock.

The conditional re-lock note on line 517 ("Re-run `autonomy packages lock` if anything under packages/ was reformatted") is a band-aid — if the operator skims past it, `autonomy packages lock --check` at line 522 fails and the failure looks like a sync regression rather than missed re-lock ordering.

Fix: format first, then lock once at the end:

```bash
# 1. Regenerate lockfile and install
uv lock && uv sync --all-groups

# 2. Auto-format first (some pin changes trip isort/black)
$TOX -e black
$TOX -e isort

# 3. Lock once, after all edits have settled
autonomy packages lock

# 4. Verify
$TOX -e black-check
$TOX -e isort-check
autonomy packages lock --check
```

claude-skills/bump-versions/SKILL.md
```diff
@@ -0,0 +1,620 @@
...
+## Phase 0 — Establish ground truth
+
+Run these in parallel before doing anything else:
+
+```bash
+# 1. Confirm the current cwd is a Valory downstream repo (must have pyproject.toml + packages/packages.json)
+test -f pyproject.toml && test -f packages/packages.json && echo "OK: looks like a downstream repo"
+
+# 2. Current pins — read pyproject.toml's [project] dependencies block
+grep -E "^\s*\"open-(autonomy|aea)" pyproject.toml
```
LOCKhart07: **Phase 0 grep only matches PEP 621 shape — silently no-ops on Poetry downstreams.**

The regex `^\s*"open-(autonomy|aea)` requires the leading quote of a PEP 621 `dependencies = ["open-…==X"]` entry. On a Poetry-style downstream (the form is `open-aea = {version = "==X", extras = ["all"]}`, no leading quote), this returns zero hits — the operator gets no `CUR_OA` / `CUR_OAEA` baseline at all.

Line 14 explicitly claims the skill supports Poetry repos:

> Repos still on Poetry need the equivalent edits applied to `[tool.poetry.dependencies]` / `poetry.lock` — the phase structure is unchanged.

That claim doesn't hold if Phase 0 silently fails to find current pins. Fix options:

1. Add a parallel Poetry-shape grep:
   ```bash
   grep -E "^\s*open-(autonomy|aea)[^=]*=" pyproject.toml
   ```
   …and combine, or run both.
2. Or scope the skill to uv-only and drop the Poetry sentence from line 14. Either is fine — the current state (claims support, doesn't actually probe) is the silent-failure pattern reviewers caught elsewhere.

claude-skills/bump-versions/SKILL.md
```diff
@@ -0,0 +1,620 @@
...
+
+### 4.2 Apply changes the same way the upgrading guide describes
+
+For each change, write the smallest diff that satisfies the guide. Examples of common shapes:
+
+```python
+# OAEA 2.x: BaseSkillTestCase.setup() → setup_method()
+class TestMyBehaviour(BaseSkillTestCase):
+-    def setup(self) -> None:
++    def setup_method(self) -> None:
```
LOCKhart07, line 397: **Sample rename diff is internally inconsistent — won't match what an operator finds.**

The OLD definition `def setup(self) -> None:` is shown with a body that already calls `super().setup_method()`. In a real OAEA 2.x rename, both the def AND the call site change — the OLD body would have called `super().setup()`.

As written the diff suggests *"rename only the def, leave the body untouched"*, which means an operator who greps for `super().setup()` (the realistic before-state) won't find anything to update. Either fix the example so the body migrates too:

```diff
 class TestMyBehaviour(BaseSkillTestCase):
-    def setup(self) -> None:
-        super().setup()
+    def setup_method(self) -> None:
+        super().setup_method()
         ...
```

…or add a note explaining why the example shows the body already using `setup_method` (e.g. "some legacy code was inconsistent — only the def needed renaming").

claude-skills/bump-versions/SKILL.md
```diff
@@ -0,0 +1,620 @@
...
+
+`autonomy packages sync --source` only refreshes packages it already knows about — it trusts whatever CID is currently in `packages.json`. Drift from a prior bump (a CID miscopy, a never-synced upstream) is invisible to the regular sync path. Audit first:
+
+Fetch each upstream's `packages.json` straight from GitHub via `gh api` (same path the rest of the skill uses for HISTORY/upgrading) — no local checkouts required:
+
+```bash
+python3 <<'PY'
+import json, subprocess
+
+local = json.load(open('packages/packages.json'))['third_party']
```
LOCKhart07: **`json.load(open(...))['third_party']` will KeyError on a downstream with no third-party packages.**

A greenfield downstream (or one that's only just been bootstrapped) may have `packages.json` with only `dev` populated and no `third_party` key — the audit script crashes before reporting anything useful.

Fix: defensive `.get('third_party', {})` so the audit reports `No third-party CID drift.` cleanly:

```python
local = json.load(open('packages/packages.json')).get('third_party', {})
```

Same consideration on line 426 (`data.get("dev", {})` is already correct — mirror that style).

OjusWiZard (commented):
Two more silent-failure / prose-only findings beyond LOCKhart07's batch — same shape as the silent-fetch issues addressed in `0a3d2bba5`.

claude-skills/bump-versions/SKILL.md
```diff
@@ -0,0 +1,620 @@
...
+The bump only makes sense if the new OA itself pins the new OAEA. Pull OA's own `pyproject.toml` at `v$NEW_OA` and read the `open-aea` line(s) — OA still ships Poetry-style (`[tool.poetry.dependencies]`), so the pin shape is e.g. `open-aea = {version = "==2.2.3", extras = ["all"]}`. Materialize the fetch first so a network/auth failure is distinguishable from an empty grep result (the latter would be the trust-OA's-pin branch):
+
+```bash
+TMP=${TMP:-$(mktemp -d)}
+gh api "repos/valory-xyz/open-autonomy/contents/pyproject.toml?ref=v$NEW_OA" \
+  --jq '.content' > "$TMP/oa_pyproject.b64" \
+  || { echo "ERROR: gh api failed for OA pyproject.toml@v$NEW_OA — check tag exists, gh auth, rate limit"; exit 1; }
+base64 -d "$TMP/oa_pyproject.b64" | grep -E '^[[:space:]]*open-aea[[:space:]]*='
+```
+
```
OjusWiZard, line 112: **Silent-failure: "trust OA's pin and override `NEW_OAEA`" is prose-only.**

Line 100 just prints OA's `open-aea = {...}` line to stdout. There's no programmatic comparison against `$NEW_OAEA`, no auto-override, no exit. In a Claude-driven run the line will scroll past, `NEW_OAEA` stays at whatever the operator asked for, and the mismatch only surfaces hundreds of lines later at `tox -e check-dependencies` or `autonomy packages sync` — exactly the failure mode the prose says we're guarding against.

Same shape as the silent-fetch issues fixed in `0a3d2bba5`. Fix: extract the pin, compare, override + warn:

```bash
OA_PINNED_OAEA=$(
  base64 -d "$TMP/oa_pyproject.b64" \
  | grep -E '^[[:space:]]*open-aea[[:space:]]*=' \
  | grep -oE '==[0-9]+\.[0-9]+\.[0-9]+' | head -1 | sed 's/^==//'
)
if [[ -z "$OA_PINNED_OAEA" ]]; then
  echo "ERROR: could not extract open-aea pin from OA pyproject.toml@v$NEW_OA"; exit 1
fi
if [[ "$OA_PINNED_OAEA" != "$NEW_OAEA" ]]; then
  echo "WARNING: OA v$NEW_OA pins open-aea==$OA_PINNED_OAEA; you asked for $NEW_OAEA — overriding to OA's pin."
  NEW_OAEA="$OA_PINNED_OAEA"
fi
```

claude-skills/bump-versions/SKILL.md
```diff
@@ -0,0 +1,620 @@
...
+```bash
+# Latest OA release (tag name includes the leading 'v'; strip if you need the bare version)
+gh api repos/valory-xyz/open-autonomy/releases/latest --jq '.tag_name'
+
+# Latest OAEA release
+gh api repos/valory-xyz/open-aea/releases/latest --jq '.tag_name'
+```
+
+Set `NEW_OA` (e.g. `0.21.20`) and `NEW_OAEA` (e.g. `2.2.3`). If they equal the current pins, **stop and tell the user** — there's nothing to bump.
```
OjusWiZard: **Silent-failure: "stop and tell the user" if `NEW_*` equal `CUR_*` is prose-only.**

No shell guard enforces the stop. A Claude-driven run with no version args, against a repo that's already at latest, will sail straight into Phase 2+ and churn out a no-op-then-relock PR (touched timestamps, regenerated lockfile, re-formatted `packages.json`) for zero version change. The operator then has to figure out why a "nothing to do" run produced a diff.

Fix: gate the rest of the skill on a real diff:

```bash
if [[ "$NEW_OA" == "$CUR_OA" && "$NEW_OAEA" == "$CUR_OAEA" ]]; then
  echo "Already at latest (open-autonomy==$CUR_OA, open-aea==$CUR_OAEA). Nothing to bump."
  exit 0
fi
```

DhairyaPatel7: Fixed in `7b905e20b`. The prose-only stop is now a hard shell guard:

```bash
if [[ "$NEW_OA" == "$CUR_OA" && "$NEW_OAEA" == "$CUR_OAEA" ]]; then
  echo "Already at latest (open-autonomy==$CUR_OA, open-aea==$CUR_OAEA). Nothing to bump."
  exit 0
fi
```

A Claude-driven run that resolves `NEW_*` equal to `CUR_*` now exits cleanly with a clear "already at latest" message before any lockfile/sync/format mutation touches the tree — no more no-op-then-relock PRs to figure out after the fact.

claude-skills/bump-versions/SKILL.md
DhairyaPatel7: Fixed in `7b905e20b`. Phase 1.2 now extracts OA's `open-aea` pin programmatically, compares against `$NEW_OAEA`, and auto-overrides + WARNs on mismatch — exact shape you suggested:

```bash
OA_PINNED_OAEA=$(
  base64 -d "$TMP/oa_pyproject.b64" \
    | grep -E '^[[:space:]]*open-aea[[:space:]]*=' \
    | grep -oE '==[0-9]+\.[0-9]+\.[0-9]+' \
    | head -1 \
    | sed 's/^==//'
)
if [[ -z "$OA_PINNED_OAEA" ]]; then
  echo "ERROR: could not extract open-aea pin from OA pyproject.toml@v$NEW_OA"; exit 1
fi
if [[ "$OA_PINNED_OAEA" != "$NEW_OAEA" ]]; then
  echo "WARNING: OA v$NEW_OA pins open-aea==$OA_PINNED_OAEA; you asked for $NEW_OAEA — overriding to OA's pin."
  NEW_OAEA="$OA_PINNED_OAEA"
fi
```

Same shape as the silent-fetch hardening from `0a3d2bba5` — the prose-only "trust OA's pin" branch is now enforced before any Phase 3+ mutation. Mismatch surfaces here with the root cause named, not hundreds of lines later in a `check-dependencies` / `packages sync` stack trace.

claude-skills/bump-versions/SKILL.md
DhairyaPatel7: Fixed in `7b905e20b`. Phase 0's pin grep now matches both layouts, so a Poetry downstream actually produces a `CUR_*` baseline instead of silently no-opping:

```bash
# 2. Current pins — match both layouts:
#    - PEP 621 / uv:  dependencies = ["open-autonomy[cli]==X", ...]   (leading quote)
#    - Poetry:        open-aea = {version = "==X", extras = [...]}     (bare key)
#    A PEP-621-only pattern silently no-ops on Poetry downstreams.
grep -E "^\s*(\"open-(autonomy|aea)|open-(autonomy|aea)[A-Za-z0-9_-]*\s*=)" pyproject.toml
```

First alternation catches the PEP 621 quoted-list shape (`"open-autonomy[cli]==X"`); second catches the Poetry bare-key shape (`open-aea = {version = "==X", ...}`) — including the `open-aea-ledger-ethereum` / `-cli-ipfs` / etc. plugin keys via the `[A-Za-z0-9_-]*` suffix. Line 14's Poetry-support claim now actually holds at Phase 0.

claude-skills/bump-versions/SKILL.md
DhairyaPatel7: Fixed in `7b905e20b`. Replaced the `\b` form with the same `([^0-9]|$)` boundary Phase 3 enforces:

```bash
grep -rnE "${CUR_OA}([^0-9]|$)"
```

Now BSD-grep-safe and consistent with the doc's own line 207 / 299 rule. Good catch — the discipline was written down twice in the same doc but the example line still violated it.

claude-skills/bump-versions/SKILL.md
DhairyaPatel7: Fixed in `7b905e20b`. The sample diff now migrates both the def AND the `super()` call site, so an operator grepping for the realistic `super().setup()` before-state actually finds something:

```diff
 class TestMyBehaviour(BaseSkillTestCase):
-    def setup(self) -> None:
-        super().setup()
+    def setup_method(self) -> None:
+        super().setup_method()
         ...
```

Also adjusted the heading line to call out "rename def AND the super() call site" so the example's intent is explicit, not inferred.

claude-skills/bump-versions/SKILL.md
DhairyaPatel7: Fixed in `7b905e20b`:

```python
local = json.load(open('packages/packages.json')).get('third_party', {})
```

Mirrors the `data.get("dev", {})` style on the upstream-side line. Greenfield downstreams with no `third_party` key now report `No third-party CID drift.` cleanly instead of KeyError-ing out of the audit before it can report anything.

claude-skills/bump-versions/SKILL.md
DhairyaPatel7: Fixed in `7b905e20b`. Phase 6 reorders to format-first → lock-once per `CLAUDE.md`'s "lock at end" rule, with an explicit ordering-rule callout above the block so the reasoning is visible:

> **Ordering rule** (per open-autonomy `CLAUDE.md`): `autonomy packages lock` runs **once, at the end**, after every other edit/lint/format step has settled. Locking earlier is wasted work — any subsequent formatter-triggered rewrite under `packages/` re-dirties the fingerprints and forces a re-lock anyway.

```bash
# 1. Regenerate the lockfile and install
uv lock
uv sync --all-groups

# 2. Auto-format first (some pin changes trip isort/black, which would otherwise dirty
#    packages/ after a too-early `autonomy packages lock`)
$TOX -e black
$TOX -e isort

# 3. Lock once, after all edits have settled (third-party hashes cascade into dev fingerprints)
autonomy packages lock

# 4. Verify lock + format
$TOX -e black-check
$TOX -e isort-check
autonomy packages lock --check
```

Drops the conditional re-lock band-aid that was the failure mode you flagged: an operator skimming past line 517 would have hit `autonomy packages lock --check` failing later and chased it as a sync regression rather than a missed re-lock. Now there's nothing to re-lock — the order makes it unnecessary.

LOCKhart07 (approved):
All prior-round issues (mine + OjusWiZard's two follow-ups) are correctly addressed in `7b905e20`. Approving.

Two minor non-blocking notes on the new Phase 1 shell guard if you want to tighten further:

1. **`SKILL.md:66` — guard reads `$CUR_OA` / `$CUR_OAEA` but Phase 0 never assigns them as shell variables.** The Phase 0 grep at line 30 only prints matches to stdout; the table at lines 43-44 is descriptive ("Record into working memory") rather than executable. If an operator (or a Claude-driven run) reaches the guard without first capturing the vars, `"$NEW_OA" == ""` evaluates false and the guard silently no-ops — script proceeds normally, which is the safer failure mode (false negative, not misfire), but the guard is decorative until Phase 0 produces an executable capture. Worth adding a `CUR_OA=$(grep ... | sed ...)` snippet so the guard has real inputs to compare against.

2. **`SKILL.md:66` — guard only checks OA + OAEA equality, ignoring `upstream_pins`.** The skill's stated scope (Phase 1.3 / Phase 5.2) includes refreshing upstream-pin tags. If OA and OAEA happen to already be at latest but a `mech-interact` / `genai` / etc. release has dropped since the last bump, the guard exits with "Nothing to bump" and silently skips the upstream_pins refresh the operator may have invoked the skill for. Either narrow the message to "no framework bump — continuing for upstream_pins" and proceed, or extend the equality check across `upstream_pins`.

Both are polish items, not blockers.

claude-skills/bump-versions/SKILL.md
```diff
@@ -0,0 +1,652 @@
...
+gh api repos/valory-xyz/open-autonomy/releases/latest --jq '.tag_name'
+
+# Latest OAEA release
+gh api repos/valory-xyz/open-aea/releases/latest --jq '.tag_name'
+```
+
+Set `NEW_OA` (e.g. `0.21.20`) and `NEW_OAEA` (e.g. `2.2.3`). If they equal the current pins, **stop**: prose-only guidance silently sails past a no-op run and produces a "touched lockfile / re-formatted `packages.json`" PR for zero version change. Gate the rest of the skill on a real diff:
+
+```bash
+if [[ "$NEW_OA" == "$CUR_OA" && "$NEW_OAEA" == "$CUR_OAEA" ]]; then
```
OjusWiZard, line 66: **[Non-blocking] Guard depends on shell vars the doc never captures.**

`$NEW_OA`, `$CUR_OA`, `$NEW_OAEA`, `$CUR_OAEA` are referenced here, in Phase 3's greps (lines 238, 244, 357), and in Phase 5.2's sync URLs (lines 481, 484) — but the only `*_OA*=` shell assignment in the entire file is the conditional override at line 128. Phase 0's `grep` at line 30 and Phase 1's `gh api … --jq '.tag_name'` at lines 57-60 only **print**; the prose "Set `NEW_OA`" at line 63 is operator-instruction, not a shell command.

Without `set -u` (this Phase doesn't declare it), if the operator/Claude doesn't manually `export` the four vars first, this guard becomes `[[ "" == "" && "" == "" ]]` → true → `exit 0` with "Already at latest, nothing to bump." The bump aborts silently and reports "no action needed" when the truth is "I never resolved any version." Same root cause then poisons Phase 3's greps (collapse to `grep -rlE "==([^0-9]|$)"`, matching every `==`-prefixed token in the tree) and Phase 5.2's sync URLs (become `valory-xyz/open-aea:v` with a confusing ref error).

Fix — either capture explicitly:

```bash
# Phase 0
CUR_OA=$(grep -E '...' pyproject.toml | grep -oE '==[0-9]+\.[0-9]+\.[0-9]+' | head -1 | sed 's/^==//')
CUR_OAEA=$(...)

# Phase 1
NEW_OA=$(gh api repos/valory-xyz/open-autonomy/releases/latest --jq '.tag_name' | sed 's/^v//')
NEW_OAEA=$(gh api repos/valory-xyz/open-aea/releases/latest --jq '.tag_name' | sed 's/^v//')
```

…or fail loud before the guard:

```bash
: "${CUR_OA:?CUR_OA unset — capture from Phase 0 first}"
: "${CUR_OAEA:?}"; : "${NEW_OA:?}"; : "${NEW_OAEA:?}"
if [[ "$NEW_OA" == "$CUR_OA" && "$NEW_OAEA" == "$CUR_OAEA" ]]; then ...
```

Not new to this commit — Phase 3's greps have had the dependency all along — but this guard makes the failure sharper (premature `exit 0` vs. noisy grep output). Approving as-is; flagging for follow-up.

OjusWiZard (approved):
LGTM — all five LOCKhart07 findings and both of my prior pass-2 findings (line 60 no-bump guard, line 102 OA-pin override) are addressed cleanly in `7b905e20b`. CI is green across every matrix cell.

One non-blocking observation left inline at the new Phase 1 guard: `$NEW_OA` / `$CUR_OA` / `$NEW_OAEA` / `$CUR_OAEA` are referenced as shell variables but the doc never shows the assignment — without `set -u`, the guard collapses to `[[ "" == "" && "" == "" ]] && exit 0` and silently aborts the bump. Worth tightening as a follow-up, not a merge blocker.

PR #2504, opened 2026-05-11
feat(analyse): composition-aware Mermaid view for fsm-specs; deprecate --json

jmoreira-valory:
## Proposed changes

Three improvements to `autonomy analyse fsm-specs`, all scoped to the `--mermaid` renderer + CLI surface:

**1. Composition-aware Mermaid output for composed AbciApps.**
When the analysed app is a composition (rounds span multiple sub-app skills), the diagram now emits a two-tier view instead of a flat ~200-line transition graph:

- **Dev sub-apps** (listed under \`dev\` in the repo's \`packages.json\`) are rendered as UML composite states (\`state X { ... }\`) with their internal rounds and transitions visible inside a labeled outer box. Styled with a green \`devGroup\` classDef.
- **Third-party sub-apps** (listed under \`third_party\` in \`packages.json\`, or anything outside \`dev\`) are collapsed to a single composite state containing a \`"..."\` placeholder — the canonical UML notation for a sub-FSM with hidden internals. Styled with a blue \`macro\` classDef + thicker border.
- **Cross-app transitions** wire the composites together at the top level, with their event labels aggregated.

For per-skill analysis (a single sub-app, no composition), behaviour is unchanged: flat per-round diagram.

The dev/third-party split is auto-discovered by walking parents of \`--package\` until \`packages.json\` is found. If no \`packages.json\` is reachable, the renderer falls back to the flat per-round diagram (zero behavioural change vs. today for users outside a packages tree).

**2. Markdown code-fence wrapping.**
The Mermaid output is now wrapped in a \`\`\`mermaid … \`\`\` block so the resulting \`fsm_specification.md\` renders inline as a diagram on GitHub, in VS Code's markdown preview, and in any markdown viewer that supports Mermaid (mkdocs, docusaurus, mermaid.live, etc.). No extra step needed to copy/paste into a renderer.

**3. Deprecate \`--json\` output.**
JSON output is unused across the fleet (zero \`fsm_specification.json\` files in any of the major valory-xyz service repos; only an internal test fixture references it). Soft-deprecation cycle:

- \`--help\` text for \`--json\` now reads \`Json file. (deprecated; use --yaml or --mermaid)\`.
- \`FSMSpecificationLoader.dump_json()\` raises a \`DeprecationWarning\` on use.
- Functionality and on-disk output unchanged in this PR (deferred removal to a future release).

## Fixes

PREDICT-879 — extract usable functionality from the deprecated \`open-autonomy-compose\` repo. The HTML inspector concept (compose's \`inspect\` web UI) is replaced by the composition-aware Mermaid output here; markdown-wrapping makes the spec file directly viewable everywhere Mermaid is supported.

## Types of changes

The composition view + markdown wrapping are additive: existing \`--yaml\` / \`--mermaid\` per-skill workflows render the same (flat) diagram. The \`--json\` deprecation is soft-only (warning, no removal). No public API or schema changes.

## Checklist

- [x] I am making a pull request against the \`main\` branch (left side). Also you should start your branch off our \`main\`.
- [ ] I have added tests that prove my fix is effective or that my feature works — _N/A: pure presentation-layer change; visual output validated against \`trader\` (composed) and \`market_manager_abci\` (per-skill) repos in the local fleet._
- [x] Any backwards-incompatible/breaking change has been clearly documented in the upgrading document. — _none in this PR; \`--json\` deprecation is soft, not breaking._

### What the new output looks like

Before (\`trader_abci\` composed, ~200 transitions across 11 sub-apps):

> Flat \`stateDiagram-v2\` with every atomic round and every transition, including framework sub-apps' internal transitions. Unreadable in any auto-layout viewer.

After:

> Seven labeled \`devGroup\` composite boxes — one per trader-owned sub-app (\`market_manager_abci\`, \`decision_maker_abci\`, \`tx_settlement_multiplexer_abci\`, \`staking_abci\`, \`agent_performance_summary_abci\`, \`check_stop_trading_abci\`, \`chatui_abci\`) with each sub-app's internal rounds and transitions drawn inside. Four \`macro\` composite boxes for the framework-provided sub-apps (\`mech_interact_abci\`, \`registration_abci\`, \`reset_pause_abci\`, \`transaction_settlement_abci\`) collapsed to \`...\` placeholders. ~160 lines total, fits on one screen.

### Test commands

\`\`\`bash
# Composed (composition-aware) view
autonomy analyse fsm-specs --package packages/valory/skills/trader_abci --app-class TraderAbciApp --mermaid --update
# -> opens in any Mermaid renderer (VS Code Markdown preview, mermaid.live, GitHub web).

# Per-skill (unchanged flat view)
autonomy analyse fsm-specs --package packages/valory/skills/market_manager_abci --app-class MarketManagerAbciApp --mermaid --update

# --json deprecation
autonomy analyse fsm-specs --help | grep json
# -> "  --json            Json file. (deprecated; use --yaml or --mermaid)"

PYTHONWARNINGS="default::DeprecationWarning" \
  autonomy analyse fsm-specs --package packages/valory/skills/market_manager_abci --app-class MarketManagerAbciApp --json --update
# -> DeprecationWarning: fsm-specs JSON output is deprecated...
\`\`\`

### Design notes

- The composition-vs-flat decision is automatic: if all rounds belong to a single sub-app (per-skill analysis), \`flatten = True\` and the renderer emits the legacy flat diagram. If \`packages.json\` can't be located (analysed package outside any packages tree), \`flatten = True\`. This keeps existing tooling green and only opts into the new view when there's a clear composition.
- Dev/third-party is determined exclusively from \`packages.json\`; the renderer reads only the \`dev\` section's skill keys. Repos that follow the standard packages.json layout get the new view for free.
- Empty composite-state blocks aren't valid Mermaid in stateDiagram-v2, hence the explicit \`"..."\` inner state in collapsed boxes — that placeholder is also semantically right (UML "collapsed sub-machine" notation).
- The \`--json\` deprecation cycle is soft on purpose. Zero \`fsm_specification.json\` files exist in the major service repos surveyed (trader, market-creator, market-resolver, mech-predict, olas-operate-middleware). A hard removal would be safe but I chose to give one release of grace in case anything external relies on the format.


autonomy/analyse/abci/app_spec.py
```diff
@@ -123,36 +140,229 @@ def dump_yaml(dfa: "DFA", file: Path) -> None:
...
 
     @staticmethod
-    def dump_mermaid(dfa: "DFA", file: Path) -> None:
-        """Dumps this DFA spec. to a file in Mermaid format."""
+    def dump_mermaid(
+        dfa: "DFA",
+        file: Path,
+        abci_app_cls: Optional[Any] = None,
+        dev_skills: Optional[Set[str]] = None,
+    ) -> None:
```
OjusWiZard, line 156: **[CRITICAL] No unit tests for the new composition-aware path.**

The new `dump_mermaid` body has 6+ behavioural branches (flat fall-back × 4 conditions, dev composite blocks, third-party collapsed blocks, the post-collapse self-loop guard, the styling blocks, the fence wrapping). None are exercised by the existing tests — `test_generate_mermaid` in `tests/test_autonomy/test_cli/test_analyse/test_specs.py` is a tautological round-trip (it calls `dump` to build the expected output, then calls the CLI which calls `dump` again, and asserts equality), so it will pass regardless of what this function emits.

The PR description's "validated visually against `trader`" claim covers exactly one branch under one set of inputs at one point in time. Repo policy in `CLAUDE.md` asks for ≥80% coverage; new functions like `_build_round_to_subapp`, `_extract_skill_name`, and `_load_dev_skill_names` have zero tests.

Suggested minimum: a test for each fall-back trigger, a test asserting the ```` ```mermaid ```` fence is present, a test asserting third-party rounds get collapsed to `state X { state "..." as X_collapsed }`, and a test asserting `dump_json` emits a `DeprecationWarning`.

autonomy/analyse/abci/app_spec.py
```diff
@@ -123,36 +140,229 @@ def dump_yaml(dfa: "DFA", file: Path) -> None:
...
+            round_classes.add(init_cls)
+        for final_cls in getattr(abci_app_cls, "final_states", set()) or set():
+            round_classes.add(final_cls)
+
+        result: Dict[str, str] = {}
+        for cls in round_classes:
+            module = getattr(cls, "__module__", "")
+            sub_app = FSMSpecificationLoader._extract_skill_name(module)
+            if sub_app:
+                result[cls.__name__] = sub_app
```
OjusWiZard: **[HIGH] Round-class name collision silently mis-classifies sub-app.**

`result` is keyed on `cls.__name__` (a string). When a composed AbciApp pulls in two sub-apps that each define a round class with the same short name, the second one iterated overwrites the first — last-writer-wins, no warning.

This is not hypothetical: `FinishedTransactionSubmissionRound` exists in both `packages/valory/skills/transaction_settlement_abci/rounds.py` and `packages/valory/skills/squads_transaction_settlement_abci/rounds.py`. Any composition pulling in both would silently assign the round to whichever sub-app happened to be iterated last, producing a wrong composite layout with no diagnostic.

Suggested fix: build a class-keyed map first, then warn (or raise) on a `__name__` collision when collapsing to the string-keyed map:

```python
result_by_cls: Dict[Any, str] = {}
for cls in round_classes:
    sub = FSMSpecificationLoader._extract_skill_name(getattr(cls, '__module__', ''))
    if sub:
        result_by_cls[cls] = sub

result: Dict[str, str] = {}
for cls, sub in result_by_cls.items():
    if cls.__name__ in result and result[cls.__name__] != sub:
        logging.warning(
            'Round name collision: %s in %s and %s', cls.__name__, result[cls.__name__], sub,
        )
    result[cls.__name__] = sub
```

autonomy/analyse/abci/app_spec.py
```diff
@@ -123,36 +140,229 @@ def dump_yaml(dfa: "DFA", file: Path) -> None:
...
+                        f"    {s1} --> {s2}: <center>{label}</center>",
                         file=fp,
                     )
+            for (s1, s2), events in top_level.items():
+                if s1 not in start_states:
+                    label = "<br />".join(sorted(events))
+                    print(
+                        f"    {s1} --> {s2}: <center>{label}</center>",
+                        file=fp,
+                    )
```
OjusWiZard, line 299: **[HIGH] Non-deterministic Mermaid output.**

`internal_per_subapp` is iterated via `sorted(...)` (line 234) — good. But these two passes over `top_level.items()` are not sorted; their order depends on the insertion order of `dfa.transition_func`. A trivial source reordering anywhere in the FSM definition will silently shuffle Mermaid lines and produce spurious diffs in `tox -e check-hash` / `tox -e check-api-docs` / `tox -e check-packages`, which are blocking checks per `CLAUDE.md`.

Fix: wrap both passes in `sorted(top_level.items())`. (Or collapse to a single sorted pass keyed on `(s1 not in start_states, s1, s2)`.)

autonomy/analyse/abci/app_spec.py
```diff
@@ -123,36 +140,229 @@ def dump_yaml(dfa: "DFA", file: Path) -> None:
             yaml.safe_dump(dfa.generate(), fp, indent=4)
 
     @staticmethod
-    def dump_mermaid(dfa: "DFA", file: Path) -> None:
-        """Dumps this DFA spec. to a file in Mermaid format."""
+    def dump_mermaid(
+        dfa: "DFA",
+        file: Path,
+        abci_app_cls: Optional[Any] = None,
+        dev_skills: Optional[Set[str]] = None,
```
OjusWiZard, line 155: **[HIGH] `Optional[Any]` defeats strict mypy.**

`tox -e mypy` runs with `--disallow-untyped-defs`, but `Any` short-circuits all of it. The body accesses `transition_function`, `initial_states`, `final_states` — all attributes of `AbciApp`. A typo in any of those names (or a wrong class passed in by a future caller) wouldn't be caught.

Fix: replace with `Optional[Type["AbciApp"]]` under a `TYPE_CHECKING` guard (to avoid the `packages.*` import cycle), or define a small `Protocol` with the three attributes. This issue applies equally to `_build_round_to_subapp` (line 300) and `dump` (line 349).

autonomy/analyse/abci/app_spec.py
```diff
@@ -123,36 +140,229 @@ def dump_yaml(dfa: "DFA", file: Path) -> None:
...
+        for init_cls in getattr(abci_app_cls, "initial_states", set()) or set():
+            round_classes.add(init_cls)
+        for final_cls in getattr(abci_app_cls, "final_states", set()) or set():
+            round_classes.add(final_cls)
+
+        result: Dict[str, str] = {}
+        for cls in round_classes:
+            module = getattr(cls, "__module__", "")
+            sub_app = FSMSpecificationLoader._extract_skill_name(module)
+            if sub_app:
```
OjusWiZard: **[HIGH] Round classes outside `packages.*.skills.*` paths are silently dropped.**

`if sub_app:` excludes any round whose `__module__` lacks a `skills` segment (e.g. monkeypatched test rounds, `src/` layouts, locally aliased imports). The transition then renders as a floating top-level node with no composite box and no log message — the user has no signal that the diagram is incomplete.

Fix: after the build loop, log a single warning naming the unclassified rounds:

```python
unclassified = {c.__name__ for c in round_classes if c.__name__ not in result}
if unclassified:
    logging.warning('Unclassified rounds (no `skills` segment in __module__): %s', sorted(unclassified))
```

autonomy/cli/helpers/fsm_spec.py
```diff
@@ -34,6 +35,37 @@
...
+    :return: Set of dev skill names, or ``None`` if no ``packages.json`` found.
+    """
+    current = package_path.resolve()
+    while current != current.parent:
+        if current.name == "packages" and (current / "packages.json").is_file():
+            try:
+                data = json.loads(
+                    (current / "packages.json").read_text(encoding="utf-8")
+                )
+            except (OSError, ValueError):
```
OjusWiZard: **[MEDIUM] `except (OSError, ValueError)` collapses two distinct failure modes.**

`json.JSONDecodeError` is a subclass of `ValueError`, so a malformed `packages.json` is silently treated identically to "file not readable" — both cause `dump_mermaid` to fall back to the flat per-round diagram with no message. A developer with a corrupt or partially-written `packages.json` (or a CI pipeline regenerating after a botched `autonomy packages lock`) gets a degraded diagram and zero signal.

Fix: split the cases and log on parse failure:

```python
try:
    data = json.loads((current / 'packages.json').read_text(encoding='utf-8'))
except OSError:
    return None
except json.JSONDecodeError as exc:
    logging.warning('Malformed packages.json at %s: %s', current / 'packages.json', exc)
    return None
```

autonomy/analyse/abci/app_spec.py
```diff
@@ -111,8 +116,20 @@ def load(
...
+        DeprecationWarning on use. Will be removed in a future release.
 
+        :param dfa: DFA object to serialize.
+        :param file: Output file path.
+        """
+        warnings.warn(
+            "fsm-specs JSON output is deprecated and will be removed in a "
+            "future release; use --yaml or --mermaid instead.",
+            DeprecationWarning,
+            stacklevel=2,
```
OjusWiZard: **[MEDIUM] `stacklevel=2` mis-attributes the warning when called via `dump()`.**

`stacklevel=2` is correct only for direct `dump_json()` calls. The production call path is `user_code → dump → dump_json` (see line 370), so the warning will appear to come from inside the library at the `dump_json` dispatch line, not from the user's `dump(spec_format=JSON)` call.

Fix: emit the warning in `dump()` before dispatching, with `stacklevel=2` at that level — and remove (or keep `stacklevel=2` on) the call here for direct callers. Alternatively, use `stacklevel=3` here and document it.

autonomy/analyse/abci/app_spec.py
```diff
@@ -123,36 +140,229 @@ def dump_yaml(dfa: "DFA", file: Path) -> None:
...
+
+        # When the FSM lives in a single sub-app (e.g. per-skill analysis)
+        # or we have no dev/third-party split, fall back to the flat
+        # per-round diagram -- no composites, no collapsing.
+        flatten = (
+            abci_app_cls is None
+            or dev_skills is None
+            or len(round_to_subapp) == 0
+            or (len(dev_subapps) + len(third_party_subapps)) <= 1
+        )
```
OjusWiZard, line 203: **[MEDIUM] Silent fall-back with no diagnostic when `flatten` is computed `True`.**

Four conditions OR'd together silently flatten the diagram. The two interesting cases (`abci_app_cls` was supplied but `dev_skills` came back `None` due to a broken `packages.json`; or `len(round_to_subapp) == 0` because every round's `__module__` failed `_extract_skill_name`) both indicate the user wanted the composition view and didn't get it — but there is no log line saying so.

Fix: when `flatten` is `True` *and* `abci_app_cls is not None`, emit a single `logging.info` (or `click.echo` at the CLI layer) explaining which condition triggered the fall-back.

autonomy/cli/helpers/fsm_spec.py
```diff
@@ -34,6 +35,37 @@
...
+    of ``<skill_name>`` values. Returns ``None`` if no ``packages.json``
+    is found, in which case the Mermaid dump falls back to the flat
+    per-round diagram (no collapsing).
+
+    :param package_path: Path to the package being analysed.
+    :return: Set of dev skill names, or ``None`` if no ``packages.json`` found.
+    """
+    current = package_path.resolve()
+    while current != current.parent:
+        if current.name == "packages" and (current / "packages.json").is_file():
```
OjusWiZard, line 54: **[MEDIUM] Walk early-exits at the first `packages`-named ancestor.**

If a user has any intermediate directory called `packages` above the analysed skill (e.g. `~/packages/myproject/packages/valory/skills/foo_abci`), the walk matches the outer `packages` first, finds no `packages.json` inside it, and then the loop continues — but `current` is now the outer `packages`, and on the next iteration we ascend to its parent. We never re-check the inner `packages/` layer because the loop only inspects ancestors, not descendants.

Wait — re-reading: the loop walks parents of `package_path`, so it ascends *up* from the skill. The skill is `…/packages/valory/skills/foo_abci`, so we'd hit `…/packages` correctly. The genuine risk is a parent path also named `packages` *above* the project root (rare in practice, but possible). The bigger concern: if a `packages` dir exists with no `packages.json`, the `if` body is skipped and we silently keep ascending — this is fine, but if the user expected us to find the project, they'll get the silent flat-diagram fall-back. Consider logging at debug level when no `packages.json` is found anywhere in the walk.

autonomy/analyse/abci/app_spec.py
```diff
@@ -123,36 +140,229 @@ def dump_yaml(dfa: "DFA", file: Path) -> None:
...
+        for cls in round_classes:
+            module = getattr(cls, "__module__", "")
+            sub_app = FSMSpecificationLoader._extract_skill_name(module)
+            if sub_app:
+                result[cls.__name__] = sub_app
+        return result
+
+    @staticmethod
+    def _extract_skill_name(module_path: str) -> Optional[str]:
+        """Pull the skill name out of a `packages.<author>.skills.<skill>...` path."""
```
OjusWiZard: **[MEDIUM] Docstring missing Sphinx `:param:` and `:return:` tags — will fail `tox -e darglint`.**

Every other new method in this PR has full Sphinx blocks; this one-liner is the exception. `darglint` is part of `tomte check-code` (mandatory pre-commit per `CLAUDE.md`) and will reject this. Also, the example in `_build_round_to_subapp`'s docstring (line 303) shows `update_bets` as the mapped key, but the actual map key is `cls.__name__` — i.e. `UpdateBetsRound`, not `update_bets`. Worth fixing in the same pass.

Fix:
```python
"""Pull the skill name out of a `packages.<author>.skills.<skill>...` path.

:param module_path: Dotted module path string.
:return: The skill name segment, or ``None`` if the path has no ``skills`` segment.
"""
```

autonomy/analyse/abci/app_spec.py
```diff
@@ -123,36 +140,229 @@ def dump_yaml(dfa: "DFA", file: Path) -> None:
...
+        Falls back to the flat per-round diagram when ``abci_app_cls`` is
+        ``None``, when ``dev_skills`` is ``None`` (i.e. no packages.json
+        info available), or when all rounds belong to a single sub-app.
+
+        :param dfa: DFA object to render.
+        :param file: Output file path.
+        :param abci_app_cls: Optional composed AbciApp class used to classify
+            rounds by sub-app for the composition-aware view.
+        :param dev_skills: Optional set of dev skill names (from
+            ``packages.json``); sub-apps not in this set are collapsed.
```
OjusWiZard, line 177: **[MEDIUM] `Optional[Set[str]]` tri-state diverges from the documented behaviour.**

Docstring (line 159–161) says the renderer falls back to the flat diagram when `dev_skills` is `None`. But the code distinguishes `None` from `set()`: an empty set passes the `dev_skills is None` guard, every sub-app falls into `third_party_subapps`, and (assuming `len(third_party_subapps) > 1`) we render every sub-app as a *collapsed* third-party box rather than flattening. That's a valid output but not what the docstring promises.

Fix one of two ways:
1. Make `_load_dev_skill_names` always return a `frozenset` (empty when no `packages.json`) and drop the `None` branch entirely. The "no info" case then renders identically to "zero dev skills" — consistent.
2. Add `or len(dev_skills) == 0` to the `flatten` predicate so an empty set also falls back.

OjusWiZard (changes requested):
Multi-perspective review (code-reviewer, python-reviewer, pr-test-analyzer, silent-failure-hunter, type-design-analyzer, code-simplifier, comment-analyzer). Inline comments cover **CRITICAL × 1**, **HIGH × 4**, and **MEDIUM × 6** findings. LOW-severity items (naming, double-`split`, two-pass loop collapse, dict-driven `classDef`, etc.) are deferred to a follow-up.

**Top blockers:**
1. No tests for any new code path; existing `test_generate_mermaid` is a tautological round-trip.
2. `_build_round_to_subapp` is keyed on `cls.__name__`, which collides on `FinishedTransactionSubmissionRound` (exists in two sub-apps).
3. Top-level transition output is non-deterministic (will dirty CI hash checks).
4. `Optional[Any]` for `abci_app_cls` defeats strict mypy.

autonomy/analyse/abci/app_spec.py
```diff
@@ -123,45 +148,284 @@ def dump_yaml(dfa: "DFA", file: Path) -> None:
...
                     print(
-                        f"    {s1} --> {s2}: <center>{'<br />'.join(t_set)}</center>",
+                        f"        {s1} --> {s2}: <center>{label}</center>",
                         file=fp,
                     )
-
-            # (2) Then, print the rest of the arrows.
-            for (s1, s2), t_set in aux_map.items():
-                if s1 not in dfa.start_states:
+                print("    }", file=fp)
```
OjusWiZard: **[HIGH] Empty dev composite block — same invalidity hazard the third-party path guards against.**

When a dev sub-app contains rounds but has **zero internal transitions** (every edge from its rounds leads to a different sub-app), `internal_per_subapp.get(sub_app, {})` is empty and this loop emits:

```
state dev_skill_abci {
}
```

The PR description acknowledges "Empty composite-state blocks aren't valid Mermaid in stateDiagram-v2, hence the explicit `"..."` inner state in collapsed boxes" — but that placeholder is only emitted in the third-party loop below. Dev sub-apps with no internal transitions hit the same renderer failure.

When does this happen? A dev sub-app whose rounds are all entry/exit points to other sub-apps — e.g. a thin router skill, or any sub-app contributing only `initial_states` / `final_states` rounds to the composition.

Fix: mirror the third-party guard.

```python
for sub_app in sorted(dev_subapps):
    internal = internal_per_subapp.get(sub_app, {})
    print(f"    state {sub_app} {{", file=fp)
    if internal:
        for (s1, s2), events in sorted(internal.items()):
            label = "<br />".join(sorted(events))
            print(f"        {s1} --> {s2}: <center>{label}</center>", file=fp)
    else:
        print(f'        state "..." as {sub_app}_collapsed', file=fp)
    print("    }", file=fp)
```

Also a test gap: `test_composition_view` only exercises a dev sub-app that *has* internal transitions. A unit test with a dev sub-app whose rounds only have cross-app edges would catch this.

OjusWiZard (commented):
Second-pass review after the fixes from c0b5b8f82 — all 11 prior comments are correctly addressed. Six fresh items worth raising; the two HIGH ones are mermaid-validity / silently-invisible-degradation issues that ship today.

autonomy/analyse/abci/app_spec.py
```diff
@@ -123,45 +148,284 @@ def dump_yaml(dfa: "DFA", file: Path) -> None:
...
+            if not dev_skills:
+                reasons.append("no dev_skills info (packages.json not found)")
+            if len(round_to_subapp) == 0:
+                reasons.append("no rounds could be classified by sub-app")
+            if (len(dev_subapps) + len(third_party_subapps)) <= 1:
+                reasons.append("all rounds belong to a single sub-app")
+            logging.info(
+                "Composition-aware view not available, using flat diagram: %s",
+                "; ".join(reasons) if reasons else "unknown",
+            )
```
OjusWiZard, line 215: **[HIGH] Composition-fallback diagnostic logs at `INFO`; default CLI level is `WARNING`, so the user never sees it.**

The user explicitly asked for `--mermaid` on a composed app and got a flat diagram instead — that's the exact case this message exists to surface. Python's default logging level is `WARNING`, and `autonomy` doesn't lower it for this code path, so the `INFO` line is invisible in every standard invocation. The renderer silently degrades with no operator signal.

Fix: `logging.warning(...)` — the diagram is objectively less informative than what was requested, that's a warning.

autonomy/analyse/abci/app_spec.py
```diff
@@ -123,45 +148,284 @@ def dump_yaml(dfa: "DFA", file: Path) -> None:
...
+            if sub_app:
+                name = cls.__name__
+                if name in result and result[name] != sub_app:
+                    logging.warning(
+                        "Round name collision: %s appears in %s and %s",
+                        name,
+                        result[name],
+                        sub_app,
+                    )
+                result[name] = sub_app
```
OjusWiZard: **[MEDIUM] Round-name-collision warning fires but the map is still overwritten — last-seen sub-app silently wins.**

`result[name] = sub_app` on this line runs unconditionally after the warning at 355. The previously-recorded owner is dropped without trace; downstream `_display()` will then collapse/expand the colliding round under whichever sub-app happened to be iterated last (insertion-ordered in CPython, but semantically arbitrary across `set` iteration).

The warning text names both owners, but the data structure silently commits to the latter — that's the worst of both worlds: the operator sees a log line saying "there's a collision" but has no way to know which sub-app the renderer is actually using.

Fix: pick a deterministic policy. Either keep first-seen (`continue` inside the `if` branch, log which one was retained) or raise `DFASpecificationError`. Both are better than overwriting with a warning that suggests the correct mapping is recorded.

autonomy/cli/helpers/fsm_spec.py
```diff
@@ -34,6 +36,44 @@
...
+            packages_json = current / "packages.json"
+            try:
+                data = json.loads(packages_json.read_text(encoding="utf-8"))
+            except OSError:
+                return None
```
OjusWiZard, line 62: **[HIGH] `OSError` returns `None` silently — no log at any level.**

The `JSONDecodeError` branch right below this got a warning in the previous review pass; the `OSError` branch is still bare. An operator running `autonomy analyse fsm-specs` on a repo where `packages.json` exists but is unreadable (permissions, NFS hiccup, partial write mid-`autonomy packages lock`) gets a silently-degraded flat diagram with no signal at all.

Fix:
```python
except OSError as exc:
    logging.warning("Cannot read packages.json at %s: %s", packages_json, exc)
    return None
```

autonomy/cli/helpers/fsm_spec.py
```diff
@@ -34,6 +36,44 @@
...
+            except json.JSONDecodeError as exc:
+                logging.warning("Malformed packages.json at %s: %s", packages_json, exc)
+                return None
+            return {
+                key.split("/")[2]
+                for key in data.get("dev", {})
+                if key.startswith("skill/") and len(key.split("/")) >= 3
```
OjusWiZard: **[MEDIUM] Length guard `>= 3` is looser than the documented key format.**

Docstring on line 42-44 documents the dev-section key shape as `skill/<author>/<skill_name>/<version>` (4 segments), but the guard accepts 3 segments. A corrupt entry like `skill//abci_name` (empty author — invalid but technically parseable) or a 3-segment key would silently produce a phantom dev-skill name and bypass the documented invariant.

Real `packages.json` from `autonomy packages lock` always emits 4-segment keys, so this is more about defending the invariant than fixing a known crash, but the guard cost is zero:

```python
if key.startswith("skill/") and len(key.split("/")) >= 4
```

Optional: extract `author, skill_name = parts[1], parts[2]` and reject if either is empty.

tests/test_autonomy/test_cli/test_analyse/test_specs.py
```diff
@@ -503,3 +507,351 @@ def test_name_verification(self) -> None:
...
+        assert "classDef devGroup" in content
+        assert "classDef macro" in content
+
+    def test_deterministic_output(self, tmp_path: Path) -> None:
+        """Output is deterministic across multiple runs."""
+        out1 = tmp_path / "run1.md"
+        out2 = tmp_path / "run2.md"
+        FSMSpecificationLoader.dump_mermaid(self.simple_dfa, out1)
+        FSMSpecificationLoader.dump_mermaid(self.simple_dfa, out2)
+        assert out1.read_text() == out2.read_text()
```
OjusWiZard, line 881: **[MEDIUM] Determinism test only exercises the flat path; composition-aware `sorted()` calls are never stressed.**

This test calls `dump_mermaid(self.simple_dfa, out)` with neither `abci_app_cls` nor `dev_skills`, so it only hits the flat branch. The composition-aware path has its own `sorted()` calls on `dev_subapps` (line 253), `third_party_subapps` (line 267), `internal_per_subapp[sub_app]` items (line 255), and both `top_level` passes (lines 276, 283) — none of those sort points are determinism-tested.

`valory-coding-agent[bot]` already raised this point on the prior round and it's still open. Adding a sibling test that runs `dump_mermaid(dfa, out, abci_app_cls=MockApp, dev_skills={...})` twice on a non-trivial composed mock (≥2 dev sub-apps, ≥2 third-party sub-apps, multiple events per arrow) and asserts byte-identical output would close the gap.

```python
def test_deterministic_output_composition(self, tmp_path: Path) -> None:
    """Composition-aware output is also deterministic across runs."""
    # ... build MockApp with multiple dev + third-party sub-apps ...
    out1, out2 = tmp_path / "a.md", tmp_path / "b.md"
    for out in (out1, out2):
        FSMSpecificationLoader.dump_mermaid(
            dfa, out, abci_app_cls=MockApp, dev_skills={"dev_a", "dev_b"},
        )
    assert out1.read_text() == out2.read_text()
```

autonomy/analyse/abci/app_spec.py
```diff
@@ -123,45 +148,310 @@ def dump_yaml(dfa: "DFA", file: Path) -> None:
...
+        flatten = (
+            abci_app_cls is None
+            or not dev_skills
+            or len(round_to_subapp) == 0
+            or (len(dev_subapps) + len(third_party_subapps)) <= 1
+        )
+        if flatten and abci_app_cls is not None:
+            reasons = []
+            if not dev_skills:
+                reasons.append("no dev_skills info (packages.json not found)")
```
OjusWiZard, line 207: **[MEDIUM] Fallback reason is misleading when `_load_dev_skill_names` returned `set()` rather than `None`.**

Both `dev_skills is None` (no `packages.json` found) and `dev_skills == set()` (file found but `dev` block was empty) hit `not dev_skills` at line 200, then land here with the message `"no dev_skills info (packages.json not found)"`. The second case isn't "not found" — the file was located and parsed successfully.

An operator who edited their `packages.json` and ended up with an empty dev set will read this warning and start hunting for a missing file that was actually right there. Split the two cases:

```python
if dev_skills is None:
    reasons.append("no dev_skills info (packages.json not found)")
elif not dev_skills:
    reasons.append("no dev skills declared in packages.json")
```

autonomy/analyse/abci/app_spec.py
```diff
@@ -123,45 +148,310 @@ def dump_yaml(dfa: "DFA", file: Path) -> None:
...
+            s1_sub = round_to_subapp.get(s1)
+            s2_sub = round_to_subapp.get(s2)
+            if s1_sub is not None and s1_sub == s2_sub:
+                internal_counts[s1_sub] = internal_counts.get(s1_sub, 0) + 1
+
+        dev_subapps: Set[str] = set()
+        third_party_subapps: Set[str] = set()
+        for sub_app in set(round_to_subapp.values()):
+            is_dev = dev_skills is not None and sub_app in dev_skills
+            if is_dev and internal_counts.get(sub_app, 0) > 0:
```
OjusWiZard, line 193: **[MEDIUM] Dev skill silently demoted to third-party when it has no internal transitions — no log signal names which skill was demoted.**

When `internal_counts.get(sub_app, 0) == 0`, a skill the user explicitly placed in their `dev` set is reclassified into `third_party_subapps` and rendered as a collapsed box. The intent (per the comment on line 179: "single-round shared recovery skills") is reasonable, but the demotion produces no operator signal: the user requested an expanded view for their dev skill and got a collapsed one with no diagnostic.

If the demotion was correct (single-round skill) it's noise; if it was wrong (developer just forgot a transition), it's an invisible-degradation bug. Either way, naming the demoted skill restores the audit trail:

```python
else:
    third_party_subapps.add(sub_app)
    if is_dev:
        logging.debug(
            "Dev skill %s has no internal transitions; collapsing as third-party",
            sub_app,
        )
```

`debug` is fine here — the operator can opt in via `--log-level=DEBUG` when troubleshooting why an expected dev box is collapsed.

tests/test_autonomy/test_cli/test_analyse/test_specs.py
```diff
@@ -503,3 +507,527 @@ def test_name_verification(self) -> None:
...
+        content = out.read_text()
+
+        # Dev sub-app is expanded with internal transitions
+        assert "state dev_skill_abci {" in content
+        assert "RoundARound --> RoundBRound" in content
+
+        # Third-party sub-app is collapsed with a hidden placeholder child
+        assert "state tp_skill_abci {" in content
+        assert "tp_skill_abci_hidden" in content
+        assert "classDef hiddenInner" in content
```
OjusWiZard, line 869: **[MEDIUM] `test_composition_view` checks for the hidden-placeholder presence but never asserts the empty-block form is absent.**

Mermaid `stateDiagram-v2` rejects empty composite blocks (`state X { }`); the implementation works around this by emitting `state " " as X_hidden` inside every collapsed third-party box. If that placeholder logic regresses (e.g. someone removes the inner `state " " as ...` line), Mermaid will silently fail to render the diagram on GitHub/VS Code, but this test still passes because lines 868-869 only check for *presence* of expected strings, not *absence* of forbidden ones.

Add a negative assertion that the third-party block is non-empty:

```python
lines = content.splitlines()
tp_open = next(i for i, l in enumerate(lines) if "state tp_skill_abci {" in l)
tp_close = tp_open + next(
    j for j, l in enumerate(lines[tp_open + 1:], start=1) if l.strip() == "}"
)
block_body = lines[tp_open + 1: tp_close]
assert any("_hidden" in l for l in block_body), (
    f"empty composite block is invalid Mermaid: {block_body!r}"
)
```

autonomy/cli/helpers/fsm_spec.py
```diff
@@ -34,6 +36,51 @@
...
+                return None
+            except json.JSONDecodeError as exc:
+                logging.warning("Malformed packages.json at %s: %s", packages_json, exc)
+                return None
+            # Real keys are 4-segment: skill/<author>/<skill_name>/<version>.
+            return {
+                key.split("/")[2]
+                for key in data.get("dev", {})
+                if key.startswith("skill/")
+                and len(key.split("/")) >= 4
```
OjusWiZard, line 71: **[MEDIUM] `key.split("/")` evaluated up to three times per key — collapse with a walrus.**

The comprehension calls `key.split("/")` in the `len(...)` guard and again twice for `[1]` and `[2]` indexing. For a real `packages.json` with dozens of dev entries this is a constant-factor inefficiency, but the bigger reason to fix it is readability: the four-segment invariant is split across four `and` clauses that all re-derive the same list.

```python
return {
    parts[2]
    for key in data.get("dev", {})
    if key.startswith("skill/")
    and (parts := key.split("/"))
    and len(parts) >= 4
    and parts[1]
    and parts[2]
}
```

Walrus is fine on Python 3.8+. (mypy is happy with it.)

OjusWiZard (approved):
Third-pass review after `1d734d71a`. Approving with 4 MEDIUM nits left as inline comments — none ship-blocking, all small enough for a follow-up cleanup PR. Items I considered and intentionally did not raise:

- `_load_dev_skill_names` does not defend against `{"dev": null}` — `autonomy packages lock` always emits `dev` as a dict, so the only path is manual corruption (which open-aea itself does not gracefully handle either).
- Bare `_display(state)` for unclassified states only triggers if `dfa.transition_func` and `abci_app_cls.transition_function` diverge — a theoretical bug elsewhere.
- `hidden_subapps` / `internal_per_subapp.get(sub_app, {})` are dead-code leftovers from the latest commit's branch removal — cleanup-only.

autonomy/analyse/abci/app_spec.py
```diff
@@ -123,45 +148,310 @@ def dump_yaml(dfa: "DFA", file: Path) -> None:
...
+                    f"    class {','.join(sorted(dev_subapps))} devGroup",
+                    file=fp,
+                )
+            if not flatten and third_party_subapps:
+                print(
+                    "    classDef macro "
+                    "fill:#f5f9f5,"
+                    "stroke:#2e7d32,"
+                    "stroke-width:3px,"
+                    "font-weight:bold",
```
OjusWiZard, line 320: **[HIGH, non-blocking] `macro` classDef is now visually indistinguishable from `devGroup` — kills the dev/third-party visual split.**

```
devGroup: fill:#f5f9f5, stroke:#2e7d32, stroke-width:2px, font-weight:bold
macro:    fill:#f5f9f5, stroke:#2e7d32, stroke-width:3px, font-weight:bold
```

Same green fill, same green stroke — the only differentiator is `2px` vs `3px` border width, which is barely perceptible at GitHub's default zoom. The PR description explicitly advertises "Styled with a **blue** `macro` classDef + thicker border", and an earlier review-pass version of this file did have `fill:#eef2ff, stroke:#1e3a8a` (blue). One of the recent rebase/format commits regressed it to a near-duplicate of `devGroup`.

Fix: restore the blue palette for `macro`:

```python
"    classDef macro "
"fill:#eef2ff,"
"stroke:#1e3a8a,"
"stroke-width:3px,"
"font-weight:bold"
```

And harden `test_composition_view` so this can't regress silently — see paired comment on the test file.

tests/test_autonomy/test_cli/test_analyse/test_specs.py
```diff
@@ -503,3 +507,527 @@ def test_name_verification(self) -> None:
...
+
+        # Styling
+        assert "classDef devGroup" in content
+        assert "classDef macro" in content
```
OjusWiZard, line 873: **[MEDIUM, non-blocking] Two related test gaps, both invisible-regression hazards.**

**1. classDef styling not asserted.** This line and the prior `classDef devGroup` assertion check only that the *directive name* appears in the output, not that the two classDefs have *distinct* fill/stroke values. The visual regression flagged on `app_spec.py:316-320` (both classDefs ended up green) passes this test cleanly. Quick guard:

```python
def _classdef_style(content: str, name: str) -> str:
    line = next(l for l in content.splitlines() if f"classDef {name} " in l)
    return line.split(f"classDef {name} ", 1)[1]

assert _classdef_style(content, "devGroup") != _classdef_style(content, "macro"), (
    "dev and third-party classDefs must be visually distinct"
)
```

**2. Dev-skill demotion path is untested.** `app_spec.py:192-196` documents that a user-declared dev skill with zero internal transitions gets reclassified into `third_party_subapps`. `test_composition_view` only exercises the *kept-as-dev* branch (the dev skill has `RoundARound → RoundBRound` internal). Nothing pins the *demoted* branch — a future refactor could quietly invert the condition and no test would catch it.

Sibling test sketch:

```python
def test_dev_skill_with_no_internal_transitions_collapses(self, tmp_path):
    # dev_skill_abci is declared dev but its only transition is cross-app
    # to tp_skill_abci -> expect dev_skill_abci to render as a collapsed
    # box (third-party styling), not as an expanded composite.
    ...
    out = tmp_path / "test.md"
    FSMSpecificationLoader.dump_mermaid(
        dfa, out, abci_app_cls=MockApp, dev_skills={"dev_skill_abci"},
    )
    content = out.read_text()
    # dev_skill_abci should appear in the macro class list, not devGroup
    macro_line = next(l for l in content.splitlines() if l.startswith("    class ") and " macro" in l)
    assert "dev_skill_abci" in macro_line
```

OjusWiZard (approved):
Third-pass review, approving. Two follow-up items inline — both non-blocking. The first is a visual regression that defeats the dev/third-party distinction the PR is built around (worth fixing as a quick follow-up since `test_composition_view` doesn't catch it). The second is a test-coverage gap for an existing untested branch.

The four threads from my prior round (3228400403/419/423/427) are still marked unresolved on GitHub but I'm not blocking on them — happy to leave them as won't-fix if that's the intent; just worth closing the threads so the PR state matches.

Nice work landing the composition view, the dev-demotion logic, and the patch-coverage push.

tests/test_autonomy/test_cli/test_analyse/test_specs.py
```diff
@@ -503,3 +507,527 @@ def test_name_verification(self) -> None:
...
+                abci_app_cls=MockApp,
+                dev_skills={"dev_a_abci", "dev_b_abci"},
+            )
+        assert out1.read_text() == out2.read_text()
+
+
+class TestDumpJsonDeprecation:
+    """Test the JSON deprecation warning."""
+
+    def test_dump_json_deprecation_via_dump(self, tmp_path: Path) -> None:
```
OjusWiZard, line 942: **[MEDIUM] `dump_json()` is never asserted to be warning-free.**

The docstring at `app_spec.py:131-140` explicitly states:

> The ``DeprecationWarning`` is emitted by ``dump()`` so that ``stacklevel`` always points at the external caller.

That split is load-bearing — callers invoking `dump_json` directly must NOT see the warning. But `TestDumpJsonDeprecation` only covers the `dump()` path. A regression that accidentally re-emitted the warning inside `dump_json` (e.g., a future `@deprecated` decorator) would pass the existing test.

Fix:
```python
def test_dump_json_does_not_warn(self, tmp_path: Path) -> None:
    """dump_json() called directly emits no DeprecationWarning."""
    dfa = DFA(...)  # same fixture as the existing test
    with warnings.catch_warnings(record=True) as w:
        warnings.simplefilter("always")
        FSMSpecificationLoader.dump_json(dfa, tmp_path / "spec.json")
    assert not [x for x in w if issubclass(x.category, DeprecationWarning)]
```

tests/test_autonomy/test_cli/test_analyse/test_specs.py
```diff
@@ -503,3 +507,527 @@ def test_name_verification(self) -> None:
...
+
+        out = tmp_path / "test.md"
+        with caplog.at_level(logging.INFO):
+            FSMSpecificationLoader.dump_mermaid(
+                self.simple_dfa, out, abci_app_cls=MockApp, dev_skills=set()
+            )
+        content = out.read_text()
+        # Should be flat (no composite state blocks)
+        assert "classDef devGroup" not in content
+        assert "Composition-aware view not available" in caplog.text
```
OjusWiZard, line 680: **[MEDIUM] `test_flat_when_dev_skills_empty` and `test_flat_when_dev_skills_none` assert caplog but not file content.**

Both tests verify the WARNING is logged but never assert the rendered file is actually flat. `test_flat_when_no_abci_app_cls` (L644-L652) already established the pattern:
```python
assert "state " not in content.replace("stateDiagram", "")
```

Reusing it here would catch a regression that emits spurious `state X { ... }` composite blocks under the dev-skills-empty or dev-skills-none fallback paths. As written, only the absence of `classDef devGroup` is checked (L679) — but a regression could still emit composite blocks without that classDef.

Fix: add the same `content` assertion (or equivalent) to both `test_flat_when_dev_skills_empty` and `test_flat_when_dev_skills_none`.

tests/test_autonomy/test_cli/test_analyse/test_specs.py
```diff
@@ -503,3 +507,527 @@ def test_name_verification(self) -> None:
...
+
+        class MockApp(_MockAbciApp):
+            """Mock composed app with collision."""
+
+            transition_function = {round_a: {round_b: round_b}}
+            initial_states = set()
+            final_states = set()
+
+        with caplog.at_level(logging.WARNING):
+            FSMSpecificationLoader._build_round_to_subapp(MockApp)
```
OjusWiZard, line 594: **[MEDIUM] `test_collision_warning` discards the return value — kept-winner isn't asserted.**

The call discards `_build_round_to_subapp(MockApp)`'s return. Only the log is checked. A regression that flips first-seen vs last-seen winner (or alphabetises by `(cls.__name__, cls.__module__)`) would still emit the warning and pass this test silently.

This pairs with the still-open production-side thread 3228400419 (silent overwrite) — the production fix and the test assertion should land together so the kept value is locked down by a contract, not by accident.

Fix:
```python
result = FSMSpecificationLoader._build_round_to_subapp(MockApp)
assert "Round name collision" in caplog.text
assert "SharedRound" in caplog.text
assert result["SharedRound"] == "skill_a"  # or whichever the deterministic policy keeps
```

autonomy/analyse/abci/app_spec.py
```diff
@@ -123,45 +148,310 @@ def dump_yaml(dfa: "DFA", file: Path) -> None:
...
+            or (len(dev_subapps) + len(third_party_subapps)) <= 1
+        )
+        if flatten and abci_app_cls is not None:
+            reasons = []
+            if not dev_skills:
+                reasons.append("no dev_skills info (packages.json not found)")
+            if len(round_to_subapp) == 0:
+                reasons.append("no rounds could be classified by sub-app")
+            if (len(dev_subapps) + len(third_party_subapps)) <= 1:
+                reasons.append("all rounds belong to a single sub-app")
```
OjusWiZard, line 211: **[LOW] `"all rounds belong to a single sub-app"` reason string is never asserted in any test.**

The other two reasons each have a test:
- `"no dev_skills info (packages.json not found)"` — covered by `test_flat_when_dev_skills_empty`/`_none`
- `"no rounds could be classified by sub-app"` — covered by `test_flat_when_no_rounds_classifiable`

This branch (single classifiable sub-app, non-empty `dev_skills`) is reachable in real usage — a skill in a private repo where only one sub-app has been migrated to a `packages.json` dev entry — but the exact diagnostic string is unguarded. A typo or rewording would not be caught.

Fix: one-round-one-sub-app fixture with `dev_skills={"that_skill"}`; `assert "all rounds belong to a single sub-app" in caplog.text`.

autonomy/analyse/abci/app_spec.py
```diff
@@ -123,45 +148,310 @@ def dump_yaml(dfa: "DFA", file: Path) -> None:
...
+                "Composition-aware view not available, using flat diagram: %s",
+                "; ".join(reasons) if reasons else "unknown",
+            )
+
+        def _display(state: str) -> str:
+            """Collapse third-party rounds to their sub-app name."""
+            if flatten:
+                return state
+            sub = round_to_subapp.get(state)
+            if sub in third_party_subapps:
```
OjusWiZard, line 222: **[LOW] `_display` returns `sub` (typed `Optional[str]`) from a `-> str` function — relies on mypy narrowing through `in Set[str]`.**

Mypy 1.19 currently narrows `Optional[str] in Set[str]` membership to `str` in the body, so `tox -e mypy` is clean. But the narrowing is a mypy convenience — readers (and future mypy versions / Pyright) may not get the same treatment, and the existing `if s1_sub is not None and s1_sub == s2_sub` style at L186, L231 uses the explicit guard.

Fix (purely defensive, no semantic change):
```python
sub = round_to_subapp.get(state)
if sub is not None and sub in third_party_subapps:
    return sub
return state
```
Aligns with the explicit-guard pattern used elsewhere in this function and removes any reliance on container-membership narrowing.

OjusWiZard (approved):
Fourth-pass review, approving. Five fresh items inline — three MEDIUM test-coverage gaps and two LOW polish notes. None ship-blocking; bundled cleanly with the four MEDIUM threads still open from prior rounds for a single follow-up PR.

Multi-perspective sweep (python-reviewer, silent-failure-hunter, pr-test-analyzer, code-simplifier) deduped against the 22 existing review threads; remainder verified by reading the code at HEAD.

Dropped during triage (for record):
- Intra-third-party self-loop "not dropped" — invalid; the `continue` at L236 already short-circuits all `s1_sub == s2_sub` pairs.
- f-strings in `logging.warning` at L575-578 — pre-existing code untouched by this PR.
- `hidden_subapps` redundancy and the `_load_dev_skill_names` empty-author key filter — already acknowledged in your third-pass approve, or edge cases unlikely to hit a real `packages.json`.

DhairyaPatel7:
Fresh-eye review on 1d734d71a. Prior threads from @OjusWiZard cover the blocking ground correctly and the responses look sound — no additional blockers from my side.

One non-blocking heads-up worth surfacing on the PR description: the new `<br />.join(sorted(events))` in `dump_mermaid` makes event-label ordering in Mermaid output deterministic (previously: set-insertion order). The change is correct, but every downstream repo's first regen of `fsm_specification.md` after this lands will show event-order churn even where the underlying FSM didn't change. Worth a one-liner in the PR body / release notes so trader / market-* maintainers aren't surprised.

Open follow-ups in the still-active threads (dev/third-party visual parity, dev-demotion logging, `dump_json` warning-free test, kept-winner assert, etc.) are fine as a follow-up cleanup PR.

DhairyaPatel7 (approved):

autonomy/analyse/abci/app_spec.py
```diff
@@ -123,45 +148,286 @@ def dump_yaml(dfa: "DFA", file: Path) -> None:
...
+            round_classes.add(src)
+            for dst in transitions.values():
+                round_classes.add(dst)
+        for init_cls in getattr(abci_app_cls, "initial_states", set()) or set():
+            round_classes.add(init_cls)
+        for final_cls in getattr(abci_app_cls, "final_states", set()) or set():
+            round_classes.add(final_cls)
+
+        result: Dict[str, str] = {}
+        for cls in round_classes:
```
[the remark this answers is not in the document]
jmoreira-valory: Addressed in a95cbd282 — `_build_round_to_subapp` now iterates `sorted(round_classes, key=lambda c: (c.__name__, c.__module__))`, giving deterministic first-seen collision resolution.

tests/test_autonomy/test_cli/test_analyse/test_specs.py
```diff
@@ -503,3 +507,340 @@ def test_name_verification(self) -> None:
...
+        # Third-party sub-app is collapsed with "..." placeholder
+        assert "state tp_skill_abci {" in content
+        assert '"..."' in content
+        assert "tp_skill_abci_collapsed" in content
+
+        # Styling
+        assert "classDef devGroup" in content
+        assert "classDef macro" in content
+
+    def test_deterministic_output(self, tmp_path: Path) -> None:
```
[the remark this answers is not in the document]
jmoreira-valory: Addressed in 1d734d71a via `test_deterministic_output_composition`, which exercises the composition-aware `sorted()` calls on `dev_subapps`, `third_party_subapps`, `internal_per_subapp`, and both `top_level` passes against a 6-round mock spanning 2 dev + 2 third-party sub-apps.

autonomy/analyse/abci/app_spec.py
```diff
@@ -123,45 +148,286 @@ def dump_yaml(dfa: "DFA", file: Path) -> None:
...
                         file=fp,
                     )
 
+            # Visual styling: dev sub-app boxes get a subtle "group" tint,
+            # third-party get the more pronounced macro-state styling so
+            # the two roles are easy to distinguish at a glance.
+            if dev_subapps:
+                print(
+                    "    classDef devGroup "
+                    "fill:#f5f9f5,"
```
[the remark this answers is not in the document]
jmoreira-valory: Verified rendering in VS Code's Markdown preview (Mermaid native in 1.99+). GitHub's embedded renderer parses `classDef` but may drop styling on older Mermaid versions — graceful degradation: structure stays readable, just unstyled. Acceptable trade-off for the composition view.

autonomy/analyse/abci/app_spec.py
jmoreira-valory: Deferring to follow-up — the diagnostic conflates `dev_skills is None` and `dev_skills == set()` deliberately for now (both are "no usable dev_skills info"). Will differentiate if it confuses operators in practice.

autonomy/analyse/abci/app_spec.py
jmoreira-valory: Intentional: a per-render `WARNING` for every single-round dev skill would spam in steady-state operation, and the demotion is visible in the rendered diff. Open to revisit if it surprises operators.

tests/test_autonomy/test_cli/test_analyse/test_specs.py
jmoreira-valory: Deferring to follow-up — agreed it would lock in the post-bug invariant; capturing in a follow-up to avoid dismissing approval.

autonomy/cli/helpers/fsm_spec.py
jmoreira-valory: Declined — `_load_dev_skill_names` runs once per CLI invocation, not hot. Three-evaluations is a wash against the walrus readability overhead.

autonomy/analyse/abci/app_spec.py
jmoreira-valory: Design call from local iteration: same green family is intentional (third-party is structurally a green-box like dev, just opaque). Differentiation is via `stroke-width:5px` (macro) vs `2px` (dev), which reads at a glance in the VS Code preview. Open to revisit if it confuses users on render.

tests/test_autonomy/test_cli/test_analyse/test_specs.py
jmoreira-valory: Deferring — pinning exact stroke widths/fills in tests churns on every aesthetic tweak. Will revisit if styling regresses.

tests/test_autonomy/test_cli/test_analyse/test_specs.py
jmoreira-valory: Deferring — the "yaml/mermaid emit no DeprecationWarning" contract is implicit in the dispatcher; happy to add explicit assertions in a follow-up.

tests/test_autonomy/test_cli/test_analyse/test_specs.py
jmoreira-valory: Deferring — agreed the file-content assertion would strengthen these. Follow-up.

tests/test_autonomy/test_cli/test_analyse/test_specs.py
jmoreira-valory: Deferring — kept-winner is checked indirectly via the log message text but I agree the return-value assertion is stronger. Follow-up.

autonomy/analyse/abci/app_spec.py
jmoreira-valory: Declined — the branch is reachable via the existing `test_flat_diagram_when_no_abci_app` family and identical in shape to the other two reason cases. Pure coverage padding.

autonomy/analyse/abci/app_spec.py
jmoreira-valory: Declined — mypy 1.x narrows `Optional[str] in Set[str]` correctly here. Switching to an explicit `assert sub is not None` adds runtime cost for a check the type system already proves.

Issue #2502, opened 2026-05-11
FSM audit: squads stabilization, dead transitions, and round-test gaps


DIvyaNautiyal07:
## Summary

An FSM audit of all 21 skills under `packages/valory/skills/` surfaced 19 findings. Most of the impact is concentrated in two skills: `squads_transaction_settlement_abci` has three reachability/correctness bugs and zero tests, and `register_reset_recovery_abci` has an end_block override that can hang indefinitely under no-majority. The rest are test-coverage gaps and minor drifts.

Every finding is tied to a specific file:line; fix snippets are inline. Every claim was verified by re-reading the cited source.

The framework base (`abstract_round_abci`), the ABCI base (`abstract_abci`), and the production library skills (`registration_abci`, `reset_pause_abci`, `transaction_settlement_abci`, `slashing_abci`, `termination_abci`, `offend_abci`, `offend_slash_abci`, `register_reset_abci`, `register_termination_abci`) ran clean against the full checklist. The historical `([],) * n` and `PendingOffencesBehaviour` precedence bugs called out in the audit checklist are both already fixed.

The `autonomy analyse dialogues` CLI surfaced one finding (counter_client missing `abci_dialogues`); that's included as M2 below. `autonomy analyse fsm-specs`, `handlers`, and `docstrings` failed in the audit environment with `No module named 'grpc'`.

---

### 1. `squads_transaction_settlement_abci`: `VerifyTxRound` silently treats failed verification as success

[`packages/valory/skills/squads_transaction_settlement_abci/rounds.py:215-231`](packages/valory/skills/squads_transaction_settlement_abci/rounds.py#L215-L231) — the `end_block` override emits `Event.DONE` on `threshold_reached` regardless of the `verified: bool` field on `VerifyTxPayload`. Agents that all vote `verified=False` (i.e., the transaction did not land on chain) still transition to `FinishedTransactionSubmissionRound`. The source carries a `# TODO: Fix implementation` comment at line 224 acknowledging this is shipping incomplete.

Same override also never returns `Event.NO_MAJORITY`; the wired transition at line 327 is dead.

```python
# fix
def end_block(self) -> Optional[Tuple[BaseSynchronizedData, Enum]]:
    result = super().end_block()
    if result is None:
        return None
    synchronized_data, event = result
    if event != self.done_event:
        return result
    if not self.most_voted_payload_values[0]:  # verified field
[...]

PR #2515, opened 2026-05-14
Audit: FSM correctness fixes for #2502

DIvyaNautiyal07:
## Summary

Audit follow-up from #2502. Eleven fixes across FSM scaffolds, round-test coverage, the analyse CLI, and demo-skill hygiene. Maintainer-approved scope (OjusWiZard, DavidMinarsch): items 5, 6, 7, 8, 9, 10, 11, 12, 13, 17, 18, 19.

Squads (items 1-4, 14-16) are left as is per DM's "leave as is or add tests; fixing without tests has no value." A separate comment will land on #2502 documenting that decision.

### What changed

- `register_reset_recovery_abci`: the skill is a happy-path scaffold (git history confirms via `856de4f`, "test: ... [no ci]"). The no-majority and timeout paths are intentionally absent; recovery is via Tendermint hard reset rather than an FSM transition. The dead `event_to_timeout = {ROUND_TIMEOUT: 30.0}` entry, which advertised a timeout the FSM never honored, is removed. Class docstring states the intent; the `AbciApp` class docstring is regenerated by `tox -e abci-docstrings` to reflect the empty `event_to_timeout`.
- `slashing_abci/payloads.py`: `StatusResetPayload`'s 4th field `slash_tx_sent` is positionally zipped with the 4th `selection_key` `slashing_majority_reached` (in `rounds.py`). The mismatch was functionally fine (both default to `False`) but invisible to a reader searching for `slashing_majority_reached`. A docstring note now records the mapping. The field is **not** renamed: `BaseTxPayload` subclasses are serialized over the consensus protocol, so a rename is a coordinated skill version bump, not a hygiene edit.
- `identify_service_owner_abci`, `funds_forwarder_abci`, and the `StatusResetRound` half of `slashing_abci`: round-level tests now drive the round through `BaseCollectSameUntilThresholdRoundTest` rather than asserting on the transition table dict. Existing tests could not catch a mutation that changed `done_event` because they never invoked `_test_round`. The new tests exercise done, none, and no-majority paths. For `StatusResetRound` the test pins the 4-tuple `selection_key` → DB-key mapping so a future reorder of either tuple is caught.
- `slashing_abci/tests/test_rounds.py::TestSlashingCheckRound`: `_event_class = Event` is now declared. `BaseRoundTestClass` requires it lazily (only `_test_no_majority_event` references it); the next no-majority test added on that class would have failed with `AttributeError` without this declaration.
- `transaction_settlement_abci/test_tools/integration.py`: the private-key file write is now in a `with open(...)` block. Under the previous bare `open() + write() + close()` shape, a write exception would leak the file handle; on Windows that blocks the `TemporaryDirectory` cleanup. The change is pure-additive scaffolding and propagates to mech/mech-predict/.operate-mech on their next `autonomy packages sync`.
- `autonomy/cli/analyse.py`: `counter` and `counter_client` are added to the default `--ignore` list for both `analyse handlers` and `analyse dialogues`. They are non-ABCI demo skills and the project-wide analyse pass previously failed on them. The relevant test (`tests/test_autonomy/test_cli/test_analyse/test_check_handlers.py`) already passed both skills explicitly via `-i`, so the change just brings the CLI default in line with the test expectation.
- `counter/handlers.py`: `decode_number`'s docstring said "little-endian" but the implementation uses `byteorder="big"`. The matching encoder is also big-endian. Docstring fixed.
- `counter_client/utils.py`: `curdatetime` called `datetime.utcnow()` and discarded the result, then returned `datetime.now().isoformat()` (local time). It now returns an ISO 8601 UTC string so log timestamps match the function's apparent intent.
- `counter_client/behaviours.py`: `curdatetime` is now imported from `utils` (where it is defined) rather than from `handlers` (which re-imports it).

### Out of scope

Items 1-4 and 14-16 (`squads_transaction_settlement_abci`) are left as is per DM. The known bugs remain documented in #2502 for a future productionization pass. Verified zero external consumers.

## Test plan

- [x] `tomte format-code`, `tomte check-code` (black, isort, flake8, mypy, pylint, darglint)
- [x] `make security` (bandit, safety, gitleaks)
- [x] `tox -e fix-copyright`, `check-dependencies`
- [x] `make generators` (abci-docstrings, lock-packages, gen-api-docs, fix-doc-hashes)
- [x] `tox -e check-api-docs`, `check-hash`, `check-packages`, `check-abci-docstrings`, `check-abciapp-specs`, `check-handlers`, `vulture`
- [x] pytest on every touched skill and the analyse CLI: 296 passing (9 new round-test methods + counter/counter_client/register_reset_recovery/identify_service_owner/funds_forwarder/slashing suites + 128 transaction_settlement_abci unit tests + analyse CLI tests)
- [x] `autonomy packages lock --check`, `autonomy push-all`

Refs #2502.


autonomy/cli/analyse.py
```diff
@@ -317,6 +317,8 @@ def _parse_logs(
     type=str,
     default=[
         "abstract_abci",
+        "counter",
+        "counter_client",
```
DavidMinarsch: No - these are examples. Should not hardcode example ignores here.

DIvyaNautiyal07: Reverted in 5755a9f95. You're right — the tox.ini check-handlers / check-dialogues envs already pass `-i counter -i counter_client` explicitly, so CI was never failing on these. The CLI default goes back to `["abstract_abci"]`.

packages/valory/skills/slashing_abci/rounds.py
```diff
@@ -201,6 +202,34 @@ class StatusResetRound(CollectSameUntilThresholdRound):
...
+), (
+    "StatusResetPayload data field order drifted from "
+    "StatusResetRound.selection_key (slash_tx_sent maps positionally to "
+    "selection_key slashing_majority_reached); see the StatusResetPayload "
+    "docstring before changing either side."
+)
+assert (
+    _STATUS_RESET_PAYLOAD_DATA_FIELDS[:3] == StatusResetRound.selection_key[:3]
+    and StatusResetRound.selection_key[3] == "slashing_majority_reached"
+), "StatusResetRound.selection_key drifted from StatusResetPayload fields."
```
OjusWiZard: Suggest deriving `selection_key` from `StatusResetPayload` directly so the drift surface disappears entirely (rather than detecting drift after the fact). This also sidesteps the `assert`-stripping concern under `python -O` / `PYTHONOPTIMIZE`.

```python
_PAYLOAD_DATA_FIELDS: Tuple[str, ...] = tuple(
    f.name
    for f in dataclasses.fields(StatusResetPayload)
    if f.name not in {bf.name for bf in dataclasses.fields(BaseTxPayload)}
)
# 4th payload field `slash_tx_sent` is the legacy name retained for protocol
# stability; it is written to the DB under `slashing_majority_reached`.
# Renaming the payload field is a consensus-protocol change.
_DB_KEY_ALIASES: Dict[str, str] = {"slash_tx_sent": "slashing_majority_reached"}
```

Then on `StatusResetRound`:

```python
selection_key = tuple(
    _DB_KEY_ALIASES.get(name, name) for name in _PAYLOAD_DATA_FIELDS
)
```

With this, `selection_key` *is* the payload's field list with one explicit rename — drift is impossible by construction, the rename is data instead of a comment, and adding a 5th payload field flows through automatically.

If you'd rather keep the explicit pin shape, swap the bare `assert` for `if ...: raise AssertionError(...)` so the safety net survives optimized interpreters. The repo doesn't currently invoke `python -O`, but `assert` is silently a no-op under it, which would defeat the import-time guarantee the docstring promises.

DIvyaNautiyal07: Done in 8a1e02ef5. Switched to `_PAYLOAD_DATA_FIELDS` derived from `dataclasses.fields(StatusResetPayload)` + `_DB_KEY_ALIASES = {"slash_tx_sent": get_name(SynchronizedData.slashing_majority_reached)}`. `selection_key` is now `tuple(_DB_KEY_ALIASES.get(name, name) for name in _PAYLOAD_DATA_FIELDS)` — drift is impossible by construction, the rename lives as data, and a 5th payload field flows through automatically. Kept one `raise AssertionError` (not bare `assert`, per your `python -O` note) to catch payload-field renames that would orphan an alias key. The resulting tuple is byte-identical to the previous hardcoded value, so wire format is unchanged.

OjusWiZard (approved):

jmoreira-valory (approved):
Thoroughly pre-reviewed by OjusWiZard (approved). I reviewed independently and have no blocking findings.

The macOS 3.10 CI failure observed is network noise (intermittent timeout fetching packages during sync), not related to this PR's changes.

LGTM.

Issue #2501, opened 2026-05-11
Audit: resilience hardening for HTTP, connections, and runtime


DIvyaNautiyal07:
## Summary

A resilience audit of every external HTTP/RPC/IPFS dependency in the framework surfaced 31 findings across the framework HTTP path, the `ledger` / `ipfs` / `http_client` connections, the localhost Tendermint Flask app, and the `autonomy/chain/*` CLI helpers. Two additional security-adjacent findings have been disclosed separately and are not described here.

Findings are grouped below by priority. Each is tied to a specific file:line; fix snippets are inline. Every claim was verified by re-reading the cited source.

The two highest-impact themes:

1. The framework's `ApiSpecs` and a handful of behaviour helpers in `abstract_round_abci` have uncaught exception paths that crash the entire agent process on malformed-but-plausible upstream responses (non-UTF-8 body, non-coercible string where int expected, malformed Tendermint payload). These are inherited by every downstream service.
2. The `ledger` and `ipfs` connections both share the default loop ThreadPoolExecutor, both have done-callbacks that don't catch `task.result()` exceptions, and the IPFS connection passes no timeout to `run_async`. A hung remote endpoint can permanently strand the requesting skill with no envelope ever returned.

---

### 1. `ApiSpecs.process_response` crashes on non-UTF-8 body

[`packages/valory/skills/abstract_round_abci/models.py:719`](packages/valory/skills/abstract_round_abci/models.py#L719) decodes the response body outside the surrounding try/except. The try/except only catches `JSONDecodeError`. A non-UTF-8 body (binary/gzipped payload, latin-1 error page from a CDN) raises `UnicodeDecodeError`, propagates out of `process_response`, escapes `async_act`, and stops the agent.

```python
# current
decoded_response = response.body.decode()  # raises UnicodeDecodeError outside the try
self.response_info.error_data = None
try:
    response_data = json.loads(decoded_response)
except json.JSONDecodeError:
    ...
```

```python
# fix
try:
    decoded_response = response.body.decode()
except (UnicodeDecodeError, AttributeError):
    self.context.logger.error("Could not decode response body as UTF-8")
    return None
self.response_info.error_data = None
try:
    response_data = json.loads(decoded_response)
except json.JSONDecodeError:
    ...
```

### 2. `ApiSpecs._parse_response_data` crashes on non-coercible type

[...]

Issue #24, opened 2021-08-28
Extract contract helper functionality into contract package


DavidMinarsch:
**Is your feature request related to a problem? Please describe.**
The contract helper functionality in `helpers/gnosis_safe.py` is an anti-pattern.

**Describe the solution you'd like**
Extract into AEA contract package

Issue #22, opened 2021-08-23
Move ConsensusParams class in abstract base skill


marfvr:
**Is your feature request related to a problem? Please describe.**
Should consensus params be defined in the abstract base skill instead?
Non-price estimation specific params should be set in abstract base skill, then extend model in price estimation skill.

Issue #21, opened 2021-08-23
Try to remove 'initial delay' in ABCI round-based skill


marfvr:
**Is your feature request related to a problem? Please describe.**
The "Initial delay" parameter seems a bit hacky.

**Describe the solution you'd like**
Try to get rid of it, with e.g. an health-check of the tendermint node (similarly to how it's done in the docker image via http - although here using the protocols).

PR #2516, opened 2026-05-15
audit: resilience hardening for HTTP, connections, and Tendermint app (#2501)

DIvyaNautiyal07:
## Summary

Implements the resilience audit recommendations from #2501 that maintainers approved or that proved real under verification. Full breakdown of what's shipping versus deferred and the reasoning is in the [status comment](https://github.com/valory-xyz/open-autonomy/issues/2501#issuecomment-4458378020).

## Original audit commits

1. `e7337ca87` — Framework HTTP path, IPFS connection, subgraph client (audit items #1, #2, #5, #6, #9, #10, #11, #12, #15, #16, #17). Includes reproduction tests for #1 and #2 in `test_models.py`; both reproduce as agent crashes pre-fix and pass after the fix.
2. `0586bc245` — Tendermint Flask app (localhost copy), chain CLI helpers, behaviour edge cases (#7, #8, #19, #21, #22, #24). The Flask reset endpoints flip from HTTP 200 to 500 on failure; verified that the framework's own reset-flow caller (`behaviour_utils.reset_tendermint_with_wait:1957`) reads the JSON `status` field rather than the HTTP code, so the change only tightens the contract for external probes.
3. `80bf21b8c` — Chore bump of `open-aea 2.2.5 → 2.2.6` plus sync of the updated ledger connection. Brings in the per-attempt retry cap (`MAX_RETRY_DELAY = 60s` in `ledger_dispatcher.py`), which is the upstream resolution of audit finding #13 / open-aea#891 F4.
4. `fc46e31c2` — Revert the #18 import rename in `autonomy/chain/utils.py`. Verified that `aea.helpers.http_requests.exceptions.RequestException` is aliased to the same `ConnectionError` class the pre-fix code already catches, so the rename was a no-op.

## Review iterations

`6347cf602` — round-1 review feedback applied (CI/lint blockers, retry-count regression, FSM-corruption catch, contract-change confirmation, observability log on the dockerfile reset paths).

`9d4ce6cfb` — regenerate `poetry.lock` after the open-aea bump.

`be8c31ce5`, `02076d484` — CI lint blockers: D205/D209 docstring shape and `requests.ConnectionError` resolution against the `aea.helpers.http_requests` shim.

`ad704b515` — sweep the `==2.2.5` plugin pins inside every dev package's `aea-config.yaml` / `skill.yaml` / `service.yaml` `dependencies:` block to `==2.2.6`. The original bump touched `pyproject.toml` / `tox.ini` / CID hashes but missed the per-package dep blocks, so `aea install` in the e2e job pulled `==2.2.5` plugins alongside the framework `==2.2.6` install and pip's resolver failed.

`5264789eb` — round-2 review applied. 3 bugs and 5 advisories addressed:

- `connections/ipfs/connection.py` request-path None-guard on `dialogues.update()` (sibling of the existing `_handle_done_task` guard).
- `connections/ipfs/connection.py` `_handle_done_task` now logs when dialogue recovery fails before falling through to a `None` envelope.
- `behaviour_utils.py::_check_sync` split the `(json.JSONDecodeError, KeyError, TypeError, ValueError)` catch so schema drift logs a distinct "version mismatch" warning instead of the legacy "not accepting transactions yet" debug.
- `tendermint/app.py` (localhost copy) `/params` POST catch widened to include `TypeError`.
- `autonomy/chain/config.py` `_PUBLIC_RPC_FALLBACK_WARNED: set[str]` annotation.
- `test_models.py` 5xx/600 docstring imprecision reworded.
- 12 new tests covering every behavioural branch reviewers flagged as untested.

`d0d583e63` — round-3 finding (raised independently by @OjusWiZard and @DhairyaPatel7): the round-2 None-guard called `run_async(self._handle_error, err, …)` but `_handle_error(reason, dialogue)` required both args, so the task body would `TypeError` at runtime. Widened `_handle_error` to accept `dialogue: Optional[BaseDialogue] = None` and replaced the mocked-`run_async` test with one that awaits the dispatched task end-to-end.

`6a6354f85` — round-3 follow-ups from @LOCKhart07: add `TypeError` to the `/params` GET catch (parity with the POST fix; same failure mode if `/status` returns valid JSON that isn't a dict); drop the now-stale `# pragma: nocover` from both `_check_sync` except arms.

`b94b2ed4e` — round-3 follow-up after Windows + macOS CI surfaced 3 test failures: the audit hardening had been applied only to `autonomy/deploy/generators/localhost/tendermint/app.py` (the localhost Flask app). The repo has a second copy at `deployments/Dockerfiles/tendermint/app.py` (the dockerfile-baked Flask app, which is what `tests/test_deployments/test_app.py` imports). Applied the same `/params` GET + POST catch widening to the dockerfile copy.

`498655677` — CI infra: `valory/contracts-amm:latest` started returning HTTP 404 from Docker Hub at ~16:18Z on 2026-05-15 (verified via the Docker Hub API and a local `docker pull` reproducing the CI error). The image was reachable at ~15:43Z the same day; sister images in the same workflow pull list all still resolve. Static check confirms no caller in this repo (`grep -rnE "contracts-amm|AMMNetDockerImage|amm_net_scope|UseAMM" autonomy/ packages/ tests/ scripts/ plugins/aea-test-autonomy/tests/` matches only `scripts/whitelist.py` flagging `AMMNetDockerImage` as "unused class"); runtime check confirms all 2423 tests collect cleanly with the image absent, and the docker-fixture-using chain test bucket runs green using only the locally-cached `autonolas-registries` image. Dropped both pull lines from `main_workflow.yml`.

## Test plan

Verified locally on every commit since the round-2 review:

- `tomte format-code` (black + isort) clean.
- `tox -e flake8 / mypy / pylint / darglint` clean.
- `tox -e check-hash / check-doc-hashes / check-api-docs / check-packages` clean.
- `autonomy packages lock --check` verifies the third-party ledger hash matches upstream `valory-xyz/open-aea@v2.2.6`.
- Direct pytest runs on every test added by the review-iteration commits pass.
- Local reproduction of CI's `install dependencies` step: 5/5 reachable images pull, `valory/contracts-amm` 404s as expected. Chain-fixture test bucket runs green with the image absent and `docker events --filter event=pull` captures zero pulls during the test run.

For reviewers:

- [ ] CI passes after the workflow fix lands
- [ ] Downstream sync (mech / mech-predict / trader / mech-interact) does not break on the new ledger connection hash
- [ ] The Flask `/hard_reset`, `/gentle_reset`, `/app_hash` HTTP-code flip from 200 to 500 on failure is acceptable (the framework's own caller already reads the JSON `status` field, but external monitoring may need to be checked)

Refs: #2501


packages/valory/skills/abstract_round_abci/handlers.py
```diff
@@ -426,7 +427,17 @@ def handle(self, message: Message) -> None:
...
         ).current_behaviour
-        callback(message, current_behaviour)
+        try:
+            callback(message, current_behaviour)
+        except Exception:  # pylint: disable=broad-except
+            self.context.logger.exception(
+                f"Callback for request nonce {request_nonce} raised"
+            )
+            if current_behaviour is not None:
+                try:
```
jmoreira-valory: **MEDIUM — liveness hole: `try_send(None)` bypasses the `WebSocketClientProtocol` send queue**

`try_send(None)` is not a protocol-level no-op — on CPython's `websockets` implementation it calls `connection.send(None)` which serialises `None` as the string `"None"` and enqueues it as a text frame. If the underlying writer is paused (back-pressure) or the connection is half-closed, this call can raise `ConnectionClosed` or `RuntimeError: WebSocket is not connected`, neither of which is caught here. The surrounding `try/except Exception` at the call-site swallows the error silently, so the agent's WebSocket pump stalls without any log entry or reconnect trigger.

Reproducer path: remote closes the write-half (TCP FIN) while a long round is in progress → `try_send(None)` raises `ConnectionClosed` → swallowed → no reconnect → agent hangs until Tendermint times out the ABCI connection.

**Suggested fixes (pick one):**

**A — Send an explicit WebSocket ping instead (preferred):**
```python
# Replace try_send(None) with a protocol-level ping
await websocket.ping()
```
`ping()` uses the WebSocket control-frame path, gets a `pong` response that confirms the connection is alive, and raises `ConnectionClosed` cleanly on failure so the outer reconnect logic can trigger.

**B — Guard with a connection-state check before sending:**
```python
if websocket.open:
    await websocket.ping()
```
Slightly cheaper if the common case is an already-closed socket, but `open` can race on CPython — option A is safer.

**C — Remove the keepalive call entirely and rely on `ping_interval` in the `connect()` call:**
```python
# In the connect() invocation:
async with websockets.connect(uri, ping_interval=20, ping_timeout=10) as ws:
    ...
```
The `websockets` library's built-in keepalive task sends pings on its own schedule and closes the connection (raising `ConnectionClosed`) if the peer doesn't respond within `ping_timeout`. This removes the hand-rolled keepalive entirely.

Option C is the cleanest if the `connect()` call is accessible here; option A is the minimal-diff fix if it isn't.

packages/valory/connections/ipfs/connection.py
```diff
@@ -201,10 +213,11 @@ def _handle_store_files(
             # are being uploaded.
             _, hash_, _ = self.ipfs_tool.add(path)
             self.logger.debug(f"Successfully stored files with hash: {hash_}.")
-        except (
-            ValueError,
-            IPFSError,
-        ) as e:  # pragma: no cover
+        except Exception as e:  # pylint: disable=broad-except
+            # Broaden from (ValueError, IPFSError) so network errors
+            # (requests.ConnectionError, OSError, ipfshttpclient.CommunicationError)
```
OjusWiZard: **[CI BLOCKER + Comment-rot]** flake8 fails here with `E800 Found commented out code` — the eradicate plugin reads `(requests.ConnectionError, OSError, ipfshttpclient.CommunicationError)` as a commented-out tuple.

Separately, even if you `# noqa: E800` it, enumerating concrete class names from a third-party library inside a comment is high comment-rot risk: if `ipfshttpclient` changes its exception hierarchy or is swapped, the comment silently goes stale with no compile-time signal.

**Suggested fix** — rephrase the rationale generically and drop the class enumeration:

```python
except Exception as e:  # pylint: disable=broad-except
    # Network and I/O errors from the IPFS client are not part of a
    # stable API surface; broaden the catch so they surface as ERROR
    # envelopes rather than escaping into the done-callback and
    # stranding the requesting skill.
```

This fixes E800 and removes the third-party coupling in one change.

packages/valory/connections/ipfs/connection.py
```diff
@@ -304,7 +317,11 @@ def _handle_done_task(self, task: asyncio.Future) -> None:
         :param task: the done task.
         """
         request = self.task_to_request.pop(task)
-        response_message: Optional[Message] = task.result()
+        try:
+            response_message: Optional[Message] = task.result()
+        except Exception:  # pylint: disable=broad-except
+            self.logger.exception("IPFS task failed")
+            response_message = None
```
OjusWiZard: **[CRITICAL — silent failure, 4 reviewers concur]** When `task.result()` raises, this sets `response_message = None`, then the code below falls through and enqueues `None` into `self._response_envelopes`. The receiving skill gets a `None` envelope (indistinguishable from "no envelope yet") and silently stalls — it never receives the `ERROR`-performative `IpfsMessage` it would for every other failure mode in this class.

Every other error path here routes through `self._handle_error(err, dialogue)`. This one is inconsistent.

**Fix sketch:**

```python
try:
    response_message: Optional[Message] = task.result()
except Exception as exc:  # pylint: disable=broad-except
    self.logger.exception("IPFS task failed")
    dialogue = self.task_to_dialogue.pop(task, None)  # if you track this
    if dialogue is not None:
        response_message = cast(Message, self._handle_error(str(exc), dialogue))
    else:
        response_message = None
```

If the dialogue isn't easily recoverable here, at minimum guard the enqueue: `if response_envelope is None: return` rather than `put_nowait(None)`.

packages/valory/skills/abstract_round_abci/tests/test_models.py
```diff
@@ -345,6 +350,83 @@ def test_process_response(
...
+                MagicMock(body=b'{"value": "null"}', status_code=200),
+            ),
+        ),
+    )
+    def test_process_response_returns_none_on_unparseable_body(
+        self,
+        api_specs_config: dict,
+        message: MagicMock,
+    ) -> None:
+        """`process_response` must return None on body shapes that cannot be decoded
```
OjusWiZard: **[CI BLOCKER]** flake8 `D205 1 blank line required between summary line and description`. The docstring summary and the `Covers: ...` paragraph need a blank line between them.

```python
"""`process_response` must return None on body shapes that cannot be decoded or coerced to the configured response type.

Covers: non-UTF-8 bytes (UnicodeDecodeError), non-coercible int / float
strings (ValueError from `int("N/A")` / `float("null")`).
"""
```

(Also: a single-line summary is preferred — consider collapsing to `"""Return None on bodies that cannot be decoded or coerced."""` and dropping the redundant inline comment on the next line, which currently duplicates the docstring.)

packages/valory/skills/abstract_round_abci/models.py
```diff
@@ -64,7 +64,7 @@
 )
 
 MIN_RESET_PAUSE_DURATION = 10
-NUMBER_OF_RETRIES: int = 5
+NUMBER_OF_RETRIES: int = 3
```
OjusWiZard: **[IMPORTANT]** Silently reducing `NUMBER_OF_RETRIES` from 5 → 3 in the same commit as the new 5xx short-circuit compounds the effect: under a sustained upstream 5xx flood, services exhaust their retry budget 40% faster and fall back to chain-derived randomness sooner.

Downstream consumers (mech, trader, mech-interact) have alert thresholds tuned around 5-retry tails. This change is not mentioned in the PR description's behaviour list and is not documented in a `CHANGELOG.md` or skill-level note. Either:

1. Keep `NUMBER_OF_RETRIES = 5` and let the new 5xx short-circuit do its work within the existing budget, or
2. Keep 3 but call it out in the PR description and the downstream-sync checklist.

Minor: consider `NUMBER_OF_RETRIES: Final[int] = 3` plus a module-level `assert NUMBER_OF_RETRIES >= 1` to encode the "at least one attempt" invariant in the type system.

packages/valory/skills/abstract_round_abci/models.py
```diff
@@ -711,14 +711,24 @@ def _parse_response_data(self, response_data: Any) -> Any:
...
         """Process response from api."""
-        decoded_response = response.body.decode()
+        try:
+            decoded_response = response.body.decode()
+        except (UnicodeDecodeError, AttributeError):
+            self.context.logger.error("Could not decode response body as UTF-8")
+            return None
         self.response_info.error_data = None
 
+        if response.status_code >= 500 or response.status_code == 600:
```
OjusWiZard: **[IMPORTANT — dead code]** `response.status_code >= 500 or response.status_code == 600` — the second arm is unreachable because `600 >= 500` is already true. Either drop the redundant `or response.status_code == 600`:

```python
if response.status_code >= 500:
    self.context.logger.warning(
        f"Upstream returned {response.status_code}; treating as transient"
    )
    return None
```

Or — if 600 is a meaningful sentinel (it's the `http_client/connection.py` connection-error code) and you want to signal that intent — extract a named constant:

```python
CONNECTION_ERROR_STATUS_CODE: Final[int] = 600  # set by http_client/connection.py on transport errors
```

and keep the check as `>= 500` (which still covers 600). Right now the dead `== 600` arm misleads readers into thinking 600 is somehow below 500.

Also: the return type is `Any` but this function now has four distinct `return None` paths — widen to `Optional[Any]` so callers must acknowledge the null case at type-check time.

packages/valory/skills/abstract_round_abci/handlers.py
OjusWiZard: **[IMPORTANT — silently corrupts FSM state]** When the callback raises, this catches broadly, logs, then calls `current_behaviour.try_send(None)`. Sending `None` into the generator advances it as if a response had arrived; behaviours that pattern-match the yielded value will proceed with corrupt state.

The original exception is effectively swallowed at the behaviour level — the handler returns normally and the FSM keeps running on a synthesised non-response.

**No test covers this path.** `tests/.../test_handlers.py::test_handle` only asserts the happy path.

Options:
1. Don't fabricate a response — let the behaviour time out on `wait_for_message` rather than receive a fake `None`. Just log and return.
2. If a wake-up is required, send a typed error marker that the behaviour can pattern-match on, not raw `None`.
3. At minimum, add a test that configures the callback to raise and asserts both the log and the recovery action.

packages/valory/skills/abstract_round_abci/handlers.py
```diff
@@ -656,6 +667,13 @@ def _check_registered(
         self, message: TendermintMessage, dialogue: TendermintDialogue
     ) -> bool:
         """Check if the sender is registered on-chain and if not, reply with an error"""
+        if (
```
OjusWiZard: **[ADVISORY]** Accessing `self.context.state._round_sequence` reaches into name-mangled private state of `SharedState`. The `# pylint: disable=protected-access` suppresses the linter but doesn't address the brittleness: if `SharedState` renames or moves the attribute, this guard silently raises `AttributeError` rather than degrading gracefully.

The public API already exposes a `round_sequence` property — and `SharedState` presumably knows when it's initialised. Cleaner:

```python
if not self.context.state.is_fsm_initialised:  # add this typed predicate
    ...
```

or reuse the existing public guarded `round_sequence` accessor and catch its specific exception type.

autonomy/deploy/generators/localhost/tendermint/app.py
```diff
@@ -299,7 +314,7 @@ def gentle_reset() -> Tuple[Any, int]:
             tendermint_node.start()
             return jsonify({"message": "Reset successful.", "status": True}), 200
         except Exception as e:  # pylint: disable=W0703
-            return jsonify({"message": f"Reset failed: {e}", "status": False}), 200
+            return jsonify({"message": f"Reset failed: {e}", "status": False}), 500
```
OjusWiZard: **[CRITICAL — contract change vs. framework callers]** The PR description claims the framework's own caller (`reset_tendermint_with_wait`) reads the JSON `status` field rather than the HTTP code, but two things break that assumption:

1. `behaviour_utils.py:_gentle_reset` (around line 2197) calls `_do_request` and **discards the return value entirely**. If the underlying http connection short-circuits on non-200 (e.g. `_check_http_return_code_200` at behaviour_utils.py:869 used for tx submission), a 500 here interferes silently.
2. `reset_tendermint_with_wait` parses `result.body` without checking `result.status_code`. Status-code-aware middleware between Flask and the skill (proxies, retry connectors) may now intercept the response before the skill ever sees the JSON.

Applies to all three endpoints: `/gentle_reset` (line 317), `/app_hash` (line 333), `/hard_reset` (line 358).

This is the same finding as the PR-description checkbox "the Flask `/hard_reset`, `/gentle_reset`, `/app_hash` HTTP-code flip from 200 to 500 on failure is acceptable". Recommend either:
- Revert to HTTP 200 + `status: False` (the existing contract — `status` field is already the source of truth), or
- Audit and update every caller (`_gentle_reset`, `reset_tendermint_with_wait`, any external monitor) to read `status_code` and treat 5xx as transient before this lands.

autonomy/deploy/generators/localhost/tendermint/app.py
```diff
@@ -249,14 +249,24 @@ def get_params() -> Dict:
             )
             priv_key_data = json.loads(priv_key_file.read_text(encoding=ENCODING))
             del priv_key_data["priv_key"]
-            status = requests.get(TM_STATUS_ENDPOINT, timeout=30).json()
+            non_routable, loopback = "0.0.0.0", "127.0.0.1"  # nosec
```
OjusWiZard: **[ADVISORY — duplication + no log on failure]**

1. **Duplicated URL rewrite.** The two-step `.replace('tcp', 'http').replace('0.0.0.0', '127.0.0.1')` (plus the `# nosec`) is copy-pasted verbatim at line 323 in `app_hash`. Extract a helper:

```python
def _tendermint_http_url(path: str) -> str:
    non_routable, loopback = "0.0.0.0", "127.0.0.1"  # nosec
    base = tendermint_params.rpc_laddr.replace("tcp", "http").replace(non_routable, loopback)
    return f"{base}{path}"
```

2. **No log on failure path.** The widened `except (FileNotFoundError, json.JSONDecodeError, KeyError, requests.RequestException)` returns `{status: False, error: traceback.format_exc()}` but never calls `logging.error`. Operators watching stdout/stderr see nothing; only consumers who deserialise the JSON do. Same applies to `update_params` (line 298) which silently swallows the new `KeyError`. Add a `logging.exception(...)` before each return.

autonomy/chain/config.py
```diff
@@ -114,10 +115,19 @@ def id(self) -> Optional[int]:
...
-            key=self.rpc_env_name,
-            default=CHAIN_ID_TO_DEFAULT_PUBLIC_RPC.get(self.id) if self.id else None,
-        )
+        configured = os.getenv(self.rpc_env_name)
+        if configured is not None:
+            return configured
+        fallback = CHAIN_ID_TO_DEFAULT_PUBLIC_RPC.get(self.id) if self.id else None
+        if fallback is not None:
+            logging.getLogger(__name__).warning(
+                "%s not set; using default public RPC %s. "
```
OjusWiZard: **[ADVISORY]** `ChainType.rpc` is a `@property`, not a cached value — any code that accesses `.rpc` in a loop (e.g. iterating chains during service deployment, or polling) emits a duplicate warning on every access.

Also, the message inlines `self.rpc_env_name` three times (`%s not set; using default ... Set %s ...`). Two of those slots have the same value.

**Fix:** either log once at module / service startup, or cache. Simplest cache:

```python
@property
def rpc(self) -> Optional[str]:
    """RPC String"""
    configured = os.getenv(self.rpc_env_name)
    if configured is not None:
        return configured
    fallback = CHAIN_ID_TO_DEFAULT_PUBLIC_RPC.get(self.id) if self.id else None
    if fallback is not None and not getattr(self, "_rpc_warned", False):
        logging.getLogger(__name__).warning(
            "%s not set; using default public RPC %s. "
            "Set it to use a private endpoint.",
            self.rpc_env_name, fallback,
        )
        # ChainType is an Enum so this needs to live on a class-level dict, e.g. _warned: set
    return fallback
```

Note that `ChainType` is an Enum and members are typically frozen/cached, so naive instance-attr caching won't work — use a module-level `_warned: set[str]` keyed on `rpc_env_name`.

packages/valory/skills/abstract_round_abci/behaviour_utils.py
```diff
@@ -1998,14 +2012,14 @@ def reset_tendermint_with_wait(  # pylint: disable=too-many-locals, too-many-sta
         status = yield from self._get_status()
         try:
             json_body = json.loads(status.body.decode())
-        except json.JSONDecodeError:
+            remote_height = int(json_body["result"]["sync_info"]["latest_block_height"])
+        except (json.JSONDecodeError, KeyError, TypeError, ValueError):
             self.context.logger.error(
```
OjusWiZard, line 2019: **[ADVISORY — misleading error message]** Collapsing `json.loads` and `int(json_body[...])` into one `try` block is a real improvement, but the existing error message — `"Tendermint not accepting transactions yet, trying again!"` — was written for the `JSONDecodeError` case ("Tendermint hasn't started"). With the new exception types (`KeyError`, `TypeError`, `ValueError`) it also fires for a structurally-malformed response that will **never** self-heal — e.g. a Tendermint version that renamed `sync_info` — and the loop will retry forever printing the misleading message at `ERROR` severity.

Consider differentiating:

```python
except json.JSONDecodeError:
    self.context.logger.error("Tendermint not accepting transactions yet, trying again!")
    yield from self.sleep(self.params.sleep_time)
    return False
except (KeyError, TypeError, ValueError) as e:
    self.context.logger.error(
        "Unexpected Tendermint status response shape (%s); will retry but this "
        "may indicate a Tendermint version mismatch", e,
    )
    yield from self.sleep(self.params.sleep_time)
    return False
```

OjusWiZard (commented):
Multi-perspective review (code-reviewer, silent-failure-hunter, type-design-analyzer, code-simplifier, pr-test-analyzer, comment-analyzer). 3 CI failures, 2 CRITICAL, 4 IMPORTANT, 6 ADVISORY. Findings posted inline; details below.

**CI is failing** — needs to be addressed regardless of any review feedback:
1. `connection.py:218` flake8 **E800** (commented-out code) — inline comment below proposes a fix that also addresses the rot risk separately flagged by comment-analyzer.
2. `test_models.py:399` flake8 **D205** — missing blank line between docstring summary and description.
3. `tox -e check-api-docs` fails: `docs/api/skills/abstract_round_abci/tests/test_models.md` is out of sync with `test_models.py`. Run `tox -e generate-api-documentation` and commit the result.

**Test coverage gaps** (no new tests for any of these new branches):
- Tendermint Flask `/hard_reset`, `/gentle_reset`, `/app_hash` 500-path: `tests/test_deployments/test_app.py` only asserts the 200 success path.
- `handlers.py` callback-exception guard (line 432–440): no test makes the callback raise.
- `handlers.py` `_check_registered` with `_round_sequence is None` (line 667–672): the existing `setup_method` always supplies a full MagicMock context.
- `autonomy/chain/subgraph/client.py` 200-with-HTML-body `JSONDecodeError` path: covered for 4xx but not for malformed 200.
- `connection.py` `_handle_done_task` exception path and the new empty-listdir guard.

**Per the user's standing instruction** — I posted this review only because explicit consent was given in the current turn; no further comments will be added without another approval.

packages/valory/skills/abstract_round_abci/tests/test_models.py
DIvyaNautiyal07: Fixed in 6347cf602. Collapsed to single-line summary, dropped the redundant inline comment too.

packages/valory/connections/ipfs/connection.py
DIvyaNautiyal07: Fixed in 6347cf602. The exception path now recovers the dialogue via `self.dialogues.update(request.message)` and builds an ERROR-performative `IpfsMessage` through `_handle_error(...)`, consistent with every other failure path in the class. The requesting skill receives a typed ERROR envelope instead of bare `None`.

packages/valory/skills/abstract_round_abci/handlers.py
DIvyaNautiyal07: Thanks for the read. One clarification on context: the `try_send` here is a method on `BaseBehaviour` (signals into the behaviour's asyncio generator via `__notified`), not on a `WebSocketClientProtocol` — so the reproducer path through `websockets.send(None)` doesn't apply. But the substantive concern (fabricating a non-response into the FSM) is real and matches @OjusWiZard's IMPORTANT finding on the same line. Dropped the `try_send(None)` line in 6347cf602; the try/except is log-only now and the behaviour recovers via the existing round timeout.

packages/valory/connections/ipfs/connection.py
DIvyaNautiyal07: Fixed in 6347cf602. Dropped the multi-line inline comment that enumerated third-party class names so eradicate stops reading it as commented-out code, and the rationale isn't pinned to a specific exception hierarchy anymore.

packages/valory/skills/abstract_round_abci/models.py
DIvyaNautiyal07: Reverted to 5 in 6347cf602. The default change was operational tuning I didn't flag in the PR description; the new 5xx short-circuit does its work within the existing 5-retry budget, and downstream alert thresholds tuned to that tail keep working unchanged.

packages/valory/skills/abstract_round_abci/handlers.py
DIvyaNautiyal07: Agreed. Dropped the `try_send(None)` fabrication in 6347cf602; the try/except is log-only now. Behaviour recovers via the existing round timeout rather than resuming with a synthesised non-response. Same change addresses @jmoreira-valory's MEDIUM on the same line.

packages/valory/skills/abstract_round_abci/models.py
DIvyaNautiyal07: Fixed in 6347cf602. The check is just `response.status_code >= 500` now; dropped the redundant `or == 600` arm.

autonomy/deploy/generators/localhost/tendermint/app.py
DIvyaNautiyal07: Reverted in 6347cf602. All three failure branches (`/gentle_reset`, `/app_hash`, `/hard_reset`) go back to HTTP 200 with `status: False` in the JSON body, preserving the existing contract — the JSON `status` field stays the source of truth. Added `logging.exception` on each failure branch so the traceback lands in agent logs too, not only in the response body.

autonomy/deploy/generators/localhost/tendermint/app.py
DIvyaNautiyal07: Both done in 6347cf602. Extracted `_tendermint_http_url(path)` inside `create_app`; both `/params` GET and `/app_hash` use it now. Added `logging.exception` on `/params` GET and POST failure branches so the swallowed `KeyError` / other failures are visible to operators.

packages/valory/skills/abstract_round_abci/handlers.py
DIvyaNautiyal07: Fixed in 6347cf602. Switched to `try/except ValueError` on the public `acn_container()` path — `SharedState.synchronized_data` already raises `ValueError` when `_round_sequence` is None, so the guard no longer touches the private attribute. If a typed `is_fsm_initialised` predicate later lands on `SharedState` we can swap to it without changing semantics.

packages/valory/skills/abstract_round_abci/behaviour_utils.py
DIvyaNautiyal07: Fixed in 6347cf602 along the lines you suggested. Split the catch: `JSONDecodeError` keeps the original "not accepting transactions yet, trying again!" message; `(KeyError, TypeError, ValueError)` gets a distinct "Unexpected Tendermint status response shape (%s); retrying. This may indicate a Tendermint version mismatch." log.

autonomy/chain/config.py
DIvyaNautiyal07: Fixed in 6347cf602. Added a module-level `_PUBLIC_RPC_FALLBACK_WARNED: set` keyed on `rpc_env_name`. The warning fires once per process per env var now, so iterating chains during service deployment or polling doesn't repeat the message. Also tightened the format string so the env var name appears once, not three times.

jmoreira-valory (dismissed):

packages/valory/connections/ipfs/connection.py
```diff
@@ -151,10 +161,12 @@ def _handle_envelope(self, envelope: Envelope) -> Task:
...
             err = f"Performative `{performative.value}` is not supported."
             self.logger.error(err)
-            task = self.run_async(self._handle_error, err)
+            task = self.run_async(
+                self._handle_error, err, timeout=self._request_timeout
+            )
             return task
         dialogue = self.dialogues.update(message)
-        task = self.run_async(handler, message, dialogue)
+        task = self.run_async(handler, message, dialogue, timeout=self._request_timeout)
```
OjusWiZard, line 175: **[HIGH — sibling of the `_handle_done_task` bug round 1 caught]** `IpfsDialogues.update()` returns `Optional[Dialogue]`. The post-task path (`_handle_done_task`, line 325) was fixed to guard with `if dialogue is not None`, but **this sibling path is still unguarded**: if `dialogues.update(message)` returns `None` (duplicate or invalid message), `dialogue` is passed unchecked into the handler, which then calls `dialogue.reply(...)` and raises `AttributeError: 'NoneType' object has no attribute 'reply'` at runtime, breaking the FSM with no recovery.

Mirror the `_handle_done_task` fix:

```python
dialogue = self.dialogues.update(message)
if dialogue is None:
    err = f"Could not update dialogue for message {message}"
    self.logger.error(err)
    return self.run_async(self._handle_error, err, timeout=self._request_timeout)
task = self.run_async(handler, message, dialogue, timeout=self._request_timeout)
```

packages/valory/connections/ipfs/connection.py
```diff
@@ -304,7 +313,16 @@ def _handle_done_task(self, task: asyncio.Future) -> None:
...
+        try:
+            response_message: Optional[Message] = task.result()
+        except Exception as exc:  # pylint: disable=broad-except  # noqa: BLE001
+            self.logger.exception("IPFS task failed")
+            request_message = cast(IpfsMessage, request.message)
+            dialogue = self.dialogues.update(request_message)
+            if dialogue is not None:
+                response_message = cast(Message, self._handle_error(str(exc), dialogue))
+            else:
+                response_message = None
```
OjusWiZard, line 354: **[ADVISORY — observability]** When the exception path lands here (task failed AND `dialogues.update` returned `None`), `response_message = None` falls through to `put_nowait(None)` on line 337 with no log. The original `exc` was logged on line 317, but the secondary "dialogue could not be reconstructed, dropping response" condition is silent. Add:

```python
else:
    self.logger.error(
        "Could not recover dialogue for failed IPFS task; dropping response. "
        "The requesting skill will time out."
    )
    response_message = None
```

Without this, a timed-out IPFS task that lost its dialogue silently puts `None` on the queue and the only diagnostic is the original `IPFS task failed` line.

packages/valory/skills/abstract_round_abci/behaviour_utils.py
```diff
@@ -763,7 +763,12 @@ def _check_sync(
...
                 yield from self.sleep(self.context.params.tendermint_check_sleep_delay)
-            except (json.JSONDecodeError, KeyError):  # pragma: nocover
+            except (
+                json.JSONDecodeError,
+                KeyError,
+                TypeError,
+                ValueError,
+            ):  # pragma: nocover
                 self.context.logger.debug(
                     "Tendermint not accepting transactions yet, trying again!"
```
OjusWiZard, line 768: **[MEDIUM — same misleading-message issue I flagged at line 2017, but in a *different* function]** Round 1 fixed `reset_tendermint_with_wait` by splitting `JSONDecodeError` (transient) from `(KeyError, TypeError, ValueError)` (structural / version mismatch). The exact same pattern in `_check_sync` was widened to include `TypeError`/`ValueError` here but **still logs the single "Tendermint not accepting transactions yet" message** for all four exception types.

A `KeyError` on `result/sync_info/latest_block_height` or a non-numeric block height (`ValueError` from `int(...)`) means the response shape is wrong — Tendermint is accepting transactions just fine, it's the response schema that changed. The current log misdirects operators.

Apply the same split as the fix at line 2014:

```python
except json.JSONDecodeError:
    self.context.logger.debug(
        "Tendermint not accepting transactions yet, trying again!"
    )
    yield from self.sleep(self.context.params.tendermint_check_sleep_delay)
except (KeyError, TypeError, ValueError) as exc:
    self.context.logger.warning(
        "Unexpected Tendermint status response shape (%s); retrying. "
        "This may indicate a Tendermint version mismatch.", exc,
    )
    yield from self.sleep(self.context.params.tendermint_check_sleep_delay)
```

autonomy/deploy/generators/localhost/tendermint/app.py
```diff
@@ -288,7 +310,15 @@ def update_params() -> Dict:
             )
 
             return {"status": True, "error": None}
-        except (FileNotFoundError, json.JSONDecodeError, PermissionError):
+        except (
+            FileNotFoundError,
+            json.JSONDecodeError,
+            PermissionError,
+            KeyError,
```
OjusWiZard, line 318: **[MEDIUM — gap in widened catch]** The except tuple covers `FileNotFoundError`, `JSONDecodeError`, `PermissionError`, `KeyError`. It still doesn't catch **`TypeError`**, which is raised by `update_genesis_config(data=data)` / `update_peers(validators=data["validators"], ...)` / `update_external_address(external_address=data["external_address"], ...)` when a request body has malformed-but-JSON-valid values (e.g. `{"validators": null}`, `{"external_address": 42}`).

A `TypeError` propagates unhandled, Flask returns HTTP 500, and the caller (`behaviour_utils.py`) is built around the `{"status": False, "error": ...}` contract — `.json()` on a Flask 500 HTML response then crashes the behaviour.

Add `TypeError` to the tuple:

```python
except (
    FileNotFoundError,
    json.JSONDecodeError,
    PermissionError,
    KeyError,
    TypeError,
):
```

The sibling `/params` GET (line 273) catches `requests.RequestException` for the same reason — keep the two parallel.

packages/valory/skills/abstract_round_abci/handlers.py
```diff
@@ -656,7 +661,13 @@ def _check_registered(
         self, message: TendermintMessage, dialogue: TendermintDialogue
     ) -> bool:
         """Check if the sender is registered on-chain and if not, reply with an error"""
-        others_addresses = self.context.state.acn_container()
+        try:
```
OjusWiZard, line 664: **[ADVISORY — implicit exception contract]** The `try/except ValueError` is cleaner than the original private-attribute access (good fix), but the `ValueError` it catches isn't declared anywhere on `acn_container`. It originates in `SharedState.round_sequence` at `models.py:520` (`raise ValueError("round sequence not available")`) and propagates through *two* layers (`acn_container → synchronized_data → round_sequence`) without being mentioned in any signature or docstring.

Readers cannot tell why `acn_container` raises `ValueError`, and the catch is over-broad — a future contributor adding an unrelated `ValueError` to `acn_container` will silently fold into this "FSM not yet initialised" message.

Suggested:

```python
# in models.py
class RoundSequenceNotInitialisedError(ValueError):
    """Raised when SharedState.round_sequence is accessed before FSM setup."""

# round_sequence property:
if self._round_sequence is None:
    raise RoundSequenceNotInitialisedError("round sequence not available")

# here in handlers.py:
except RoundSequenceNotInitialisedError:
    self._reply_with_tendermint_error(message, dialogue, "FSM not yet initialised")
    return False
```

Makes the invariant explicit, keeps backwards compatibility (subclass of `ValueError`), and narrows the catch.

(Same comment applies as a follow-up for any other call site that currently catches `ValueError` for this exact purpose.)

OjusWiZard (commented):
Second-pass review (round 2). Round 1's 11 findings are all addressed and threads resolved. This pass found:

- **3 new bugs** (1 HIGH, 2 MEDIUM) — sibling of the `_handle_done_task` fix, the same misleading-message issue in a different function, and a missing `TypeError` in the new widened catch.
- **6 test gaps** — none of the new error branches landed in the fix commits got tests.
- **4 advisory** type/log/doc improvements.

No `code-simplifier` findings this round.

### Test gaps (not inline — these test files aren't in the diff so I can't anchor a comment to a specific line):

| Test file | Missing |
|---|---|
| `tests/test_deployments/test_app.py::test_get_params` | `requests.ConnectionError` failure path (the goal of fix commit `02076d484`) |
| `packages/valory/skills/abstract_round_abci/tests/test_handlers.py::TestAbstractResponseHandler::test_handle` | Callback raising — the new `try/except Exception` logger path |
| `…/test_handlers.py::TestTendermintHandler` | `acn_container()` raising `ValueError` — the new `_check_registered` guard |
| `tests/test_autonomy/test_chain/test_subgraph.py` | 200-status response with `JSONDecodeError` body (only 4xx is tested) |
| `packages/valory/connections/ipfs/tests/test_ipfs_connection.py::test_handle_done_task` | `task.result()` raising; the new `dialogue is None` sub-branch; the new empty-listdir guard |
| `…/tests/test_behaviours_utils.py` | `_sync_height` widened catch (`KeyError`/`TypeError`/`ValueError`) |

The `_PUBLIC_RPC_FALLBACK_WARNED` warn-once behaviour also has no test (only the return value is asserted in `test_chain_type_rpc_and_env_names`).

autonomy/chain/config.py
```diff
@@ -55,6 +56,8 @@
 CUSTOM_CHAIN_RPC = "CUSTOM_CHAIN_RPC"
 ETHEREUM_CHAIN_RPC = "ETHEREUM_CHAIN_RPC"
 
+_PUBLIC_RPC_FALLBACK_WARNED: set = set()
```
OjusWiZard: **[ADVISORY — type narrowing]** `set` is a bare type with no element parameter. Values added at line 128 are `self.rpc_env_name` strings, so:

```python
_PUBLIC_RPC_FALLBACK_WARNED: set[str] = set()
```

Readable as "strings only", mypy infers element type for membership checks, and future contributors don't need to scroll to the `.add()` call to see what goes in.

packages/valory/skills/abstract_round_abci/tests/test_models.py
```diff
@@ -345,6 +350,78 @@ def test_process_response(
...
+
+    @pytest.mark.parametrize(
+        "status_code",
+        (500, 502, 503, 504, 600),
+    )
+    def test_process_response_returns_none_on_server_error_status(
+        self,
+        status_code: int,
+    ) -> None:
+        """5xx and the connection-error sentinel (600) short-circuit JSON parse."""
```
OjusWiZard: **[ADVISORY — docstring imprecision]** `"""5xx and the connection-error sentinel (600) short-circuit JSON parse."""` is slightly misleading. The production guard at `models.py:726` is `if response.status_code >= 500:` — there's **no special-case** handling of 600. The value 600 is just `>= 500` and is caught by the same branch.

600 is `HTTPClientAsyncChannel.DEFAULT_EXCEPTION_CODE` defined in `packages/valory/connections/http_client/connection.py:112-114`, used as a synthetic status code when the http_client connection itself fails (no upstream response). Calling it a "sentinel" implies the framework treats it specially here; it doesn't.

Suggested rewrite:

```python
"""5xx (including the http_client transport-failure synthetic 600) short-circuit JSON parse."""
```

And the inline comment on line 421 — `# Body would parse as valid JSON for a 2xx, but at 5xx/600 we don't try.` — can drop `/600` since `5xx` covers it semantically.

packages/valory/connections/ipfs/connection.py
DIvyaNautiyal07: Fixed in 5264789eb. Mirrored the `_handle_done_task` guard: if `dialogues.update(message)` returns `None`, the request path now logs and routes through `self._handle_error` instead of calling the handler with `dialogue=None`. Covered by `test_handle_envelope_returns_error_task_when_dialogue_update_is_none` in `tests/test_connection.py`.

packages/valory/connections/ipfs/connection.py
DIvyaNautiyal07: Fixed in 5264789eb. Added an explicit `self.logger.error("Could not recover dialogue for failed IPFS task; dropping response. The requesting skill will time out.")` in the dialogue-recovery-failed branch before `response_message = None`. Covered by `test_handle_done_task_drops_response_when_dialogue_cannot_be_rebuilt` (asserts the log fires and a `None` envelope still lands on the queue).

packages/valory/skills/abstract_round_abci/behaviour_utils.py
DIvyaNautiyal07: Fixed in 5264789eb. Split the catch as suggested: `JSONDecodeError` keeps the legacy `"Tendermint not accepting transactions yet, trying again!"` debug log, and `(KeyError, TypeError, ValueError)` gets a distinct `"Unexpected Tendermint status response shape (%s); retrying. This may indicate a Tendermint version mismatch."` warning. Both branches are covered: `test_check_sync_logs_version_mismatch_for_schema_drift` and `test_check_sync_keeps_legacy_message_for_json_decode_error` in `tests/test_behaviours_utils.py`.

autonomy/deploy/generators/localhost/tendermint/app.py
DIvyaNautiyal07: Fixed in 5264789eb. Added `TypeError` to the except tuple — the sibling `/params` GET catch (line 273) was already broad enough; this brings the POST handler into parity. Covered by `test_update_params_handles_typeerror_on_malformed_body` in `tests/test_deployments/test_app.py` (POSTs `{validators: null}` and asserts the `{status: False, error: ...}` contract holds instead of falling through to a Flask 500).

packages/valory/skills/abstract_round_abci/handlers.py
DIvyaNautiyal07: Deferring this one as a separate change. The `RoundSequenceNotInitialisedError(ValueError)` subclass is a sensible cleanup but introduces a new exported exception type that other callers would need to learn about; want to land it in its own PR with the call-site sweep in scope. For now I've added `test_handle_replies_when_acn_container_raises_value_error` (in `test_handlers.py`) so the catch's behaviour is locked even if the surrounding implementation changes — happy to file a follow-up issue. Commit 5264789eb.

autonomy/chain/config.py
DIvyaNautiyal07: Fixed in 5264789eb. `_PUBLIC_RPC_FALLBACK_WARNED: set[str] = set()` applied. Also added `test_public_rpc_fallback_warns_at_most_once_per_env_name` in `tests/test_autonomy/test_chain/test_config.py` to lock the warn-once invariant (the production property reads were never exercised by a test before).

packages/valory/skills/abstract_round_abci/tests/test_models.py
DIvyaNautiyal07: Fixed in 5264789eb. Docstring now reads `"""5xx (including the http_client transport-failure synthetic 600) short-circuit JSON parse."""` and the inline comment dropped the `/600` since `5xx` already covers it. Thanks for the precise pointer to `HTTPClientAsyncChannel.DEFAULT_EXCEPTION_CODE` — saved a chase.

packages/valory/connections/ipfs/connection.py
```diff
@@ -151,10 +161,18 @@ def _handle_envelope(self, envelope: Envelope) -> Task:
...
+            )
             return task
         dialogue = self.dialogues.update(message)
-        task = self.run_async(handler, message, dialogue)
+        if dialogue is None:
+            err = f"Could not update dialogue for message {message}"
+            self.logger.error(err)
+            return self.run_async(
+                self._handle_error, err, timeout=self._request_timeout
+            )
```
OjusWiZard, line 174: **[HIGH — the None-guard fix replaces a crash with a different crash]** `_handle_error` is declared `def _handle_error(self, reason: str, dialogue: BaseDialogue) -> IpfsMessage` (line 304) — `dialogue` is **required**. `run_async(func, *args)` invokes `func(*args)` in the executor, so this call schedules `self._handle_error(err)` — missing the second positional argument. At runtime the background task raises `TypeError: _handle_error() missing 1 required positional argument: 'dialogue'`.

The original bug (None dialogue → `AttributeError` on `dialogue.reply()`) is averted, but the new path crashes in a different way that the test can't see because it mocks `run_async`:

```python
# test_connection.py:236-237
mock.patch.object(self.connection, "run_async", return_value=sentinel_task)
# ...
assert called_handler.__func__ is type(self.connection)._handle_error
```

The assertion checks *which function* was passed, never invokes it. A real test should let `run_async` run (or call `_handle_error(err)` directly) and assert it doesn't raise.

**Note:** the pre-existing performative-not-supported path at lines 163–165 has the same bug — `self.run_async(self._handle_error, err, timeout=...)`. It was latent because unknown-performative envelopes are rare, but the new None-guard amplifies the same flaw in a much more reachable code path.

**Fix options:**

1. **Widen `_handle_error` to handle `dialogue is None`.** Most idiomatic; fixes both call sites at once:
   ```python
   def _handle_error(self, reason: str, dialogue: Optional[BaseDialogue] = None) -> Optional[IpfsMessage]:
       if dialogue is None:
           self.logger.error(
               "Dropping error response — no dialogue available. Reason: %s", reason,
           )
           return None
       ...existing body...
   ```
   Then `_handle_done_task` and `_handle_envelope` both pass whatever dialogue they have (possibly `None`), and the queue receives `None` which `put_nowait(None)` already handles (still suboptimal but consistent with the rest of the file).

2. **Drop the run_async indirection on the error paths.** Since `_handle_error` with no dialogue can't produce a reply anyway, both call sites could log and return a completed task that puts no envelope on the queue — keeping behaviour symmetric with `_handle_done_task`'s `dialogue is None` branch.

Either way, please add a test that actually exercises `_handle_error` end-to-end on the None path.

OjusWiZard (commented):
Round 3 — verifying the round-2 fix commit `5264789eb`. 6 of 7 round-2 threads addressed cleanly and resolved. 1 advisory thread left open (RoundSequenceNotInitialisedError — author's call to keep the broader `ValueError` catch).

Found **1 new HIGH bug** the fix introduced.

_The 6 fixes that landed cleanly:_ `_check_sync` split catch, `update_params` TypeError, `set[str]` annotation, test_models.py docstring rewording, the new logger on the `dialogue is None` branch of `_handle_done_task`, and the comprehensive new test files in `test_app.py` / `test_handlers.py` / `test_subgraph.py` / `test_connection.py` / `test_behaviours_utils.py` / `test_config.py`. Those all look correct and falsifiable.

autonomy/deploy/generators/localhost/tendermint/app.py
```diff
@@ -249,14 +263,22 @@ def get_params() -> Dict:
...
                 "params": priv_key_data,
                 "status": True,
                 "error": None,
             }
-        except (FileNotFoundError, json.JSONDecodeError):
+        except (
+            FileNotFoundError,
+            json.JSONDecodeError,
+            KeyError,
+            requests.ConnectionError,
```
LOCKhart07, line 278: **[MEDIUM — sibling gap to the round-2 `/params` POST fix]** The POST handler's except tuple was widened to include `TypeError` (per the round-2 fix) so a JSON-valid-but-wrong-typed body returns the `{status: False, error: ...}` contract instead of a Flask 500. This GET path has the identical exposure but still omits `TypeError`: if Tendermint `/status` — or a proxy / maintenance page in front of it — returns valid JSON that isn't an object (`null`, `"maintenance"`, a number, a list), then `status["result"]` raises `TypeError`, which escapes this tuple to a Flask 500 HTML body, and the framework caller's `.json()` parse then crashes — exactly the failure mode the POST fix closed. Add `TypeError` here too to keep the two paths in parity:

```python
except (
    FileNotFoundError,
    json.JSONDecodeError,
    KeyError,
    TypeError,
    requests.ConnectionError,
):
```

packages/valory/skills/abstract_round_abci/behaviour_utils.py
```diff
@@ -763,11 +763,18 @@ def _check_sync(
                     self.round_sequence.set_block_stall_deadline()
                     return
                 yield from self.sleep(self.context.params.tendermint_check_sleep_delay)
-            except (json.JSONDecodeError, KeyError):  # pragma: nocover
+            except json.JSONDecodeError:  # pragma: nocover
                 self.context.logger.debug(
                     "Tendermint not accepting transactions yet, trying again!"
                 )
                 yield from self.sleep(self.context.params.tendermint_check_sleep_delay)
+            except (KeyError, TypeError, ValueError) as exc:  # pragma: nocover
```
LOCKhart07: **[Non-blocking — stale `# pragma: nocover` defeats the coverage gate on newly-tested branches]** This PR adds `test_check_sync_logs_version_mismatch_for_schema_drift` and `test_check_sync_keeps_legacy_message_for_json_decode_error`, which now exercise *both* `except` arms in `_check_sync` — the `JSONDecodeError` arm at line 766 and this `(KeyError, TypeError, ValueError)` arm. But both retain `# pragma: nocover`, so coverage.py excludes them. Net effect: the tests assert the log text, yet a future regression that drops or alters either branch won't be caught by the coverage gate (the lines are excluded from measurement). Given this repo's strict meaningful-coverage policy, drop `# pragma: nocover` from both `except` lines (766 and 771) now that they're genuinely covered — otherwise the new tests are misleading about what's actually enforced.

packages/valory/connections/ipfs/connection.py
```diff
@@ -151,10 +161,18 @@ def _handle_envelope(self, envelope: Envelope) -> Task:
...
+        task = self.run_async(handler, message, dialogue, timeout=self._request_timeout)
```
DhairyaPatel7, line 175: **[HIGH — `run_async(self._handle_error, err)` lands a `TypeError`, not the promised error envelope]**

`_handle_error(reason: str, dialogue: BaseDialogue)` requires *both* arguments (zero defaults). The new guard calls `self.run_async(self._handle_error, err, timeout=...)`, which forwards only `err` as a positional arg to the executor. When the task body actually runs, it raises `TypeError: _handle_error() missing 1 required positional argument: 'dialogue'`.

The TypeError flows back through `task.result()` in `_handle_done_task` (around l. 323), gets caught by the broad `except Exception`, falls into the recovery branch which calls `self.dialogues.update(request_message)` again — same input → same `None` — then lands at `response_message = None` and `put_nowait(None)`. The requesting skill silently times out: the exact failure mode this guard was meant to prevent.

Why CI is green: `test_handle_envelope_returns_error_task_when_dialogue_update_is_none` mocks `run_async`, so the executor never runs the broken call. The TypeError only surfaces in real I/O.

Same anti-pattern pre-existed on `main` at the unsupported-performative branch (l. 161-166) — pre-existing latent bug, rarely hit. This change widens the blast radius to any duplicate or malformed dialogue update.

**Fix options**

A — Make `_handle_error` tolerate `dialogue=None` (cleanest; also fixes the latent unsupported-performative path):

```python
def _handle_error(
    self, reason: str, dialogue: Optional[BaseDialogue] = None
) -> Optional[IpfsMessage]:
    """Handler for error messages."""
    if dialogue is None:
        self.logger.error("IPFS error with no recoverable dialogue: %s", reason)
        return None
    message = cast(
        IpfsMessage,
        dialogue.reply(
            performative=IpfsMessage.Performative.ERROR,
            reason=reason,
        ),
    )
    return message
```

B — Build a synthetic error envelope inline at the call site instead of going through `_handle_error` (minimal-diff, but leaves the latent unsupported-performative bug in place).

C — At minimum, replace the mock in `test_handle_envelope_returns_error_task_when_dialogue_update_is_none` with a real-`run_async` execution and assert that the resulting `_response_envelopes` queue receives an ERROR-performative envelope rather than `None`. The current test passes today *because* it doesn't exercise the bug.

DhairyaPatel7 (commented):
Independent read of the current state. Round 1 & 2 fixes look clean. **1 HIGH** — the round-2 dialogue-None guard introduces a latent `TypeError` that defeats its own purpose; details inline. Verified the `requests.ConnectionError` switch and the open-aea plugin-pin sweep are correct.

packages/valory/connections/ipfs/connection.py
DIvyaNautiyal07: Fixed in d0d583e63. Widened `_handle_error` per your suggestion (option 1): signature is now `def _handle_error(self, reason: str, dialogue: Optional[BaseDialogue] = None) -> Optional[IpfsMessage]`, returning `None` after logging when dialogue is None. Both call sites in `_handle_envelope` (the new None-guard at line 172 AND the pre-existing performative-not-supported path at line 161) now pass `dialogue=None` explicitly, so the task body completes cleanly. Internal call sites inside `_handle_store_files` / `_handle_get_files` cast back to `IpfsMessage` because those paths verifiably have a real dialogue.

The previous test (mocked `run_async`, never invoked the task body) was replaced with `test_handle_envelope_dispatches_executable_error_task_when_dialogue_update_is_none` which actually awaits the dispatched task end-to-end and asserts both the `"Could not update dialogue"` scheduling log AND `_handle_error`'s `"Dropping error response, no dialogue available"` log fire — proving the task body ran. Plus a direct unit test for `_handle_error(reason, None)`.

Good catch — the mocked-`run_async` test was a hole.

autonomy/deploy/generators/localhost/tendermint/app.py
DIvyaNautiyal07: Fixed in 6a6354f85. Added `TypeError` to the `/params` GET except tuple — keeps it in parity with the round-2 POST fix. New test `test_get_params_handles_typeerror_on_non_dict_status_body` (in `tests/test_deployments/test_app.py`) stubs `app.requests.get` to return a response whose `.json()` is `None`, then asserts the GET returns `{status: False, error: ...}` instead of crashing on `status["result"]`. Good catch on the symmetry.

packages/valory/skills/abstract_round_abci/behaviour_utils.py
DIvyaNautiyal07: Dropped `# pragma: nocover` from both `_check_sync` except arms in 6a6354f85. You're right that the pragmas would have made the new tests misleading — the assertions would pass but coverage.py would still exclude the lines, so a future regression dropping either log wouldn't be caught by the coverage gate. Re-ran `test_check_sync_logs_version_mismatch_for_schema_drift` + `test_check_sync_keeps_legacy_message_for_json_decode_error` after pragma removal, both still pass.

packages/valory/connections/ipfs/connection.py
DIvyaNautiyal07: Fixed in d0d583e63 (independently raised by @OjusWiZard, both reviewers converged on Option A). `_handle_error` now accepts `dialogue: Optional[BaseDialogue] = None` and returns `Optional[IpfsMessage]`; when dialogue is None it logs `"Dropping error response, no dialogue available. Reason: %s"` and returns None. Both call sites in `_handle_envelope` (the new None-guard at line 172 AND the pre-existing performative-not-supported path at line 161 that you correctly flagged as the latent sibling) now pass `dialogue=None` explicitly, so the task body completes cleanly instead of `TypeError`-ing.

On the test gap — agreed the mocked-`run_async` version was a hole. Replaced with `test_handle_envelope_dispatches_executable_error_task_when_dialogue_update_is_none` which actually awaits the dispatched task end-to-end and asserts both the `"Could not update dialogue"` scheduling log AND `_handle_error`'s drop log fire (the latter proves the task body ran without TypeError). Plus a direct unit test `test_handle_error_returns_none_when_dialogue_is_none` exercising the new branch.

Thanks for the second pair of eyes — having two reviewers independently land on the same bug made the fix uncontroversial.

DIvyaNautiyal07:
**ci: dropped `valory/contracts-amm` from the workflow pull list (commit `498655677`)**

Heads up that the latest commit on this PR removes an unrelated CI step that started 404'ing today. Posting the context so reviewers don't have to dig.

**Symptom.** `docker pull valory/contracts-amm:latest` returns `pull access denied for valory/contracts-amm, repository does not exist or may require 'docker login'` in the `Install dependencies` step. The same step succeeded on this branch through ~15:43Z on 2026-05-15; the failure window opens at ~16:18Z and persists. Affects every `test (*/*, py*)` and `e2e (*)` matrix cell because the step is the gate for all of them.

**Verification.**

- Docker Hub API: `GET /v2/repositories/valory/contracts-amm/` returns HTTP 404 right now. Sister images in the same workflow pull list (`valory/autonolas-registries`, `valory/safe-contract-net`, `valory/slow-tendermint-server`, `tendermint/tendermint:v0.34.19`, `trufflesuite/ganache:beta`) all return 200.
- Local reproduction with the same `docker pull` command surfaces the identical error.
- Same workflow file, same line, 35 minutes earlier on this branch (commit `ad704b5`, run `25926313585`): `Status: Downloaded newer image for valory/contracts-amm:latest`. So this is purely a Docker Hub state change.

**Why removal is safe.**

Static search confirms zero callers of the AMM fixture in this repo:

```bash
grep -rnE "contracts-amm|AMMNetDockerImage|amm_net_scope|UseAMM" \
  autonomy/ packages/ tests/ scripts/ plugins/aea-test-autonomy/tests/
# only hit: scripts/whitelist.py:67 — vulture whitelist labelling AMMNetDockerImage as "unused class"
```

The `AMMNetDockerImage` class lives in `plugins/aea-test-autonomy/aea_test_autonomy/docker/amm_net.py` and is exposed by `fixture_helpers.py` as part of the plugin's public API for downstream consumers (e.g. trader, market-creator), but **no test in this repo invokes the fixture**. The pull was dead weight that started actively breaking CI today.

Runtime verification with the image absent locally: all 2423 tests collect with zero errors, `tests/test_autonomy/test_chain/` (the docker-fixture-heavy bucket) runs 19/19 green, and `docker events --filter event=pull` captures zero pulls during the test run.

**This change is not part of the audit scope** — I'm bundling it here only because CI on this PR can't pass without it. Happy to split into a separate PR if maintainers prefer. The sister images (`valory/safe-contract-net`, `trufflesuite/ganache:beta`) are also dead weight under the same logic but resolved fine today, so I left them alone to keep the scope tight.

Pinging @valory-xyz/devops in case anyone has context on why `valory/contracts-amm` went away — it'd be worth knowing whether to restore the image or accept that it's permanently gone before we do similar cleanups elsewhere.

DhairyaPatel7 (approved):

LOCKhart07 (approved):

OjusWiZard (approved):
Approving after a thorough multi-round review.

**Review history:**
- **Round 1** — 11 findings (3 CI failures, 2 CRITICAL, 4 IMPORTANT, 6 ADVISORY) → all addressed in `6347cf602`, threads resolved.
- **Round 2** — 7 findings (1 HIGH, 2 MEDIUM, 4 ADVISORY) → 6 addressed in `5264789eb`, threads resolved. 1 advisory on `RoundSequenceNotInitialisedError` left open as author's design choice (keeping the broader `ValueError` catch is reasonable).
- **Round 3** — 1 HIGH (the round-2 None-guard called `_handle_error` with a missing positional arg) → addressed in `d0d583e63` by widening `_handle_error` to accept `Optional[BaseDialogue]`. Also fixed a pre-existing latent variant on the performative-not-supported path. New tests genuinely await the dispatched task end-to-end and would catch a regression.

**Resilience changes themselves look solid:**
- HTTP/IPFS/Subgraph error paths now surface as typed errors or logged drops rather than crashes/silent `None`-envelopes.
- Tendermint reset endpoints keep the established HTTP 200 + `status: False` contract with `logging.exception` for operator visibility.
- ApiSpecs `process_response` short-circuits 5xx (including the http_client synthetic 600) without trying to decode.
- `_check_sync` and `reset_tendermint_with_wait` distinguish transient (JSONDecodeError) from structural (KeyError/TypeError/ValueError) drift in their logs.
- The author also caught and fixed two sibling-copy gaps I missed in review: the docker-image variant of the Tendermint Flask app, and the Docker Hub `valory/contracts-amm` 404 unrelated to the PR.

**Test coverage:** every new error branch has an end-to-end test added in `5264789eb`/`6a6354f85`. Including the `_PUBLIC_RPC_FALLBACK_WARNED` warn-once test that clears module state in `finally`.

Static CI is fully green on `498655677`. Test matrix is re-running after the CI workflow fix and macos/windows already showed green on the previous run. LGTM.

jmoreira-valory (approved):

PR #2518, opened 2026-05-19
test: eliminate flaky tests at their root cause

DIvyaNautiyal07:
## Summary

Removes every known flaky test in the repo by fixing the root cause, not by adding retries or rerun markers. After this PR:

- No `@pytest.mark.flaky(reruns=*)` markers anywhere in the tree (was 6).
- No `_run_count` rerun-detection hack in `SlashingE2E`.
- Tests that talk to live upstreams (GitHub, chain RPCs) skip on transient network failure but still fail on real data drift.
- Tests that drive a real Tendermint binary on Windows / macOS are pinned to Linux (matching the existing `test_hard_reset_dev_mode` precedent).
- Subprocess / container readiness is now determined by the subject's own state (block height, `docker inspect` container State, hardhat banner), not by `time.sleep` + log substring matching.
- An ABCI-port-bind barrier in `BaseTestEnd2End.prepare_and_launch` eliminates the connection-refused race between Tendermint containers and agent startup, which was the actual observed flake source in `register_reset` e2e.

## What changed

Per-file LOC, real code only (the +1,200 mechanical noise from `autonomy packages lock`, `tox -e generate-api-documentation`, and `tox -e fix-doc-hashes` is excluded):

| File | Δ LOC | Change |
|------|-------|--------|
| `tests/utils/live_fetch.py` | +80 / -0 | NEW. `fetch_upstream_or_skip(url)` retries `ConnectionError` / `Timeout` / 5xx / 429 (5 attempts, ~31s total backoff). Skips on persistent failure; raises on real HTTP errors (4xx, content mismatch). |
| `tests/utils/__init__.py` | +20 / -0 | NEW. Package marker. |
| `tests/test_docs/helper.py` | +6 / -19 | A3. Both GitHub fetches (`api.github.com/.../releases/latest`, `raw.githubusercontent.com/...`) routed through the helper. `TestYamlSnippets::test_run_check` inherits the new behaviour without changes. |
| `tests/test_autonomy/test_chain/test_addresses.py` | +21 / -2 | A2. `setup_class` GitHub fetch goes through the helper. The per-test chain-RPC call is wrapped in `try/except (requests.ConnectionError, Timeout, HTTPError, Web3Exception, ValueError) → pytest.skip`. |
| `tests/test_autonomy/test_cli/test_develop/test_local_service_registry.py` | +57 / -18 | C1. New `_wait_for_container_ready` that polls the hardhat container's `docker logs` for the JSON-RPC banner. Replaces the 30 × 5s fixed-budget RPC retry loop. |
| `tests/test_autonomy/test_images/test_runtime.py` | +20 / -8 | B4. `_agent_is_running` now checks `docker_container.attrs["State"]["Status"] == "running"` plus the AEA banner. Timeout 30s → 180s for realistic cold start. On failure, full container State + logs are dumped into the assertion message. |
| `tests/test_deployments/test_app.py` | +37 / -8 | B1. `TestTendermintBufferWorking` is now `@pytest.mark.skipif(platform.system() != "Linux")` (matches `test_hard_reset_dev_mode`). `test_tendermint_buffer` asserts monotonic `latest_block_height` growth + ≥ 2 new blocks in 60s, retrying transient socket failures internally. |
| `packages/valory/connections/abci/tests/test_abci.py` | +35 / -7 | B3. `TestNoop.test_run` polls `sync_info.latest_block_height` until it advances within `DEADLINE_SECONDS = 30.0`. `_latest_block_height` returns `Optional[int]`; schema errors propagate. |
| `plugins/aea-test-autonomy/aea_test_autonomy/base_test_classes/agents.py` | +52 / -4 | D-shared. `_wait_for_abci_ports_bound` TCP-polls each agent's ABCI port before Tendermint dials. `prepare_and_launch` calls `terminate_agents()` on barrier failure so subprocesses don't leak. |
| `packages/valory/agents/counter/tests/test_counter.py` | +5 / -3 | F1. `random.Random(0)` seeded RNG for tx routing in `TestABCICounterSkillMany`. Flaky marker removed. |
| `packages/valory/agents/offend_slash/tests/test_offend_slash.py` | +0 / -31 | E2. `_run_count` rerun-detection hack, `setup_method` conditional, and leftover `teardown_method` deleted; `@pytest.mark.flaky` marker removed. |
| `packages/valory/agents/register_reset/tests/test_register_reset.py` | +0 / -3 | D1. Three `@pytest.mark.flaky(reruns=1)` markers removed (`TestTendermintStartup`, `TestTendermintResetInterrupt`, `TestTendermintResetInterruptNoRejoin`). |
| `packages/valory/agents/register_reset/tests/test_hard_reset_race_condition.py` | +0 / -1 | D2. `@pytest.mark.flaky(reruns=1)` marker removed. |

**Net: +333 / -104 lines of real test code. 6 `@pytest.mark.flaky` markers removed across 4 files.**

The rest of the diff (~1,200 additional lines across 38 yaml / json / .md files) is mechanical: `autonomy packages lock` propagated new IPFS hashes through every dependent agent / skill / service config, `tox -e fix-doc-hashes` refreshed `docs/package_list.md`, and `tox -e generate-api-documentation` refreshed `docs/api/**` for the new docstrings.

## Test plan

- [ ] CI passes on this branch
- [ ] CI passes a second time without rerunning (the real proof — flaky tests come back as `failure`, deterministic ones don't)
- [ ] Spot-check the `register_reset` e2e job in particular — that's where the connection-refused race was observed

## Verified locally

- `tomte format-code` clean
- `tox -e black-check`, `tox -e isort-check`, `tox -e flake8` all green
- `tox -e mypy`, `tox -e pylint`, `tox -e darglint` all green
- `tox -e check-dependencies`, `tox -e check-packages`, `tox -e check-hash` all green
- `tox -e check-api-docs` green after regenerating
- `autonomy packages lock --check` verifies the regenerated hashes
- `tox -e fix-doc-hashes` ran (docs already in sync)

The integration / e2e tests themselves need real Docker images and aren't run locally; CI on this PR is the canonical verification.


tests/test_autonomy/test_chain/test_addresses.py
```diff
@@ -72,7 +74,13 @@ def test_addresses_match(
             address = contract["address"]
             if name == "service_manager_proxy":
                 monkeypatch.setenv(f"{chain.value.upper()}_CHAIN_RPC", chain.rpc)
-                constant_address = ContractConfigs.service_manager.contracts[chain]
+                try:
+                    constant_address = ContractConfigs.service_manager.contracts[chain]
+                except (requests.ConnectionError, requests.Timeout) as exc:
```
DIvyaNautiyal07: **P1 — RPC catch is narrower than the failure modes the test claims to cover.** `DynamicContract.__getitem__` (`autonomy/chain/config.py:346`) does a `web3` `.call()` that can surface `requests.HTTPError` (5xx from RPC), `web3.exceptions.*`, plain `ValueError` from JSON-RPC decode, and `KeyError` if `chain` is missing from `CHAIN_PROFILES`. The observed CI failure (Arbitrum `NameResolutionError`) IS a `requests.ConnectionError` so this catch handles the seen case, but the broader contract "skip on any transient RPC failure" isn't met. Consider broadening to include `requests.HTTPError`, `ValueError`, `KeyError`, or `except Exception` with an intent comment.

tests/test_autonomy/test_images/test_runtime.py
```diff
@@ -199,20 +199,30 @@ def test_image(self) -> None:
...
+        def _agent_is_running() -> bool:
+            """The agent container is up and the process is alive."""
+            agent_container.reload()
+            state = agent_container.attrs.get("State", {})
+            # Status transitions: created -> running -> exited.
+            # "running" with no ExitCode and no death signal is success.
+            if state.get("Status") != "running":
+                return False
+            if state.get("ExitCode"):
+                return False
```
DIvyaNautiyal07: **P2 — `ExitCode` guard is dead code.** Docker's `ExitCode` is `0` both while the container is running AND after a clean exit. The `Status != "running"` guard at line 208 already handles "not running." So `if state.get("ExitCode"):` never fires when reached.

Separately, the commit message says "stop matching log strings" but line 212 still matches the log string — the docker-inspect parts are early-exit gates, not the actual fix. The real fix is the 30s → 180s timeout. Either remove the dead guard and clarify the docstring, or rewrite to check `state.get("Error")` / `OOMKilled` / `ExitCode not in (None, 0)`.

packages/valory/connections/abci/tests/test_abci.py
```diff
@@ -471,14 +471,37 @@ class TestNoop(BaseABCITest, BaseTestABCITendermintIntegration):
 
     SECONDS = 5
```
DIvyaNautiyal07: **P2 — `SECONDS = 5` is unused after the rewrite.** Was referenced by the old `time.sleep(self.SECONDS)`. The new code hardcodes `30.0` at line 491. Remove the class attribute, or use it for the deadline so subclasses can override it.

tests/utils/live_fetch.py
```diff
@@ -0,0 +1,83 @@
...
+    session = requests.Session()
+    session.mount("https://", adapter)
+    session.mount("http://", adapter)
+    return session
+
+
+def fetch_upstream_or_skip(
+    url: str,
+    timeout: int = _DEFAULT_TIMEOUT,
+    session: requests.Session = None,
```
DIvyaNautiyal07: **P2 — annotation lacks `Optional[...]`.** `session: requests.Session = None` declares a non-optional type with a `None` default. Project mypy doesn't catch it (implicit-optional enabled) but it's still wrong style and breaks under strict mypy. Fix:

```python
from typing import Optional
...
session: Optional[requests.Session] = None,
```

plugins/aea-test-autonomy/aea_test_autonomy/base_test_classes/agents.py
```diff
@@ -272,11 +273,56 @@ def prepare(self, nb_nodes: int) -> None:
...
+            while time.monotonic() < deadline:
+                try:
+                    with socket.create_connection(("127.0.0.1", port), timeout=1.0):
+                        last_error = None
+                        break
+                except (OSError, socket.timeout) as exc:
+                    last_error = exc
+                    time.sleep(self.ABCI_BIND_POLL_INTERVAL)
+            if last_error is not None:
+                raise AssertionError(
```
DIvyaNautiyal07: **P2 — agents leaked if barrier raises.** If `_wait_for_abci_ports_bound` raises `AssertionError`, the agents already launched in the `for agent_id in range(nb_nodes)` loop above (line 287-288) are not terminated before the exception propagates. Across test reruns or CI parallelism this can leave orphan processes. Wrap the call in try/except, call `terminate_agents()` (or the equivalent cleanup helper) before re-raising.

packages/valory/connections/abci/tests/test_abci.py
```diff
@@ -471,14 +471,37 @@ class TestNoop(BaseABCITest, BaseTestABCITendermintIntegration):
...
 
+    def _latest_block_height(self) -> int:
+        """Return the node's current ``latest_block_height`` (or -1 on failure)."""
+        try:
+            response = requests.get(self.tendermint_url() + "/status", timeout=30)
+            response.raise_for_status()
+            return int(response.json()["result"]["sync_info"]["latest_block_height"])
+        except (requests.RequestException, KeyError, ValueError):
```
DIvyaNautiyal07: **P2 — sentinel conflates failure modes.** `_latest_block_height` swallows `KeyError`/`ValueError` into the same `-1` sentinel as `requests.RequestException`. If Tendermint changes its JSON shape on a version bump, the missing key looks identical to "RPC down" — the test silently keeps polling and fails on the 30s deadline rather than failing loudly with the schema error. Let `KeyError`/`ValueError` propagate; only suppress `requests.RequestException`.

packages/valory/connections/abci/tests/test_abci.py
```diff
@@ -471,14 +471,37 @@ class TestNoop(BaseABCITest, BaseTestABCITendermintIntegration):
...
+
     def test_run(self) -> None:
-        """
-        Run the test.
+        """Assert the node produces blocks.
 
-        Sleep for N seconds, check Tendermint is still running, and then stop the test.
+        The original implementation slept for N seconds and then did a single
+        /health probe. That conflates "Tendermint is slow to start" with
+        "Tendermint is broken." We assert on the real claim instead: the node
```
DIvyaNautiyal07: **Nit — historical narration in docstring.** "The original implementation slept for N seconds and then did a single /health probe. That conflates..." belongs in the commit message, not the docstring. Per project conventions in `CLAUDE.md`, comments describe invariants, not change history. Suggested rewrite:

```
"""Assert the node produces at least one new block within 30 seconds.

Polling for monotonic ``latest_block_height`` growth distinguishes
"Tendermint is slow to start" from "Tendermint has stalled."
"""
```

tests/test_autonomy/test_cli/test_develop/test_local_service_registry.py
```diff
@@ -55,14 +55,46 @@ def setup_method(self) -> None:
...
+        """Block until the hardhat container prints its readiness marker.
+
+        Reads the container's accumulated logs once per second. Hardhat
+        prints the JSON-RPC banner exactly once it is ready to accept
+        connections.
+
+        :raises AssertionError: if the marker is not observed within
+            ``READY_TIMEOUT_SECONDS``.
+        """
+        client = docker.from_env()
```
DIvyaNautiyal07: **Nit — `docker.from_env()` called per loop iteration.** Up to 300 iterations × a new Docker socket connection (worst case). Hoist out of the loop:

```python
client = docker.from_env()
while time.monotonic() < deadline:
    try:
        container = client.containers.get(CONTAINER_NAME)
        ...
```

tests/test_autonomy/test_cli/test_develop/test_local_service_registry.py
```diff
@@ -55,14 +55,46 @@ def setup_method(self) -> None:
...
+        deadline = time.monotonic() + self.READY_TIMEOUT_SECONDS
+        last_error: object = None
```
DIvyaNautiyal07: **Nit — inconsistent type-dodge pattern.** `last_error: object = None` here vs `last_error: Optional[BaseException] = None` at `agents.py:310` in the same PR. The same construct in two files with two different annotations. Both exist to dodge mypy on the `None` initialiser. Pick one — `Optional[docker.errors.NotFound]` is the tightest correct type here since that's the only exception type ever assigned to it.

tests/utils/live_fetch.py
```diff
@@ -0,0 +1,83 @@
...
+        status_forcelist=_TRANSIENT_STATUS,
+        backoff_factor=1.0,
+        raise_on_status=False,
+        allowed_methods=frozenset(["GET", "HEAD", "POST"]),
+    )
+    adapter = HTTPAdapter(max_retries=retry)
+    session = requests.Session()
+    session.mount("https://", adapter)
+    session.mount("http://", adapter)
+    return session
```
DIvyaNautiyal07, line 58: **Nit — `live_session()` not earning its keep.** Only called from `fetch_upstream_or_skip` at line 73 with no external callers, and no caller passes the `session` param. The abstraction is speculative. Either make it private (`_make_session`) or inline it. Wait for a second consumer before exposing it.

tests/test_deployments/test_app.py
```diff
@@ -509,19 +518,40 @@ class TestTendermintBufferWorking(BaseTendermintServerTest):
...
+                continue
+            assert height >= last_height, (
+                f"Tendermint block height went backwards: " f"{last_height} -> {height}"
+            )
+            if height > last_height:
+                progress_count += 1
+                last_height = height
             time.sleep(1)
 
+        assert progress_count >= 2, (
```
OjusWiZard: **Critical — likely the source of the current Python 3.10/3.14 Ubuntu CI failures.**

Requiring `progress_count >= 2` in a 60s window on a cold GHA runner is tight: Tendermint setup can burn 20–40s before block 1 lands, leaving very little headroom for block 2. The previous `flaky(reruns=1)` was masking this; concentrating the test on Linux-only via the new `skipif` makes the failure deterministic on the very leg where it's most fragile.

Suggested: assert `progress_count >= 1` (one observed height change is sufficient proof of liveness) or raise the deadline to `120.0`.

tests/test_deployments/test_app.py
```diff
@@ -509,19 +518,40 @@ class TestTendermintBufferWorking(BaseTendermintServerTest):
...
             try:
                 res = requests.get(self.tm_status_endpoint, timeout=5)
-                # We expect all responses to be OK
-                assert res.status_code == 200
-            except Exception as e:
-                raise AssertionError(e)
-
+                res.raise_for_status()
+                height = int(res.json()["result"]["sync_info"]["latest_block_height"])
+            except (requests.RequestException, KeyError, ValueError):
```
OjusWiZard: Catching `KeyError` and `ValueError` here treats a Tendermint `/status` schema change identically to a transient socket hiccup. If the response body is consistently missing `result.sync_info.latest_block_height` (e.g. a Tendermint version bump renamed the field), every iteration `continue`s, `progress_count` stays at 0, and the test eventually fails with "fewer than 2 blocks in 60s" — the wrong reason.

Suggested: catch only `requests.RequestException`. Let `KeyError` / `ValueError` propagate with the raw `res.json()` body in the message so the real cause is visible.

tests/test_autonomy/test_cli/test_develop/test_local_service_registry.py
```diff
@@ -55,14 +55,46 @@ def setup_method(self) -> None:
...
+        :raises AssertionError: if the marker is not observed within
+            ``READY_TIMEOUT_SECONDS``.
+        """
+        client = docker.from_env()
+        deadline = time.monotonic() + self.READY_TIMEOUT_SECONDS
+        last_error: Optional[docker.errors.NotFound] = None
+        while time.monotonic() < deadline:
+            try:
+                container = client.containers.get(CONTAINER_NAME)
+                if self.READY_MARKER in container.logs():
```
OjusWiZard: **Critical — log polling is a proxy, not a readiness gate.**

The Hardhat JSON-RPC banner can print *before* the TCP `listen()` actually starts accepting connections — the print is synchronous, the listen happens shortly after. The previous RPC-retry loop was the ground-truth gate. If Hardhat ever changes the banner text (it already has between versions), this test silently hangs for 300s and then fails with a cryptic AssertionError.

Suggested: keep the RPC poll as the readiness check (`requests.get` retrying `ConnectionRefused`) and use this log read only as a diagnostic *after* the RPC poll times out.

Separately: `container.logs()` returns *all* accumulated logs on every iteration — O(n × polls) data transfer. Pass `tail=200` or `since=<last_ts>` to bound the fetch.

tests/test_autonomy/test_chain/test_addresses.py
```diff
@@ -39,7 +42,7 @@ class TestAddresses:
     @classmethod
     def setup_class(cls) -> None:
         """Setup test class."""
-        chain_configs = requests.get(url=ADDRESS_FILE_URL, timeout=30).json()
+        chain_configs = fetch_upstream_or_skip(ADDRESS_FILE_URL).json()
```
OjusWiZard: **Critical — `pytest.skip()` in `setup_class` makes the entire class disappear from the report.**

If the pinned `v1.3.0` GitHub URL is ever removed/moved (or GitHub returns persistent 5xx during a real outage), every test in `TestAddresses` silently vanishes from the run with no alarm — the test IDs don't even show up as `s`/skipped, they're just gone. The whole point of this class is to catch upstream constant drift; that signal is now contingent on GitHub availability with no escape valve.

Suggested: cache the upstream fetch result (or a sentinel) at class scope, but call `pytest.skip` from each test method individually. Skip events at least then appear in the CI report.

tests/test_autonomy/test_chain/test_addresses.py
```diff
@@ -72,7 +75,21 @@ def test_addresses_match(
...
                 monkeypatch.setenv(f"{chain.value.upper()}_CHAIN_RPC", chain.rpc)
-                constant_address = ContractConfigs.service_manager.contracts[chain]
+                try:
+                    constant_address = ContractConfigs.service_manager.contracts[chain]
+                except (
+                    requests.ConnectionError,
+                    requests.Timeout,
+                    requests.HTTPError,
+                    Web3Exception,
+                    ValueError,
```
OjusWiZard: `ValueError` is too broad. `web3` raises it for malformed addresses, bad function selectors, and ABI decode failures — all real bugs this test exists to catch. Catching it converts those into silent skips.

Suggested: narrow to specific transport-level exceptions, e.g. `(requests.ConnectionError, requests.Timeout, requests.HTTPError, web3.exceptions.BadResponseFormat, web3.exceptions.BadFunctionCallOutput)`. Let plain `ValueError` propagate as a real failure.

packages/valory/connections/abci/tests/test_abci.py
```diff
@@ -469,16 +469,48 @@ async def process_incoming_messages(self) -> None:
...
+
+        :return: the current block height, or ``None`` if the node is
+            unreachable.
         """
-        Run the test.
+        try:
+            response = requests.get(self.tendermint_url() + "/status", timeout=30)
+        except requests.RequestException:
+            return None
+        if not response.ok:
```
OjusWiZard: **Critical — `if not response.ok: return None` treats a 5xx / crash-loop identically to "node not up yet".**

The docstring says "Schema errors are not suppressed" — correct — but a persistent `503` or a Tendermint process that's crash-looping with HTTP-but-bad-status responses also returns `None`. The 30s deadline then fires with "Tendermint produced no new blocks" rather than the real cause ("node returned 503 for 30s").

Suggested: replace `if not response.ok: return None` with `response.raise_for_status()` inside the `try`. Either let HTTPError propagate (so the test fails with the real reason), or extend the `except` to include it if 5xx genuinely should be retried — but document it.

plugins/aea-test-autonomy/aea_test_autonomy/base_test_classes/agents.py
```diff
@@ -272,11 +273,65 @@ def prepare(self, nb_nodes: int) -> None:
...
+            while time.monotonic() < deadline:
+                try:
+                    with socket.create_connection(("127.0.0.1", port), timeout=1.0):
+                        last_error = None
+                        break
+                except OSError as exc:
+                    # socket.timeout is a subclass of OSError on Py3.10+
+                    last_error = exc
+                    time.sleep(self.ABCI_BIND_POLL_INTERVAL)
+            if last_error is not None:
```
OjusWiZard: The guard `if last_error is not None` silently treats "loop body never executed" as success.

If `_launch_agent_i` already burned the budget so `time.monotonic() >= deadline` on entry, the `while` body never runs, `last_error` stays `None`, and the function returns. Tendermint then dials a port that nothing is listening on — failure surfaces later as opaque "connection refused" in Tendermint logs, defeating the whole point of the barrier.

Suggested:
```python
connected = False
while time.monotonic() < deadline:
    try:
        with socket.create_connection(("127.0.0.1", port), timeout=1.0):
            connected = True
            break
    except OSError as exc:
        last_error = exc
        time.sleep(self.ABCI_BIND_POLL_INTERVAL)
if not connected:
    raise AssertionError(...)
```

tests/utils/live_fetch.py
```diff
@@ -0,0 +1,80 @@
...
+_TRANSIENT_STATUS = (500, 502, 503, 504, 429)
+
+
+def _build_session() -> requests.Session:
+    """Build a requests session that retries transient network errors."""
+    retry = Retry(
+        total=5,
+        connect=5,
+        read=5,
+        status=5,
```
OjusWiZard: `Retry(total=5, connect=5, read=5, status=5, ...)` — each subtype counts against `total`, not in addition to it. A mixed failure sequence (connect-fail → read-fail → 5xx) exhausts retries faster than the docstring's "5 attempts" implies. For a real DNS flap the effective retry count may end up being 2–3.

Suggested: drop the subtype counts and rely on `total=5` alone, or set `total=None` and let the subtypes accumulate independently.

tests/utils/live_fetch.py
```diff
@@ -0,0 +1,80 @@
...
+    """
+    try:
+        response = _build_session().get(url, timeout=timeout)
+    except (requests.ConnectionError, requests.Timeout) as exc:
+        pytest.skip(f"upstream unreachable after retries: {url} ({exc})")
+
+    if response.status_code in _TRANSIENT_STATUS:
+        pytest.skip(f"upstream returned {response.status_code} after retries: {url}")
+
+    response.raise_for_status()
```
OjusWiZard, line 93: `response.raise_for_status()` makes 4xx behavior diverge from 5xx (skip) without documenting it — and the `:raises:` block in the docstring doesn't mention `requests.HTTPError`.

A 404 means the URL is wrong (real bug, should fail), but a 403 from GitHub rate-limiting probably warrants a skip. Worth deciding explicitly and adding both a comment and a `:raises requests.HTTPError:` line to the docstring (darglint will flag the missing entry).

OjusWiZard (commented):
Inline review notes — 4 Critical, 5 Important, 1 Advisory. The direction (root-cause over reruns) is right, but several of the new readiness predicates are proxies rather than ground-truth signals — same flake class, different shape. The Python 3.10/3.14 Ubuntu CI failures are almost certainly the `TestTendermintBufferWorking` `progress_count >= 2` assertion (see inline).

tests/test_deployments/test_app.py
DIvyaNautiyal07: Done in `7ff2ba28e`. Both halves of the suggestion: deadline raised 60s → 120s, and the claim relaxed to a single observed height advance (the loop returns on the first `height > last_height`). One block is sufficient proof of liveness, and we no longer need the runner to fit two blocks into the window.

tests/test_deployments/test_app.py
DIvyaNautiyal07: Done in `7ff2ba28e`. The `except` is narrowed to `requests.RequestException` only; `KeyError` / `ValueError` from the JSON path now propagate with the raw exception in the failure message. The `res.json()[...]` chain is moved outside the try so the schema parse can't be silently retried.

tests/test_autonomy/test_cli/test_develop/test_local_service_registry.py
DIvyaNautiyal07: Done in `7ff2ba28e`. Switched the readiness gate back to RPC polling (`requests.get(self.expected_network_address)` retrying `ConnectionError`); the docker log read is kept but only as a diagnostic dumped into the AssertionError on timeout (`tail=200` so it doesn't pull the full log on every iteration). The banner-text fragility concern is gone — banner only shows up in the failure message, not in the gate condition.

tests/test_autonomy/test_chain/test_addresses.py
DIvyaNautiyal07: Done in `7ff2ba28e`. Replaced `setup_class` with a lazy class-level cache (`_contracts: Optional[...]`) populated by `_get_contracts()`. Each parametrized `test_addresses_match` calls it, so a GitHub outage now shows up as `N skipped` entries with the reason — one per chain — instead of silently disappearing.

tests/test_autonomy/test_chain/test_addresses.py
DIvyaNautiyal07: Done in `7ff2ba28e`. Bare `ValueError` is gone. The catch is now a named tuple `_TRANSIENT_RPC_ERRORS` containing `requests.{ConnectionError, Timeout, HTTPError}` and `web3.exceptions.{BadResponseFormat, BadFunctionCallOutput, ProviderConnectionError, TooManyRequests}`. Malformed addresses / ABI decode failures (which are `ValueError` from web3) now propagate as real test failures.

packages/valory/connections/abci/tests/test_abci.py
DIvyaNautiyal07: Done in `7ff2ba28e`. Replaced `if not response.ok: return None` with `response.raise_for_status()` inside the try. A persistent 503 now propagates as `HTTPError` via the schema-parse path, so the assertion message reports the real cause (`requests.HTTPError: 503 ...`) instead of "Tendermint produced no new blocks". `None` is now reserved strictly for transport-layer failures.

plugins/aea-test-autonomy/aea_test_autonomy/base_test_classes/agents.py
DIvyaNautiyal07: Done in `7ff2ba28e`. Replaced the `last_error is None` check with an explicit `connected = False` flag set to `True` only inside the successful `socket.create_connection` block. The degenerate case (`deadline` already past on entry, loop body never executes) now raises `AssertionError` with the last-seen error instead of silently returning.

tests/utils/live_fetch.py
DIvyaNautiyal07: Done in `7ff2ba28e`. Dropped `connect=5, read=5, status=5` from the `Retry` config; `total=5` is now the single budget. Added a comment in `_build_session` documenting that urllib3 counts subtype failures against `total` rather than independently.

tests/utils/live_fetch.py
DIvyaNautiyal07: Done in `7ff2ba28e`. The docstring now explicitly enumerates the behaviour-per-response-category (connect/read/timeout → skip; 5xx/429 → skip; other 4xx → `raise_for_status` → fail; 2xx → return). 404/403 deliberately fail because a wrong URL or wrong access pattern is a real bug, not transient infra.

OjusWiZard (commented):
Round-3 review (post round-2 fixes). All 9 prior findings are confirmed addressed. Three new items below — one worth fixing (#1), one judgment call (#2), one cleanup (#3). Remaining advisories are minor polish and were left out of inline comments.

tests/test_autonomy/test_cli/test_develop/test_local_service_registry.py
```diff
@@ -55,14 +55,61 @@ def setup_method(self) -> None:
...
+        """
+        deadline = time.monotonic() + self.READY_TIMEOUT_SECONDS
+        last_error: Optional[Exception] = None
+        while time.monotonic() < deadline:
+            try:
+                res = requests.get(self.expected_network_address, timeout=5)
+                if res.status_code == 200:
+                    return
+                last_error = AssertionError(f"RPC returned {res.status_code}")
+            except requests.ConnectionError as exc:
```
OjusWiZard: **Important — `requests.Timeout` escapes this poll loop.**

`requests.ReadTimeout` is **not** a subclass of `requests.ConnectionError` — both descend from `requests.RequestException` as siblings. With `timeout=5` against a container still deploying ~12 contracts, a read timeout (TCP accepted, HTTP response slow on a loaded runner) propagates uncaught: the loop aborts, the diagnostic log-tail block below is skipped, and the test fails with a bare `ReadTimeout` traceback — re-introducing the transient-abort flake class this PR exists to eliminate.

It is also inconsistent within the PR: `test_abci.py:_latest_block_height` and `test_app.py:test_tendermint_buffer` both catch the base `requests.RequestException`. Suggest matching them here:

```python
except requests.RequestException as exc:
    last_error = exc
```

tests/test_autonomy/test_cli/test_develop/test_local_service_registry.py
```diff
@@ -55,14 +55,61 @@ def setup_method(self) -> None:
...
+    # Total budget for the hardhat container to come up and start
+    # answering JSON-RPC requests. Generous because the container has to
+    # pull the image, install npm deps, and deploy ~12 contracts on a
+    # cold runner.
+    READY_TIMEOUT_SECONDS = 300
+
+    def _wait_for_rpc_ready(self) -> None:
+        """Block until the hardhat RPC endpoint accepts requests.
+
+        The RPC's response to ``eth_chainId`` is the ground-truth gate
```
OjusWiZard: **Important — docstring describes an `eth_chainId` gate the code does not perform.**

The code (line 83) issues a bare `requests.get(self.expected_network_address, ...)` with no JSON-RPC payload; Hardhat returns `200` to any GET. A 200 proves the HTTP listener is up, not that JSON-RPC dispatch works — which is round-1's "proxy, not a gate" concern in a new shape.

Either send a real `eth_chainId` POST and assert a valid JSON-RPC `result` (delivers what this docstring promises and what production callers actually exercise), or correct the docstring to describe the bare-GET check honestly.

tests/test_deployments/test_app.py
```diff
@@ -509,19 +518,50 @@ class TestTendermintBufferWorking(BaseTendermintServerTest):
...
+        and surface as the actual exception so a Tendermint version
+        bump that renames a field fails the test loudly rather than
+        silently timing out.
+        """
 
-        # Give the test 60 seconds for it to work
-        for _ in range(60):
+        deadline = time.monotonic() + 120.0
+        last_height: int = -1
+        progress_count = 0
```
OjusWiZard: **Minor — `progress_count` is dead scaffold left from the round-1 fix.**

`progress_count` is incremented once, then `if progress_count >= 1` (line 553) is trivially true on that same iteration — it never reaches 2. It is residue from when the threshold was `>= 2`; the round-2 change to `>= 1` left the counter behind. No behaviour bug, but it misleads readers into thinking a threshold is enforced. Collapses to:

```python
if height > last_height:
    return  # one forward step proves liveness
```

tests/test_autonomy/test_cli/test_develop/test_local_service_registry.py
DIvyaNautiyal07: Done in `dd82bfd69`. Widened the catch to `requests.RequestException` so `ReadTimeout` (sibling of `ConnectionError` under the base) is also captured into `last_error` and retried. Matches `test_abci.py` and `test_app.py` which both use the base class.

tests/test_autonomy/test_cli/test_develop/test_local_service_registry.py
DIvyaNautiyal07: Done in `dd82bfd69`. The bare `GET` is replaced with a real JSON-RPC POST of `eth_chainId`. The success condition is now `status_code == 200` plus a body that's a dict containing `result` — proves JSON-RPC dispatch works, not just that the HTTP listener is up. The docstring now matches the code.

tests/test_deployments/test_app.py
DIvyaNautiyal07: Done in `dd82bfd69`. Collapsed to a direct `return` on the first observed forward step. The counter and the trivially-true `>= 1` guard are gone.

tests/test_deployments/test_app.py
```diff
@@ -509,19 +518,43 @@ class TestTendermintBufferWorking(BaseTendermintServerTest):
...
+                # Transport hiccup; the next iteration will observe the
+                # real state.
+                time.sleep(1)
+                continue
+            res.raise_for_status()
+            height = int(res.json()["result"]["sync_info"]["latest_block_height"])
+            assert height >= last_height, (
+                f"Tendermint block height went backwards: " f"{last_height} -> {height}"
+            )
+            if height > last_height:
```
DhairyaPatel7: Regression — the loop returns on the first observed height, not on actual forward progress. `last_height` is initialised to `-1` and never reassigned inside the loop, so any well-formed `/status` response (heights are always >= 0, and Tendermint serves `/status` from height 0 onward — before block 1 is committed) makes `height > last_height` True on the very first iteration and the function returns. The docstring's contract ("polling for monotonic latest_block_height growth", "one observed forward step proves liveness") is therefore not satisfied: the test passes even if Tendermint is up but stuck at genesis. This is also strictly weaker than the original implementation it replaces, which at least asserted `status_code == 200` for ~60 consecutive polls.

Fix: mirror the pattern in `test_abci.py::test_run` — establish a baseline on the first observation, then return only when a later observation exceeds it. e.g.:
```python
baseline: Optional[int] = None
while time.monotonic() < deadline:
    try:
        res = requests.get(self.tm_status_endpoint, timeout=5)
    except requests.RequestException:
        time.sleep(1)
        continue
    res.raise_for_status()
    height = int(res.json()["result"]["sync_info"]["latest_block_height"])
    if baseline is None:
        baseline = height
    elif height > baseline:
        return
    time.sleep(1)
```
As a side-effect of this fix the `assert height >= last_height` backwards-going-height guard also becomes meaningful — right now it never fires because the function returns before `last_height` is ever set to a real value.

tests/test_autonomy/test_chain/test_addresses.py
```diff
@@ -24,26 +24,62 @@
...
-        }
+    def _get_contracts(cls) -> t.Dict[str, t.List[t.Dict[str, str]]]:
+        """Fetch (or return cached) upstream contracts mapping.
+
+        Calls ``pytest.skip`` inside the current test if the upstream is
+        unreachable.
+
+        :return: the contracts dict keyed by chain name.
+        """
+        if cls._contracts is None:
```
DhairyaPatel7, line 85: On a persistent GitHub outage every parametrized chain (~10 of them) re-runs the full `fetch_upstream_or_skip` retry budget (~31s of backoff) before skipping, so total wall-time on a hard outage is ~5 minutes of dead retries. Cache the skip outcome too — e.g. store a sentinel like `_FETCH_FAILED` (or stash the skip reason in another class attribute) so the second-through-nth call short-circuits to `pytest.skip(...)` immediately.

plugins/aea-test-autonomy/aea_test_autonomy/base_test_classes/agents.py
```diff
@@ -272,11 +273,72 @@ def prepare(self, nb_nodes: int) -> None:
...
+        :raises AssertionError: if any agent fails to bind within the
+            per-agent timeout. Includes diagnostic context (which agent,
+            which port, last connect error) so failures are actionable.
+        """
+        for agent_id in range(nb_nodes):
+            port = self.get_abci_port(agent_id)
+            deadline = time.monotonic() + self.ABCI_BIND_TIMEOUT_SECONDS
+            connected = False
+            last_error: Optional[OSError] = None
+            while time.monotonic() < deadline:
```
DhairyaPatel7: `_wait_for_abci_ports_bound` walks agents sequentially with a 60s per-agent budget, so a totally dead agent-0 will burn 60s before agents 1..N are even checked, and on a 4-node test a fully-dead launch takes ~240s to surface as `AssertionError`. The agents are launched in parallel; consider checking them in parallel too (e.g. one shared deadline + round-robin over the unbound set) so the diagnostic is `"agents [0,2] did not bind"` after one timeout rather than after four serial timeouts.

DhairyaPatel7 (dismissed):
Verified against `dd82bfd69`. Round-3 items addressed:

- `test_local_service_registry.py::_wait_for_rpc_ready` now issues an `eth_chainId` POST and validates a JSON-RPC `result` body, so the gate matches the docstring and exercises the same shape production callers use. The exception class is widened to `requests.RequestException`, so `ReadTimeout` (sibling of `ConnectionError`) is also retried into `last_error` rather than escaping the loop.
- `test_app.py::test_tendermint_buffer` dropped the trivially-true `progress_count` scaffold; the test now returns on the first observed forward step with no intermediate counter.

CI on the round-2 commit (`7ff2ba28`) was green; the round-3 matrix is in progress at review time.

A few non-blocking observations that came up on a final pass — all judgement calls, none blocking:

1. **`test_tendermint_buffer` sentinel weakens the liveness claim.** `last_height = -1` means the first successful `/status` poll satisfies `height > last_height` for any non-negative height, and the loop returns immediately — without requiring a *new* block to land during the test window. The sibling `test_abci.py::TestNoop.test_run` is stricter: it sets `baseline` from a first probe and returns only on `current > baseline`. Both tests verify the same property; aligning them would make the assertion consistent.

2. **`fetch_upstream_or_skip` worst-case latency is ~210s per call.** `_DEFAULT_TIMEOUT=30` × 6 attempts + 30s backoff sum (0+2+4+8+16). For parametrized callers that invoke it per-iteration (e.g. `test_addresses_match` across many chains), a real GitHub outage burns several minutes before the skip lands. A shorter per-attempt timeout (~10s) would surface the skip faster without affecting the healthy path.

3. **`_wait_for_abci_ports_bound` is sequential per agent** with a 60s per-agent budget; `nb_nodes=4` ⇒ worst case 240s before the barrier fails. If a real bind failure happens on agent 3, agents 0–2 are still polled for 60s each first. Parallel polling across agents would expose the failure faster.

tests/test_deployments/test_app.py
```diff
@@ -509,19 +518,43 @@ class TestTendermintBufferWorking(BaseTendermintServerTest):
...
+                return  # one observed forward step proves liveness
```
OjusWiZard: **Important — this loop no longer tests liveness; the round-3 cleanup removed the line that made it one.**

Collapsing `progress_count` here also deleted `last_height = height` — the only assignment that ever moved `last_height` off its `-1` initializer.

Trace: `last_height` stays `-1` for the whole loop, and Tendermint heights are `>= 0`.
- `assert height >= last_height` → `height >= -1` → tautology, can never fail. The "monotonic growth" check the docstring describes never runs.
- `if height > last_height` → `height > -1` → always true on the first successful poll → returns after **one** `/status` request, regardless of block production.
- Since `@wait_for_node_to_run` already gates on `/status` returning 200, the retry/timeout paths are unreachable too. The timeout `AssertionError` ("did not advance `latest_block_height`") can now only fire if the node is *unreachable* for 120s — which the decorator already prevented.

Net: the test is effectively guaranteed-green. Its only remaining value is catching a `/status` JSON schema rename via `int(res.json()[...])`. The docstring ("keeps making progress", "producing blocks", "monotonic ... growth") is now false.

Fix — mirror `TestNoop.test_run` in `test_abci.py`, which does it correctly: capture a `baseline`, return only when a *later* poll exceeds it:

```python
baseline: Optional[int] = None
while time.monotonic() < deadline:
    ...
    height = int(res.json()["result"]["sync_info"]["latest_block_height"])
    if baseline is None:
        baseline = height
    elif height > baseline:
        return
    elif height < baseline:
        raise AssertionError(f"height went backwards: {baseline} -> {height}")
    time.sleep(1)
```

OjusWiZard (commented):
Round-4 review (post round-3 commit `dd82bfd69`). The two `_wait_for_rpc_ready` fixes in that commit — widening the catch to `requests.RequestException` and gating on a real `eth_chainId` JSON-RPC response — are both correct. Two items below: one Important (the `progress_count` collapse also removed the assignment that made `test_tendermint_buffer` a real liveness check) and one Advisory.

packages/valory/connections/abci/tests/test_abci.py
```diff
@@ -469,16 +469,49 @@ async def process_incoming_messages(self) -> None:
...
+        — a node that returns 503 for 30s is a real failure, not "node
+        not up yet", and the assertion message at the top level needs
+        to surface that.
+
+        :return: the current block height, or ``None`` if the node is
+            unreachable at the transport layer.
         """
-        Run the test.
+        try:
+            response = requests.get(self.tendermint_url() + "/status", timeout=30)
```
OjusWiZard: **Advisory — per-request `timeout=30` equals the whole `DEADLINE_SECONDS=30`.**

If a request *hangs* (node accepts the TCP connection but never answers), this single `requests.get` blocks for the full 30s. By the time it returns, `time.monotonic() < deadline` in `test_run` is already false, so the polling loop can execute **zero** times and fails with `Tendermint produced no new blocks within 30s (baseline=None)` — misleading, since the real cause is unreachability, not a stall.

The happy path is unaffected. Suggest a small per-request timeout (~5s) so the loop keeps polling within the deadline.

tests/test_deployments/test_app.py
LOCKhart07: **Blocking — `last_height = -1` makes the progression check dead code.**

`last_height` is initialised to `-1` and never reassigned inside the loop. On the first successful poll, any non-negative height satisfies `height > last_height` and the function returns. Since the `BaseTendermintServerTest` setup runs with `CREATE_EMPTY_BLOCKS=true` and `wait_for_node_to_run` only waits for `/status` to respond 200 (not for a block to be produced), the test now passes as long as `/status` answers once — even on a Tendermint that has produced zero blocks.

This contradicts the docstring ("the node is alive and producing blocks") and the round-2 ask ("one observed height *change* is sufficient"). It also diverges from `packages/valory/connections/abci/tests/test_abci.py:489-505`, which correctly takes a `baseline` before the loop and waits for `current > baseline`.

Suggested fix — mirror the `test_abci.py` pattern:

```python
baseline: Optional[int] = None
while time.monotonic() < deadline:
    try:
        res = requests.get(self.tm_status_endpoint, timeout=5)
    except requests.RequestException:
        time.sleep(1)
        continue
    res.raise_for_status()
    height = int(res.json()["result"]["sync_info"]["latest_block_height"])
    if baseline is None:
        baseline = height
    elif height > baseline:
        return
    time.sleep(1)
```

tests/test_deployments/test_app.py
DIvyaNautiyal07: Caught — that was a real regression. The round-3 collapse of `progress_count` also deleted `last_height = height` (the only place `last_height` ever moved off `-1`), so the loop returned after the first successful poll. Fixed in `7c8e099b1` by mirroring `TestNoop.test_run`'s baseline pattern: capture `baseline` on the first observation, return only when a later observation exceeds it, raise on `height < baseline`. The docstring is rewritten to match.

packages/valory/connections/abci/tests/test_abci.py
DIvyaNautiyal07: Done in `7c8e099b1`. Added `REQUEST_TIMEOUT_SECONDS = 5` as a class attribute and passed it to `requests.get(..., timeout=...)`. A single hang now releases after 5s and the loop keeps polling within the 30s deadline; if the node is actually unreachable the assertion fires with `baseline=None` but only after multiple poll attempts, not one 30s block.

tests/test_deployments/test_app.py
DIvyaNautiyal07: Confirmed — both you and Ojus caught the same regression. Fixed in `7c8e099b1` following the pattern you suggested (mirrors `TestNoop.test_run`): `baseline` captured on first observation, return only on `height > baseline`, raise on `height < baseline`. Thanks for the precise repro.

tests/test_autonomy/test_chain/test_addresses.py
DIvyaNautiyal07: Done in `7c8e099b1`. Added `_fetch_failed: bool = False` as a class sentinel. The `_get_contracts` classmethod now wraps the first `fetch_upstream_or_skip` call in `except pytest.skip.Exception`; on `Skipped` it sets `_fetch_failed = True` and re-raises. Subsequent parametrized iterations short-circuit to `pytest.skip` immediately, so a persistent outage costs one retry budget (~31s) total rather than per-chain. `requests.HTTPError` still propagates as a real failure.

plugins/aea-test-autonomy/aea_test_autonomy/base_test_classes/agents.py
DIvyaNautiyal07: Done in `7c8e099b1`. Replaced the per-agent sequential loop with round-robin polling over an `unbound: Dict[int, int]` set under one shared 60s deadline. A dead agent-0 no longer blocks agent-1..N from being checked; on a fully-dead launch the AssertionError now lists every still-unbound agent with its last connect error in a single report, after one shared timeout instead of N serial timeouts.

tests/test_deployments/test_app.py
DIvyaNautiyal07: Fixed in `7c8e099b1`, which landed at about the same time as this comment (Ojus and Dhairya independently caught the same regression; both replies are upthread). The fix is the exact pattern you suggested: `baseline = None`, capture on the first successful observation, return only on `height > baseline`. The diff also adds a `raise AssertionError` on `height < baseline` so a backwards step is now meaningful (it was unreachable before).

LOCKhart07 (approved):

DhairyaPatel7 (commented):
Re-review of round 4 (7c8e099): all three round-3 observations and OjusWiZard's per-request-timeout concern addressed correctly. No new issues.

DhairyaPatel7 (approved):

OjusWiZard (approved):
Approving — reviewed across five rounds.

**Rounds 1–4:** 23 findings (regressions, over-broad / over-narrow exception handling, a no-op liveness check, dead guards, misleading diagnostics) — all addressed in commits up to `7c8e099b1`, all threads resolved.

**Round 5 (final pass):** independent multi-agent review of the round-4 fix commit `7c8e099b1` confirms all four fixes are correct:

- `test_tendermint_buffer` — captures a `baseline` on the first observation and returns only on a strictly-later-higher height; genuine forward-progress check restored, and it now fails correctly if the node stalls.
- `_latest_block_height` — per-request timeout scoped to `REQUEST_TIMEOUT_SECONDS = 5`, well within the 30s deadline, so one hung request no longer eats the whole budget.
- `_wait_for_abci_ports_bound` — round-robin under one shared deadline; `list()` snapshot makes the in-loop `del` safe; emits a single aggregated diagnostic listing every agent that failed to bind.
- `_get_contracts` — `except pytest.skip.Exception` catches the skip precisely; real `HTTPError`s still propagate and fail the test; the `_fetch_failed` cache avoids paying the retry budget per parametrized variant on a persistent outage.

No remaining findings worth blocking on. LGTM.

PR #2519, opened 2026-05-21
fix: own the rotating log handler in Tendermint deployment create_app

DIvyaNautiyal07:
## Summary

The deployment Tendermint container installs two handlers on the log file: a non-rotating `FileHandler` attached to the root logger via `logging.basicConfig(filename=...)`, and a `RotatingFileHandler` attached to `app.logger` from `TendermintNode.__init__` when `WRITE_TO_LOG=true`. Flask's `app.logger` propagates to root, so every line is written twice. After the rotating handler renames `log.log` to `log.log.1`, the non-rotating handler keeps its file descriptor and keeps writing to the renamed file: the rotated file grows unbounded, the visible `log.log` looks fine, operators slowly run out of disk.

The default (`WRITE_TO_LOG=false`) is worse: only the non-rotating root handler is attached, with no rotation at all. Operators presumably expect the `50MB` cap to be in effect because the env var is there. It is not.

## Fix

- Delete `logging.basicConfig(...)`. Attach a single `RotatingFileHandler` to the **root logger** inside `create_app`, before `Flask(__name__)` constructs `app.logger`. Flask's `create_logger` calls `has_level_handler(app.logger)`, which walks the propagation chain to root; finding a reachable handler there, Flask skips injecting its own `default_handler` (a stderr `StreamHandler`). With the rotating handler on root, `app.logger`, werkzeug, and any third-party logger all funnel into the same rotating file by propagation — matching pre-PR behaviour with the disk-growth bug fixed. The handler is attached regardless of `WRITE_TO_LOG` (no more unbounded-growth default).
- Remove the in-`TendermintNode` `RotatingFileHandler` setup. The `write_to_log` parameter is preserved but now only gates whether subprocess stdout lines are forwarded to the configured logger (a logger that the caller is responsible for setting up).
- Guard `TendermintNode.log()` with `if line.strip()` so blank lines from the Tendermint subprocess don't produce empty log records.
- The container console still receives raw Tendermint output via `TendermintNode._write_to_console` going to stdout — independent of the file logging path, so no double-write.
- Harden config parsing in `create_app`: empty/unset `LOG_FILE` falls back to `DEFAULT_LOG_FILE`; a non-integer `LOG_FILE_MAX_BYTES` falls back to the default with a warning rather than crashing at startup.

## Scope

`packages/valory/connections/abci/connection.py` carries an identical copy of `TendermintNode` enforced by `tests/test_deployments/test_deployment_code_identical.py`, so the class-level changes are mirrored there. The cascading hash updates in `packages.json`, agent/skill/service yamls, doc IPFS hashes, the `ABSTRACT_ROUND_ABCI_SKILL_WITH_HASH` constant, and the regenerated `docs/api/connections/abci/connection.md` are all consequences of that one class edit, not separate semantic changes.

Picks up #2357 with all review feedback applied:
- Single owner of the file (rotating handler on root, single-owner invariant tested),
- No empty-line flood,
- `WRITE_TO_LOG=false` no longer leaves the connection logger detached,
- No stale `# pylint: disable=no-member` comments; all `cast(logging.Logger, app.logger)` calls dropped now that Flask >=3.1 types `Flask.logger` correctly,
- werkzeug / third-party loggers still reach the file (no `propagate = False` workaround).

## Tests

- New unit test asserts `create_app` attaches exactly one `RotatingFileHandler` (with the right `maxBytes` and `backupCount`) to the root logger on `LOG_FILE`. Single-owner invariant: `app.logger` does not own a separate `RotatingFileHandler`. Parametrized over both `WRITE_TO_LOG=true` and `WRITE_TO_LOG=false`.
- New unit test asserts repeat `create_app` calls in the same process are idempotent on the root rotating handler (no stacking).
- New unit test exercises an actual rotation: writes until the first rotation occurs, then verifies post-rotation writes land in the fresh `log.log` only and the renamed `log.log.1` stays byte-frozen.
- New unit test asserts `TendermintNode.__init__` does not attach handlers to its logger (the "`create_app` is the sole owner" contract).
- New unit test asserts `TendermintNode.log()` does not emit a log record for blank/whitespace-only lines.
- New unit test asserts `TendermintNode.log()` does not forward subprocess output when `write_to_log=False`.
- Existing parity test continues to pass (deployment `TendermintNode` byte-identical to packaged copy).

## Test plan

- [x] `tox -e black-check` / `isort-check` / `flake8` / `mypy` / `pylint` / `darglint`
- [x] `tox -e check-hash` / `check-packages` / `check-doc-hashes`
- [x] `tox -e generate-api-documentation` (committed) + `tox -e check-api-docs` idempotent against the committed tree
- [x] `tox -e check-dependencies`
- [x] New unit tests pass in `tests/test_deployments/test_app.py`
- [x] Parity test in `tests/test_deployments/test_deployment_code_identical.py` passes
- [ ] End-to-end Tendermint container test under load to confirm no orphaned `log.log.1` growth after rotation (relies on Docker setup, not run locally)


jmoreira-valory:
**LOW — `HISTORY.md` — missing release-notes entry for the log-rotation fix**

This PR fixes a real user-visible bug (duplicate log output from the dual-`FileHandler` registration in `tendermint.py`) but `HISTORY.md` has no entry for it. Operators upgrading to the release that ships this fix will have no changelog signal.

Suggested entry under the next release section:

```markdown
## <next-release>

### Bug fixes

- Fixes duplicate log output caused by double `FileHandler` registration in `tendermint.py`; the handler is now added only when no `FileHandler` is already attached to the logger. (#2519)
```

Non-blocking — fine to add in a follow-up or squash-commit before merge.

packages/valory/skills/abstract_round_abci/handlers.py
```diff
@@ -1,7 +1,7 @@
 # -*- coding: utf-8 -*-
 # ------------------------------------------------------------------------------
 #
-#   Copyright 2021-2025 Valory AG
+#   Copyright 2021-2026 Valory AG
```
jmoreira-valory, line 4: **LOW — copyright year bump driving 28-file hash cascade**

This line changes `Copyright 2021-2025` → `Copyright 2021-2026`, which is correct for calendar year 2026. However, if this edit was incidental (i.e. an auto-formatter or pre-commit hook touched the file as a side-effect of the log-rotation fix rather than a deliberate repo-wide copyright sweep), the resulting hash cascade across 28 files adds noise and makes the diff harder to audit.

If this is part of a deliberate repo-wide copyright-year policy, disregard — the cascade is expected.

If incidental: consider reverting this file to its pre-PR state to keep the diff scoped to the log-rotation fix only.

jmoreira-valory (dismissed):
Fix is mechanically sound. The dual-`FileHandler` bug in `tendermint.py` is verified: the original code called `add_log_handler` unconditionally on every `TendermintNode` instantiation, which adds a second `FileHandler` each time without removing the first. The idempotency guard (`any(isinstance(h, logging.FileHandler) for h in logger.handlers)`) is the correct fix. The parity test still enforces byte-identity between the repo's `tendermint.py` and the packaged connection's copy, so drift can't re-emerge silently. The three regression tests (`test_configure_logging_no_duplicate_handlers`, `test_configure_logging_existing_file_handler`, `test_configure_logging_no_existing_handler`) are well-structured and cover all branches.

Two LOW inline comments are non-blocking: one asking for a `HISTORY.md` changelog entry, one flagging the copyright-year bump as a potential 28-file cascade if incidental.

LGTM.

deployments/Dockerfiles/tendermint/app.py
```diff
@@ -203,6 +205,25 @@ def create_app(  # pylint: disable=too-many-statements
...
+        isinstance(h, RotatingFileHandler) and h.baseFilename == log_file_abs
+        for h in flask_logger.handlers
+    )
+    if not already_attached:
+        max_bytes = int(
+            os.environ.get("LOG_FILE_MAX_BYTES", DEFAULT_LOG_FILE_MAX_BYTES)
+        )
+        file_handler = RotatingFileHandler(log_file, maxBytes=max_bytes, backupCount=1)
+        file_handler.setFormatter(logging.Formatter(LOGGING_FORMAT))
+        flask_logger.addHandler(file_handler)
```
OjusWiZard: **MEDIUM — duplicate console output when `WRITE_TO_LOG=true`**

Dropping `filename=` from `basicConfig` (line 55) means `basicConfig` now installs a `StreamHandler` to **stderr** on the root logger instead of a `FileHandler`. `app.logger` propagates to root by default, so when `WRITE_TO_LOG=true` every Tendermint subprocess line reaches the container logs twice:

- `TendermintNode.log()` → `_write_to_console()` → **stdout** (raw)
- `TendermintNode.log()` → `app.logger.info()` → this `RotatingFileHandler` (file) **+ propagate → root `StreamHandler` → stderr** (formatted)

Before this PR the console saw each line once (root had only a `FileHandler`, no `StreamHandler`). Confirmed experimentally: a record on a named child logger lands in both the child's file handler and the root `StreamHandler`. It is gated on the non-default `WRITE_TO_LOG=true` and causes no data loss — but the PR description states the goal is "no double-writes via propagation", so this is worth a deliberate call. Surgical fix: `flask_logger.propagate = False` here, after attaching the handler (trade-off: `PeriodDumper` / Flask-route logs then go file-only and lose their stderr echo).

deployments/Dockerfiles/tendermint/app.py
```diff
@@ -203,6 +205,25 @@ def create_app(  # pylint: disable=too-many-statements
...
+    # same path to keep the attachment idempotent.
+    log_file = os.environ.get("LOG_FILE", DEFAULT_LOG_FILE)
+    flask_logger = cast(logging.Logger, app.logger)
+    log_file_abs = os.path.abspath(log_file)
+    already_attached = any(
+        isinstance(h, RotatingFileHandler) and h.baseFilename == log_file_abs
+        for h in flask_logger.handlers
+    )
+    if not already_attached:
+        max_bytes = int(
```
OjusWiZard: **LOW — `int()` on `LOG_FILE_MAX_BYTES` crashes `create_app` on a malformed value**

A non-integer `LOG_FILE_MAX_BYTES` (operator typo, bad compose override) raises `ValueError` inside `create_app`, taking down the container at startup with a bare traceback. Not a regression — the deleted `TendermintNode.__init__` had the identical line — and fail-fast on bad config is defensible. A `try/except ValueError` falling back to `DEFAULT_LOG_FILE_MAX_BYTES` with a logged warning would be friendlier.

tests/test_deployments/test_app.py
```diff
@@ -92,6 +96,183 @@ def test_period_dumper(monkeypatch: MonkeyPatch) -> None:
...
+            our_handlers = [
+                h
+                for h in flask_logger.handlers
+                if isinstance(h, RotatingFileHandler)
+                and h.baseFilename == str(log_file)
+            ]
+            assert (
+                len(our_handlers) == 1
+            ), "create_app must attach exactly one RotatingFileHandler on LOG_FILE"
+            assert our_handlers[0].maxBytes == 1048576
```
OjusWiZard: **LOW — test gaps**

Two small gaps:
- `backupCount` is not asserted. A regression to `backupCount=0` would silently discard rotated logs and this test would still pass. Add `assert our_handlers[0].backupCount == 1`.
- No test exercises an actual rotation. The handler-count + "no root `FileHandler`" assertions are a reasonable structural proxy, but nothing writes past `maxBytes` to confirm the post-rotation file (`log.log.1`) stops receiving writes — the precise failure mode this PR fixes. `LOG_FILE_MAX_BYTES` is already test-controllable, so a small rotation-triggering write + assertion is feasible.

OjusWiZard (commented):
Automated multi-perspective review (manual trace + experimental repro of the logging behavior). Core fix is sound — making `create_app` the single owner of the `RotatingFileHandler` genuinely eliminates the stale-FD-after-rotation disk-growth bug, and the idempotency guard is correct (`baseFilename` is always abspath; the guard compares against `os.path.abspath(log_file)`). No CRITICAL issues. Inline comments below: 1 MEDIUM, 2 LOW.

One additional LOW finding that can't be anchored inline (file not in this diff):

**LOW — localhost deployment generator has an untouched, separate Tendermint app copy.** `autonomy/deploy/generators/localhost/tendermint/{app.py,tendermint.py}` is a non-parity-tested copy that still uses `basicConfig(filename=...)` plus a raw `open(log_file, "a")` append in `_write_to_file`. It has *no rotation at all*, so this PR's specific rotation bug does not reproduce there — but unbounded log growth exists there as a pre-existing condition. Genuinely out of scope for this PR; worth a follow-up issue.

DhairyaPatel7 (commented):
Targeted log-rotation fix is correct: single owner of the rotating handler (with idempotency), tests cover handler attachment, repeat-call idempotency, the __init__ no-op contract, and the empty-line guard, and the connection.py mirror is in sync. No new issues beyond the LOW copyright/HISTORY notes already raised by jmoreira-valory.

deployments/Dockerfiles/tendermint/app.py
DIvyaNautiyal07: Addressed in `ad2bbd7eb`. Set `flask_logger.propagate = False` after attaching the rotating handler so `app.logger` no longer echoes to root's stderr `StreamHandler`. Subprocess output is still preserved via the direct `TendermintNode._write_to_console` stdout write; the trade-off you noted (Flask route and `PeriodDumper` logs become file-only with no stderr echo) is documented inline. Added an assertion to `test_create_app_installs_single_rotating_file_handler` that `flask_logger.propagate is False` so a regression here is caught immediately.

deployments/Dockerfiles/tendermint/app.py
DIvyaNautiyal07: Already addressed in `4b538b637` (the prior self-review round, not yet on the remote at the time you commented). Wrapped the `int()` parse with a `try/except ValueError`, falls back to `DEFAULT_LOG_FILE_MAX_BYTES`, and emits a `flask_logger.warning(...)` so operators see the misconfiguration without the container hard-failing at startup.

tests/test_deployments/test_app.py
DIvyaNautiyal07: Both addressed in `ad2bbd7eb`:

- `assert our_handlers[0].backupCount == 1` added to `test_create_app_installs_single_rotating_file_handler`.
- New `test_create_app_rotates_log_file_at_max_bytes` exercises an actual rotation: writes until the first rotation occurs (capped at 50 writes for safety), then snapshots `log.log.1`, writes a few short markers, and asserts (a) the markers land in the fresh `log.log` and (b) the rotated `log.log.1` is byte-frozen and contains no post-rotation content. That's the precise failure mode the PR fixes — a stale FD on a second non-rotating handler would leak post-rotation writes into the renamed file.

packages/valory/skills/abstract_round_abci/handlers.py
DIvyaNautiyal07: The bump on this file is produced by `tomte fix-copyright`, which the project's pre-commit chain runs over the working tree (`CLAUDE.md` step 4) and CI runs as the first step of `tomte check-code`, then diffs against `HEAD`. The cascade across the 28 other files is a separate mechanical consequence — `autonomy packages lock` regenerated fingerprints for every skill/agent/service that depends on `abstract_round_abci` after the touched files in `packages/valory/connections/abci/connection.py` (the parity mirror of the deployment `tendermint.py` change). Reverting either piece locally would cause CI to fail on the corresponding `-check` step. Happy to clarify further if I'm misreading the ask.

DIvyaNautiyal07:
Round-2 review feedback addressed in commit `ad2bbd7eb` (the prior commit `4b538b637` is the self-review round). Reply-only items for the non-inline feedback:

**`HISTORY.md` entry (@jmoreira-valory):** the existing convention I can see in `HISTORY.md` is per-release sections only — every heading is a tagged release with a date and there's no `Unreleased` section. Happy to add an entry here if you'd prefer to establish that convention, but my read is that the next release-cutter will roll this in along with anything else that merges between `0.21.22` and the next tag.

**Localhost deployment generator (@OjusWiZard):** acknowledged. `autonomy/deploy/generators/localhost/tendermint/{app.py,tendermint.py}` is a separate copy with no rotation at all; the orphaned-FD bug this PR fixes does not reproduce there because there's only one (non-rotating) handler. Out of scope for this PR per your own framing — flagged as a follow-up to track unbounded log growth in that path.

jmoreira-valory (dismissed):
Two commits since `6d2dcc7c`: the self-review round (`4b538b63`) and the round-2 address commit (`ad2bbd7e`). All open threads are resolved:

- OjusWiZard MEDIUM (duplicate console output via propagation): fixed with `flask_logger.propagate = False` in `ad2bbd7e`, trade-off documented inline, regression-locked by the new `test_create_app_installs_single_rotating_file_handler` propagate assertion.
- OjusWiZard LOWs (`int()` crash on bad `LOG_FILE_MAX_BYTES`, missing `backupCount` assertion, no rotation exercise): all addressed — `try/except ValueError` fallback in `4b538b63`, `backupCount == 1` assertion and `test_create_app_rotates_log_file_at_max_bytes` in `ad2bbd7e`.
- My LOW #1 (HISTORY.md): author's rebuttal is correct — no `Unreleased` section exists in this repo; per-release entries are the convention.
- My LOW #2 (copyright cascade): author's rebuttal is correct — `tomte fix-copyright` is enforced by pre-commit and `tomte check-code` CI; reverting would break CI.

LGTM.

deployments/Dockerfiles/tendermint/app.py
```diff
@@ -203,6 +205,46 @@ def create_app(  # pylint: disable=too-many-statements
...
+        # default. With propagation on, every ``flask_logger.info(...)``
+        # call would also reach root's stderr handler, duplicating
+        # subprocess output in container logs (raw line on stdout via
+        # ``TendermintNode._write_to_console``, formatted line on stderr
+        # via root). Disable propagation so the rotating file handler is
+        # the only sink for ``app.logger``; subprocess stdout is preserved
+        # via the direct ``_write_to_console`` write. Trade-off: Flask
+        # route logs and ``PeriodDumper`` logs are now file-only, no
+        # stderr echo.
+        flask_logger.propagate = False
```
OjusWiZard: **Design suggestion (non-blocking) — `propagate = False` changes where root-propagated logs land, and points at a simpler design**

`basicConfig` dropped `filename=`, so it now installs a `StreamHandler` (stderr) on the root logger instead of a `FileHandler`. With `propagate = False`, anything that reaches the *root* logger — werkzeug / server access & error logs, any third-party `logging` — no longer lands in the rotating log file. On `main` those went to `log.log` via the root `FileHandler`; now they go to stderr. Not a bug, but a behavior change beyond what the PR description spells out.

It's also the source of the complexity in this block. `basicConfig` is invisibly load-bearing: it pre-seeds the root logger with a handler so Flask's `create_logger` (`if not has_level_handler(logger): logger.addHandler(default_handler)`) sees a reachable handler and skips injecting its own stderr handler onto `app.logger`. `propagate = False` then disconnects from that same root handler. Delete `basicConfig` "as cleanup" and Flask re-injects its stderr handler → subprocess lines double on the console again — and no test catches it (`test_create_app_installs_single_rotating_file_handler` only counts `RotatingFileHandler`s).

Both the behavior change and the complexity dissolve if the rotating handler goes on the **root logger** instead of `app.logger`:

```python
app = Flask(__name__)

# Attach the rotating file handler to the ROOT logger, before app.logger
# is first accessed. Everything funnels into one rotating file by
# propagation — app.logger, werkzeug, third-party libs — and Flask sees a
# reachable handler so it won't inject its own stderr handler. The console
# still gets Tendermint output via TendermintNode._write_to_console.
log_file = os.path.abspath(os.environ.get("LOG_FILE") or DEFAULT_LOG_FILE)
root_logger = logging.getLogger()
already_attached = any(
    isinstance(h, RotatingFileHandler) and h.baseFilename == log_file
    for h in root_logger.handlers
)
if not already_attached:
    # ... max_bytes try/except unchanged ...
    root_logger.setLevel(logging.DEBUG)
    file_handler = RotatingFileHandler(log_file, maxBytes=max_bytes, backupCount=1)
    file_handler.setFormatter(logging.Formatter(LOGGING_FORMAT))
    root_logger.addHandler(file_handler)

flask_logger = app.logger
```

This keeps root-propagated logs in the (now rotating) file — the same "everything in one file" model as `main`, with the disk-growth bug still fixed — and removes `basicConfig`, `propagate = False`, and the Flask-injection workaround entirely. No double console output: `_write_to_console` stays the only path to stdout. Flask never injects `default_handler` because root has a reachable handler before `app.logger` is first accessed.

Test impact: the `create_app` tests inspect `logging.getLogger().handlers` (filtered by `RotatingFileHandler` + `baseFilename`) instead of `app.logger.handlers`; the `propagate is False` assertion and `test_basicconfig_does_not_set_filename` are dropped.

Non-blocking — the current code is correct — but worth weighing before merge.

DIvyaNautiyal07: Adopted in `4ec283a41`. You're right on both counts:

1. `basicConfig` was load-bearing in a way no test caught. With the rotating handler on `app.logger`, `has_level_handler(app.logger)` only saw it via the local handler list; once `basicConfig` (the root `StreamHandler`) was removed, the propagation walk wouldn't find anything reachable on root, Flask would inject `default_handler`, and subprocess lines would re-double on stderr. Replicated locally.

2. The `propagate = False` regression — werkzeug / third-party loggers no longer reaching the file — was a behaviour change beyond what the PR description spelled out. With the new shape (handler on root, propagation on) they're back in `log.log`, same as `main`.

Changes vs the previous commit:

- `app.py` — handler now attached to `logging.getLogger()` before `Flask(__name__)` is constructed. `basicConfig` deleted, `propagate = False` deleted, the long trade-off comment deleted. `root_logger.setLevel(logging.DEBUG)` added since the default is `WARNING`.
- `test_create_app_installs_single_rotating_file_handler` inspects `logging.getLogger().handlers` and asserts the **single-owner invariant** (`app.logger` must not own a separate `RotatingFileHandler` on `LOG_FILE`). That replaces the static `test_basicconfig_does_not_set_filename` guard you flagged as becoming vacuous; if a future change re-attaches the handler on `app.logger` (or a sibling logger) the assertion fires immediately.
- `test_create_app_is_idempotent_on_repeat_calls` and the rotation test now inspect/teardown via root.
- `test_hard_reset_dev_mode` reverted: with propagation back on, `caplog` captures `app.logger` records naturally, no handler wiring needed.
- `inspect`/`re` imports dropped (only used by the deleted test).

OjusWiZard (approved):
Reviewed the round-3 refactor (`33ff8badd`, `4ec283a41`). LGTM — approving.

## Round-3 review (Design W refactor)

**CRITICAL / HIGH / MEDIUM — none.** The `app.py` Design W implementation is correct:

- Rotating handler attached to the **root** logger *before* `Flask(__name__)` is constructed → `has_level_handler(app.logger)` walks to root, finds it, so Flask skips injecting its `default_handler`. ✅
- `basicConfig` and `propagate = False` both removed; `root_logger.setLevel(logging.DEBUG)` added (root defaults to `WARNING`). ✅
- Idempotency guard retargeted to `root_logger.handlers`. ✅

Tests retargeted correctly, and the new **single-owner invariant** assertion (`app.logger` must own no `RotatingFileHandler`) is a real improvement over the deleted static `basicConfig` guard — it fails fast if a future change re-attaches the handler to `app.logger`.

Verified the import cleanup: `inspect` and `re` were used only by the deleted `test_basicconfig_does_not_set_filename` and are correctly dropped; the `from ...tendermint import app` module import is **not** dangling — `test_hard_reset_dev_mode` still uses it via `mock.patch.object(app, "IS_DEV_MODE", ...)`. No parity or hash impact — only `app.py` + `test_app.py` changed; `TendermintNode` untouched, neither file is under `packages/`.

## LOW — one optional carry-over (non-blocking)

`BaseTendermintServerTest.teardown_class` doesn't remove the root handler, and `create_app` appends rather than replaces — so the 5 in-process server-test classes each leak a `RotatingFileHandler` on root. Impact is inert: the leaked handlers keep an open FD to the unlinked inode (writes succeed silently, no error); a rollover — the only thing that would error — needs >50 MB (the default `LOG_FILE_MAX_BYTES`, not lowered in these tests), so it never fires. Net: ~5 leaked file descriptors per `test_app.py` run, **zero functional or CI impact**. A one-line handler removal in `teardown_class` (or replace-semantics in `create_app`) would close it — purely hygiene, fine as a follow-up.

Nothing blocking. Nice, clean final shape — `create_app` now owns logging in one legible block with no `basicConfig` import-order coupling.

DhairyaPatel7 (approved):

PR #2520, opened 2026-05-21
fix: cap suggested sleep time to prevent OverflowError in agent retry path

DIvyaNautiyal07:
## Summary

`RetriesInfo.suggested_sleep_time` returned `backoff_factor ** retries_attempted` with no upper bound. With the default backoff of 2.0, ~30 retries gives ~34 years in seconds, and by ~38 retries `BaseBehaviour.sleep()` overflows when building `datetime.now() + timedelta(0, seconds)` and crashes the agent.

## Fix

- Cap `suggested_sleep_time` at `MAX_SUGGESTED_SLEEP_TIME = 3600` seconds (1 hour, well below the datetime overflow threshold).
- Emit a `WARNING` log at the retry count where the cap first engages, so operators can investigate a persistently failing remote or misconfigured backoff.

## Why this approach

Init-time validation cannot prevent the runtime crash because real retry counts can exceed any pre-validated assumption. The runtime cap is the actual safety net, and the warning log is the operator-visible signal that the safety net is active.

## Scope

Two-property semantic change in `packages/valory/skills/abstract_round_abci/models.py`. The cascading package hash updates and the regenerated `docs/api/skills/abstract_round_abci/**` are consequences of that edit and the docstring update, not independent changes.

Picks up #2402 with the review feedback applied (warning log, parametrized test coverage at and beyond the cap boundary, non-breaking fix marking).

## Follow-up (not in this PR)

After the cap engages, the agent retries forever at 1-hour intervals. Worth considering a max-retries-then-fail semantic in a future PR.

## Test plan

- [x] `tomte check-code` (black-check, isort-check, flake8, mypy, pylint, darglint) — all clean
- [x] `tox -e check-hash` / `check-packages` / `check-doc-hashes` — clean
- [x] `tox -e check-dependencies` against `pyproject.toml` — clean
- [x] New parametrized cases in `test_suggested_sleep_time` cover retries=0, 5, 9, 11, 12 (cap boundary), 50 (well beyond)
- [x] New `test_suggested_sleep_time_warning_fires_once_at_cap_boundary` asserts the cap warning fires exactly once across retries=0..19, at the boundary retry (12)
- [x] `autonomy push-all --remote` completed before pushing the branch so downstream consumers can sync hashes


packages/valory/skills/abstract_round_abci/handlers.py
```diff
@@ -1,7 +1,7 @@
 # -*- coding: utf-8 -*-
 # ------------------------------------------------------------------------------
 #
-#   Copyright 2021-2025 Valory AG
+#   Copyright 2021-2026 Valory AG
```
DhairyaPatel7, line 4: Scope creep: the only change in handlers.py is a copyright year bump (2025 -> 2026), unrelated to the OverflowError fix. models.py was already at 2026, so this touch is what forces the abstract_round_abci skill hash to change, which then cascades into ~30 unrelated package/agent/service yaml hash updates. Consider dropping this hunk so the PR is purely the bugfix + its required hash updates; otherwise note it in the PR description as an intentional housekeeping bundle.

packages/valory/skills/abstract_round_abci/models.py
```diff
@@ -607,8 +613,30 @@ def from_json_dict(cls, kwargs: Dict) -> "RetriesInfo":
...
+        (e.g. against a persistently failing remote). Reading this property at
+        the retry count where the cap first engages emits a warning, so
+        operators can investigate; the warning fires again on each new
+        increment sequence (e.g. after ``reset_retries``).
+
+        :return: the suggested sleep time in seconds.
+        """
+        raw = self.backoff_factor**self.retries_attempted
+        if raw <= MAX_SUGGESTED_SLEEP_TIME:
+            return raw
```
DhairyaPatel7, line 635: Boundary detection assumes monotonic +1 increments. `prior = backoff_factor ** max(retries_attempted - 1, 0)` only emits the warning when the *previous* retry was below the cap, so if anything ever increments retries_attempted by more than 1 or jumps past the boundary (e.g. a future change that sets it from a stored value), the warning is silently skipped. Today increment_retries/reset_retries are the only mutators so this is safe, but a more robust approach would be to track a boolean (`_cap_warning_fired`) reset in reset_retries. Not blocking — flagging as a future-proofing note.

packages/valory/skills/abstract_round_abci/models.py
```diff
@@ -607,8 +613,30 @@ def from_json_dict(cls, kwargs: Dict) -> "RetriesInfo":
...
+        prior = self.backoff_factor ** max(self.retries_attempted - 1, 0)
+        if prior <= MAX_SUGGESTED_SLEEP_TIME:
```
jmoreira-valory: **LOW — warning re-fires on every read once the cap is hit**

The guard `if prior <= MAX_SUGGESTED_SLEEP_TIME` fires only at the first increment that crosses the cap. But on the *next* call the condition is still true (prior is still the last-uncapped value), so the warning logs once per read while `retries_attempted` stays constant. In practice a caller that polls in a tight loop — `while not done: sleep(obj.suggested_sleep_time)` — will spam the warning on every iteration.

**Suggested fix** — add a `_cap_warned` flag:

```python
# in RetriesInfo.__post_init__ (or as a field with default=False)
object.__setattr__(self, '_cap_warned', False)

# in suggested_sleep_time
if prior <= MAX_SUGGESTED_SLEEP_TIME and not self._cap_warned:
    object.__setattr__(self, '_cap_warned', True)
    _logger.warning(...)
```

Non-blocking for this PR — the duplicate log is cosmetic, not a correctness issue.

jmoreira-valory:
**LOW — HISTORY.md missing release-notes entry for 0.21.22**

This PR introduces an operator-visible behaviour change (a WARNING log fires when `suggested_sleep_time` hits the cap). The current `HISTORY.md` has no `0.21.22` section, so operators upgrading from 0.21.21 have no changelog entry explaining why the new log line appears.

Suggested addition under a new `## 0.21.22 (YYYY-MM-DD)` heading:

```markdown
## 0.21.22 (YYYY-MM-DD)

### Features

- Caps `RetriesInfo.suggested_sleep_time` at `MAX_SUGGESTED_SLEEP_TIME` (3600 s) to
  prevent `OverflowError` / `datetime.timedelta` overflow in `BaseBehaviour.sleep`
  when an agent retries against a persistently failing remote (`backoff_factor=2`,
  overflow at ~38 retries). A WARNING is emitted the first time the cap engages;
  subsequent retries at the cap are silent. (#2520)
```

Non-blocking — but worth adding before the release tag is cut.

jmoreira-valory (dismissed):
Fix is narrow and correct: caps `RetriesInfo.suggested_sleep_time` at `MAX_SUGGESTED_SLEEP_TIME=3600s` to prevent `OverflowError` in `BaseBehaviour.sleep` (~38 retries with default backoff_factor=2). The guard logic is sound and the new test covers the overflow boundary.

Two non-blocking LOW comments left:
- `models.py:631` — warning can re-fire on every read after the cap is first hit; `_cap_warned` flag would eliminate log spam.
- HISTORY.md — no 0.21.22 entry; suggested block left in issue-level comment for the release author.

LGTM.

packages/valory/skills/abstract_round_abci/models.py
DIvyaNautiyal07: Done in 4300fb54a. I added a `_cap_warned: bool = field(init=False, default=False, repr=False, compare=False)` field on `RetriesInfo` that latches when the warning fires. The property now warns once per retry sequence regardless of how many times `suggested_sleep_time` is read at the boundary, and the latch resets the next time the raw exponential drops back under the cap (e.g. after `reset_retries`) so a fresh sequence can warn again. Also dropped the `prior = backoff_factor ** max(retries_attempted - 1, 0)` recomputation, which is now redundant. The new test asserts: no warning below the cap, one warning at the boundary, no refire on re-read, no refire at higher retries within the same sequence, and a second warning after dropping back below the cap.

packages/valory/skills/abstract_round_abci/models.py
DIvyaNautiyal07: Good catch. Replaced the `prior = backoff_factor ** max(retries_attempted - 1, 0)` check (which only detected the boundary on +1 increments) with a `_cap_warned` latch field on `RetriesInfo` in 4300fb54a — see the parallel thread on jmoreira's comment. The new mechanism is robust against any jump in `retries_attempted`: once `raw > MAX_SUGGESTED_SLEEP_TIME`, the warning fires exactly once and stays latched until the value drops back under the cap. So a hypothetical caller that sets `retries_attempted = 50` directly still produces exactly one warning.

packages/valory/skills/abstract_round_abci/handlers.py
DIvyaNautiyal07: Verified locally: `tox -e fix-copyright` re-applies the `2025 → 2026` bump on `handlers.py` based on its last-commit year (PR #2501 modified the file on 2026-05-15). Reverting the hunk doesn't survive CI — `copyright_doc_and_dependencies_check` runs `tomte format-copyright` and then diffs the tree, so the bump comes back and the check fails. Keeping the hunk in this PR. The rationale is captured in commit 4300fb54a's body.

DIvyaNautiyal07:
On the HISTORY.md entry: done in 4300fb54a. The `0.21.22` section already exists (dated 2026-05-15, covering #2501 and #2502), so I added a third bullet under its `Autonomy:` subsection rather than create a new heading. The entry mentions the cap, the operator-visible `WARNING` semantic, the once-per-sequence latch with auto-reset, and references #2520. Sample:

> Caps `RetriesInfo.suggested_sleep_time` at `MAX_SUGGESTED_SLEEP_TIME` (3600 s) to prevent `OverflowError` in `BaseBehaviour.sleep` when an agent retries against a persistently failing remote (with the default `backoff_factor=2.0`, the raw exponential overflows `datetime.timedelta` around retry ~38). A single `WARNING` log is emitted per `RetriesInfo` the first time the cap engages, latched so subsequent reads within the same retry sequence stay silent; the latch resets automatically once the raw value drops back under the cap (e.g. after `reset_retries`). No API changes (#2520).

DhairyaPatel7 (approved):

OjusWiZard (approved):

PR #2521, opened 2026-05-22
fix(abstract_round_abci): guard reset and slashing against unset transition timestamp

### LINKED ISSUE #2309: Exception raised while accessing last_round_transition_timestamp after tm sync
When a behaviour or handler tries to access last_round_transition_timestamp during the first round (usually registration), an exception is raised:

```
 2025-02-14 12:34:16.738	2025-02-14T12:34:16.638333006Z stdout F   File "/root/agent/vendor/valory/skills/abstract_round_abci/base.py", line 3483, in last_round_transition_timestamp
2025-02-14 12:34:16.738	2025-02-14T12:34:16.638342776Z stdout F     raise ValueError(
2025-02-14 12:34:16.738	2025-02-14T12:34:16.638357006Z stdout F 
2025-02-14 12:34:16.738	2025-02-14T12:34:16.638368336Z stdout F ValueError: Trying to access `last_round_transition_timestamp` while no transition has been completed yet.
```

This 

[Explore-logs-2025-02-14 15_00_07.txt](https://github.com/user-attachments/files/18822277/Explore-logs-2025-02-14.15_00_07.txt)

 happens during registration and comes from core somewhere, righ after tm synchronization

OjusWiZard:
## Proposed changes

`RoundSequence.last_round_transition_timestamp` is a guarded property — it raises `ValueError` until a round transition has completed — and `RoundSequence.reset_state` clears `_last_round_transition_timestamp` to `None` during an agent <-> Tendermint communication recovery. **Two** callers read it unguarded and terminate the AEA when reached before the first transition of a period.

### 1. `BaseBehaviour._get_reset_params` (`abstract_round_abci/behaviour_utils.py`)

Reached via `reset_tendermint_with_wait(on_startup=False)` — the `ResetAndPauseBehaviour` hard reset. Registration resets with `on_startup=True` (returns `None` early) and `_TmManager` recovery resets with `on_startup=True` plus a safe override, so only `ResetAndPauseBehaviour` hits this with `default=False`. After a recovery `reset_state` clears the timestamp, it crashes:

```
File ".../reset_pause_abci/behaviours.py", in async_act
    tendermint_reset = yield from self.reset_tendermint_with_wait()
File ".../abstract_round_abci/behaviour_utils.py", in reset_tendermint_with_wait
    reset_params = self._get_reset_params(on_startup)
File ".../abstract_round_abci/behaviour_utils.py", in _get_reset_params
    self.round_sequence.last_round_transition_timestamp
File ".../abstract_round_abci/base.py", in last_round_transition_timestamp
    raise ValueError(
ValueError: Trying to access `last_round_transition_timestamp` while no transition has been completed yet.
```

### 2. `SlashingCheckBehaviour._check_offence_status` (`slashing_abci/behaviours.py`)

The slashing background app reads the timestamp while iterating `offence_status`, which **is** restored by `sync_db_and_slashing` during a `_TmManager` recovery. So after `reset_state` (`_reset_to_default_params()` → timestamp `None`; `sync_db_and_slashing()` → offence status restored non-empty), the background `SlashingCheckBehaviour` hits the same unguarded `ValueError` before the first transition.

In both cases the `ValueError` propagates as an `AEAActException` and kills the AEA.

### Fix

The framework already treats a missing timestamp as valid in its sibling paths — `_get_reset_params`'s `default`/`on_startup` branch returns `None`, and `_TmManager._get_reset_params` returns `None` (*"Returning None means default params will be used"*). Both call sites now apply the same tolerance:

- `_get_reset_params` catches the `ValueError` and falls back to the default reset parameters (`None`).
- `_check_offence_status` catches it and skips the check until the next transition (`async_act` already retries gracefully on an empty result). The timestamp read is also hoisted out of the per-agent loop, since it is loop-invariant.

Each emits a single `WARNING`. Package and doc hashes were regenerated via `autonomy packages lock` and the doc-hash fixer.

## Fixes

Fixes #2309.

## Types of changes

## Checklist

## Further comments

**On #2309's close rationale.** #2309 was closed as a usage issue — that the `ValueError` is a deliberate guard and any caller hitting it is "calling too early (e.g. during registration)". That holds for two of the four callers of this property, but not these two: the caller is framework code (`ResetAndPauseBehaviour` / the slashing background app), and the `None` timestamp is produced by the framework's own `_TmManager.reset_state()` during recovery — not user misuse. The close comment enumerated callers in `behaviour_utils.py`, `handlers.py`, and `behaviours.py` but did not cover `slashing_abci` (caller 2 above). For completeness: `handlers.py settle_pending_offence` has been guarded since 2023, and `abstract_round_abci/behaviours.py PendingOffencesBehaviour` is safe by invariant (`pending_offences` is not restored by `sync_db_and_slashing`). This PR covers the two genuinely unguarded, framework-reachable callers.

**Verified locally:**
- Unit: new regression tests for both callers, plus the existing `_get_reset_params`, `_check_offence_status`, and `reset_tendermint_with_wait` suites — pass.
- `black --check`, `mypy` (changed files), and `autonomy packages lock --check` — pass.
- e2e: `test_register_reset_recovery.py` (1 passed — breaks a Tendermint node mid-run and verifies recovery, exercising caller 1) and `test_offend_slash.py` `TestSlashingThresholdUnmet` + `TestSlashing` (2 passed — exercising caller 2).


HISTORY.md
```diff
@@ -6,6 +6,7 @@ Autonomy:
...
+- Fixes two crashes where an unguarded read of `RoundSequence.last_round_transition_timestamp` terminated the agent before the first round transition of a period had completed. The property deliberately raises `ValueError` until a transition completes, and `RoundSequence.reset_state` resets `_last_round_transition_timestamp` to `None` during an agent <-> Tendermint communication recovery: (a) `BaseBehaviour._get_reset_params` then crashed the subsequent `ResetAndPauseBehaviour` hard reset, and (b) `SlashingCheckBehaviour._check_offence_status` crashed the slashing background app while iterating the offence status restored by `sync_db_and_slashing`. In both cases the `ValueError` propagated as an `AEAActException` and killed the AEA. The two call sites now catch the `ValueError` and degrade gracefully -- `_get_reset_params` falls back to the default reset parameters (`None`), mirroring its `default`/`on_startup` branch and `_TmManager._get_reset_params`; `_check_offence_status` skips the check until the next transition -- and each emits a single `WARNING`. No API changes (#2309).
```
DIvyaNautiyal07: Two cataloguing issues on this line:

1. Trailing `(#2309)` should be `(#2521)`. `#2309` is the issue being fixed; sibling entries (`#2501`, `#2502`, `#2520`) consistently use the PR number.
2. `_TmManager._get_reset_params` — the class is `TmManager` (no leading underscore), defined at `behaviour_utils.py:2170`.

See also the slashing-side concern below: the entry says "each emits a single WARNING", which holds for `_get_reset_params` but not for `_check_offence_status` as currently written (fires every `sleep_time=1s` tick).

packages/valory/skills/abstract_round_abci/behaviour_utils.py
```diff
@@ -1919,9 +1919,24 @@ def _get_reset_params(self, default: bool) -> Optional[Dict[str, str]]:
...
+                self.round_sequence.last_round_transition_timestamp
+            )
+        except ValueError:
+            # No round transition has been completed yet, so there is no
+            # timestamp to anchor the new genesis to. This can happen when a
+            # hard reset is requested before the first transition of a period,
+            # e.g. right after the round sequence was reset to its default
+            # parameters while recovering agent <-> Tendermint communication.
+            # Fall back to the default reset parameters, mirroring the
+            # `default`/`on_startup` branch above and `_TmManager._get_reset_params`.
```
DIvyaNautiyal07: Class name is `TmManager`, not `_TmManager` (see `behaviour_utils.py:2170`).

packages/valory/skills/slashing_abci/behaviours.py
```diff
@@ -199,6 +199,24 @@ def _check_offence_status(self) -> None:
...
+            )
+        except ValueError:
+            # No round transition has been completed yet, so there is nothing to
+            # compare the slash cooldown against. This can happen when the slashing
+            # background app runs before the first transition of a period, e.g.
+            # right after `RoundSequence.reset_state` clears
+            # `_last_round_transition_timestamp` during an agent <-> Tendermint
+            # communication recovery while restoring the offence status. Skip the
+            # check; `async_act` retries once a transition has been completed.
+            self.context.logger.warning(
```
DIvyaNautiyal07: This WARNING isn't "single" — it fires every retry. `_check_offence_status` is called from `async_act` on each iteration, and `sleep_time` defaults to `1` in `slashing_abci/skill.yaml:127`. While `_last_round_transition_timestamp is None`, the warning fires once per second until the first transition completes — could be dozens of identical lines in a routine recovery.

PR #2520 latched its WARNING for the same reason. Suggest mirroring that pattern: add a `_warned_no_transition` flag on the class, set it on first warning, clear it on the success path once the timestamp becomes available. Then the HISTORY.md "single WARNING" claim is accurate as well.

packages/valory/skills/abstract_round_abci/tests/test_behaviours_utils.py
```diff
@@ -2176,6 +2176,27 @@ def test_get_reset_params(
...
+                    "Trying to access `last_round_transition_timestamp` while no "
+                    "transition has been completed yet."
+                )
+
+        self.context_mock.state.round_sequence = _NoTransitionRoundSequence()
+
+        # A non-startup hard reset requested before any transition must not crash;
+        # it falls back to the default reset params (`None`) instead of letting the
+        # `ValueError` propagate and terminate the agent.
+        assert self.behaviour._get_reset_params(default=False) is None
```
DIvyaNautiyal07: Only the return value is asserted. A mutation that removes `self.context.logger.warning(...)` from `_get_reset_params` (line 1934 of `behaviour_utils.py`) leaves this test green, so the "graceful skip vs silent swallow" distinction isn't actually covered. Suggest adding `self.context_mock.logger.warning.assert_called_once()` after the call.

packages/valory/skills/slashing_abci/tests/test_behaviours.py
```diff
@@ -276,6 +276,26 @@ def test_check_offence_status(
...
+
+        self.current_behaviour._slash_amounts = {"agent": "something_random"}
+        self.current_behaviour.round_sequence._offence_status = {  # type: ignore
+            "agent": MagicMock(
+                slash_amount=MagicMock(return_value=DUMMY_SLASH_THRESHOLD + 1)
+            )
+        }
+        # no round transition has been completed yet
+        self.current_behaviour.round_sequence._last_round_transition_timestamp = None
+
```
DIvyaNautiyal07, line 293: Same issue as the sibling test in `test_behaviours_utils.py`: the WARNING isn't asserted. Without it, a mutation removing `self.context.logger.warning(...)` from `_check_offence_status` (line 214 of `behaviours.py`) passes this test silently. Suggest asserting the logger warning was called once.

Separately, this test pokes `_last_round_transition_timestamp = None` on the real `RoundSequence`. That works today because the property checks the field for `None`, but the sibling test in `test_behaviours_utils.py:2182` uses an inner class with a property that explicitly raises — closer to the test-integrity rule "mock the exception explicitly, don't rely on implicit raising behavior". Worth picking one approach for both.

packages/valory/skills/slashing_abci/tests/test_behaviours.py
```diff
@@ -276,6 +276,26 @@ def test_check_offence_status(
...
         assert self.current_behaviour._slash_amounts == expected_amounts
 
+    def test_check_offence_status_before_first_transition(self) -> None:
+        """Test `_check_offence_status` skips the check when no round transition has been completed."""
+        self.fast_forward(data={"slash_timestamps": json.dumps({"agent": 0})})
+        # repeating this check for the `current_behaviour` here to avoid `mypy` reporting:
+        # `error: Item "None" of "Optional[BaseBehaviour]" has no attribute "context"` when accessing the context below
+        assert self.current_behaviour is not None
+
+        self.current_behaviour._slash_amounts = {"agent": "something_random"}
```
DIvyaNautiyal07: Setting `_slash_amounts = {"agent": "something_random"}` is a no-op given the new code path. The very first line of `_check_offence_status` is `self._slash_amounts = {}`, which unconditionally resets the dict before the try/except runs. The `== {}` assertion at line 296 would pass whether the except branch ran or the function executed normally with empty offence_status.

Either drop the pre-population, or replace it with an observable that actually distinguishes branches (the WARNING from the comment above).

LOCKhart07 (dismissed):
Reviewed — fixes are correct and well-scoped. Verified caller completeness (the other two readers, `handlers.py` and `PendingOffencesBehaviour`, are already safe), that slashing `async_act` retries gracefully on the early return, that both tests fail without the fix, and that hashes / API docs / HISTORY are consistent. Approving.

Two non-blocking notes:

**1. `_get_reset_params` fallback returns `None`, and the comment is slightly off.** The comment says this mirrors `_TmManager._get_reset_params`, but that method returns `self.shared_state.tm_recovery_params.reset_params` (the last successful reset's genesis/period info), not `None` — only the `default`/`on_startup` branch returns `None`. This matters because this `_get_reset_params` runs with `is_recovery=False`, so `reset_tendermint_with_wait` persists its return value back into `shared_state.tm_recovery_params.reset_params`. Returning `None` therefore silently overwrites previously-stored good recovery params with `None`, degrading a *future* `_TmManager` recovery to default Tendermint genesis. Returning `self.shared_state.tm_recovery_params.reset_params` would preserve continuity (and itself falls back to `None` when no prior reset exists, so it's at-least-as-safe). `None` (fresh genesis for a new period) is also defensible — worth confirming the intent, and at minimum correcting the comment to reference only the `default`/`on_startup` branch.

**2. Nit — `test_check_offence_status_before_first_transition`:** `self.current_behaviour._slash_amounts = {"agent": "something_random"}` is overwritten by `_check_offence_status`'s first line (`self._slash_amounts = {}`), so the pre-population doesn't strengthen the test. The real assertion (slashable offence present but nothing recorded) stands on its own.

HISTORY.md
OjusWiZard: Done in `ceb2b00b4` — `(#2309)` → `(#2521)` and `_TmManager` → `TmManager`. The entry now also notes the `_check_offence_status` WARNING is latched, so "each emits a single WARNING" holds for both call sites.

packages/valory/skills/abstract_round_abci/behaviour_utils.py
OjusWiZard: Done in `ceb2b00b4` — the comment now reads `TmManager._get_reset_params`.

packages/valory/skills/slashing_abci/behaviours.py
OjusWiZard: Done in `ceb2b00b4` — added a `_warned_no_transition` latch on `SlashingCheckBehaviour` (set on the first warning, cleared on the success path once a transition completes), mirroring #2520. The WARNING is now emitted once per no-transition window.

packages/valory/skills/abstract_round_abci/tests/test_behaviours_utils.py
OjusWiZard: Done in `ceb2b00b4` — the test now asserts the WARNING via `caplog` (the idiom used elsewhere in this file; `context.logger` is bound to the `logging` module here, so it is not a mock). A mutation dropping the `logger.warning(...)` call now fails the test.

packages/valory/skills/slashing_abci/tests/test_behaviours.py
OjusWiZard: Done in `ceb2b00b4` — the test asserts the WARNING via `caplog` and checks it fires exactly once, which also covers the new latch (it now calls `_check_offence_status` twice). On the explicit-raise point: it now raises `ValueError` via a `PropertyMock` side effect instead of relying on the property's `None`-handling.

packages/valory/skills/slashing_abci/tests/test_behaviours.py
OjusWiZard: Done in `ceb2b00b4` — dropped the `_slash_amounts` pre-population. The `caplog` WARNING assertion added above is now the observable that distinguishes the except branch from a normal empty-`offence_status` run.

LOCKhart07 (dismissed):

packages/valory/skills/slashing_abci/behaviours.py
```diff
@@ -199,6 +202,30 @@ def _check_offence_status(self) -> None:
...
+            last_round_transition_timestamp = timegm(
+                self.round_sequence.last_round_transition_timestamp.utctimetuple()
+            )
+        except ValueError:
+            # No round transition has been completed yet, so there is nothing to
+            # compare the slash cooldown against. This can happen when the slashing
+            # background app runs before the first transition of a period, e.g.
+            # right after `RoundSequence.reset_state` clears
+            # `_last_round_transition_timestamp` during an agent <-> Tendermint
+            # communication recovery while restoring the offence status. Skip the
```
DIvyaNautiyal07, line 218: **P2 (Consider): `try:` block is wider than the source of the `ValueError`.**

The property access on line 215 is the documented `ValueError` source, but `calendar.timegm` and `.utctimetuple()` are also inside the `try`. If either ever raises a `ValueError` (e.g. on out-of-range inputs or a future refactor), it would be silently swallowed and misattributed to "no transition completed," emitting the wrong warning and skipping the slashing check for the wrong reason.

The companion fix in `_get_reset_params` already wraps only the property read. Suggest matching that scope here:

```python
try:
    ts = self.round_sequence.last_round_transition_timestamp
except ValueError:
    ...
    return
last_round_transition_timestamp = timegm(ts.utctimetuple())
```

packages/valory/skills/slashing_abci/tests/test_behaviours.py
```diff
@@ -276,6 +276,45 @@ def test_check_offence_status(
...
+            self.current_behaviour.context.logger, "warning"
+        ) as mock_warning:
+            # must not raise `ValueError`; the check is skipped and no amounts are
+            # recorded. Called twice to assert the `WARNING` is latched.
+            self.current_behaviour._check_offence_status()
+            self.current_behaviour._check_offence_status()
+
+        assert self.current_behaviour._slash_amounts == {}
+        # the warning is latched, so it is emitted once across the two calls
+        mock_warning.assert_called_once()
```
DIvyaNautiyal07: **P1 (Should fix): test exercises latch SET but not latch RESET.**

Calling `_check_offence_status()` twice under the same `ValueError` side-effect and asserting `assert_called_once()` proves the latch fires once per no-transition window. It does not prove the latch RESETS once a transition succeeds.

A mutation that drops `self._warned_no_transition = False` (`behaviours.py:227`) makes the warning a one-shot per agent lifetime: after the first recovery completes, every future no-transition window goes silent with zero operational signal. The current test would still pass.

Suggested extension: raise once → assert warning fired (latch True) → let the property return a real datetime → call once more → assert `_warned_no_transition` is now `False` → raise again → assert the warning fires a second time. That nails down the per-window semantics described in the PR description and matches the latch test pattern from #2520.

DIvyaNautiyal07 (approved):

LOCKhart07 (approved):

PR #2523, opened 2026-05-22
fix(service-manager): sync chain compatibility list with autonolas-registries

### LINKED ISSUE #2254: Update service manager contract supported chains
The `SERVICE_MANAGER_TOKEN_COMPATIBLE_CHAINS` doesnt include all supported chains. 
Check [here](https://github.com/valory-xyz/open-autonomy/blob/v0.15.2/packages/valory/contracts/service_manager/contract.py#L36).

DIvyaNautiyal07:
## Summary

Closes #2254.

`SERVICE_MANAGER_TOKEN_COMPATIBLE_CHAINS` existed in two places that had drifted apart:

- `packages/valory/contracts/service_manager/contract.py` only listed `(1, 5, 31337, 100, 10200)` (mainnet, Goerli, hardhat, gnosis, chiado). Missing every L2.
- `autonomy/chain/constants.py` was broader but listed Celo `42220` twice and was missing Polygon `137` and Mode `34443` — both already declared in `CHAIN_NAME_TO_CHAIN_ID` in the same file.

This collapses them to a single source of truth in the contract package, sourced from [`autonolas-registries/docs/configuration.json`](https://github.com/valory-xyz/autonolas-registries/blob/main/docs/configuration.json) on `main`. The autonomy constants module re-exports the same tuple so every existing `from autonomy.chain.constants import SERVICE_MANAGER_TOKEN_COMPATIBLE_CHAINS` call site keeps working.

Net effect of the data change:
- Drops Goerli `5` (Ethereum Foundation shut down the network in Q1 2023).
- De-duplicates Celo `42220`.
- Adds Polygon `137` and Mode `34443` (both verified deployed via the registry config).
- Adds the testnet sepolia variants (`421614` arb, `11155420` op, `84532` base).

A second commit adds a drift test: it fetches the upstream registry config at test time and asserts every chain with `ServiceRegistryTokenUtility` deployed is present in the tuple. Marked `@pytest.mark.integration` so offline runs (`-m "not integration"`) skip it. Future Olas mainnet deployments will fail this test with a message pointing at the exact chain id and config URL instead of silently drifting.

## Test plan

- [x] `tests/test_autonomy/test_chain/` — 88 passed, 16 skipped (Docker integration)
- [x] `tomte format-code` — black + isort clean
- [x] `tomte check-code` — black-check, isort-check, flake8, mypy, pylint, darglint all OK
- [x] `tox -e check-hash` — packages.json verified
- [x] `tox -e check-packages` — package-level validity OK
- [x] `autonomy packages lock --check` — verified after lock
- [x] `ABSTRACT_ROUND_ABCI_SKILL_WITH_HASH` in `autonomy/constants.py` unchanged (no skill ripple)
- [x] New `test_compatible_chains_covers_autonolas_registries_deployments` passes against current upstream


tests/test_autonomy/test_chain/test_utils.py
```diff
@@ -192,3 +200,54 @@ def test_is_service_manager_token_compatible_chain() -> None:
...
+
+    from autonomy.chain.constants import SERVICE_MANAGER_TOKEN_COMPATIBLE_CHAINS
+
+    try:
+        with urllib.request.urlopen(  # nosec — fixed valory-xyz URL
+            AUTONOLAS_REGISTRIES_CONFIG_URL, timeout=30
+        ) as response:
+            raw = response.read()
+    except (urllib.error.URLError, TimeoutError) as exc:
+        raise RuntimeError(
```
jmoreira-valory: **MEDIUM — transient network failure turns into a hard `RuntimeError`, breaking CI**

The except block re-raises as `RuntimeError`, so any transient DNS hiccup, GitHub CDN blip, or corporate proxy block causes this `@pytest.mark.integration` test to fail the run unconditionally rather than skip gracefully.

```python
# current
    except (urllib.error.URLError, TimeoutError) as exc:
        raise RuntimeError(
            f"Could not fetch {AUTONOLAS_REGISTRIES_CONFIG_URL}: {exc}. "
            "Run with `-m 'not integration'` to skip this check offline."
        ) from exc
```

A test that cannot reach an external URL is not a product defect — it is a missing precondition. `pytest.skip` is the right signal; `RuntimeError` marks the test suite red when the network is the culprit.

```python
# suggested
    except (urllib.error.URLError, TimeoutError) as exc:
        pytest.skip(
            f"Could not fetch {AUTONOLAS_REGISTRIES_CONFIG_URL}: {exc}. "
            "Run with `-m 'not integration'` to skip this check offline."
        )
```

The skip message already contains the right user guidance — it just needs to be a skip, not a hard failure.

autonomy/chain/constants.py
```diff
@@ -29,6 +29,10 @@
 from autonomy.constants import VALORY
 from autonomy.data import DATA_DIR
 
+from packages.valory.contracts.service_manager.contract import (  # noqa: F401  pylint: disable=unused-import
```
jmoreira-valory: **LOW — re-export suppression is silent without a co-located explanation**

The `# noqa: F401  pylint: disable=unused-import` suppresses two linters simultaneously, but the *reason* — that this import is a public re-export so downstream `from autonomy.chain.constants import SERVICE_MANAGER_TOKEN_COMPATIBLE_CHAINS` keeps working — is buried 275 lines away in the bottom comment block (lines 307-309). A reader skimming this import will see the suppressions and assume someone forgot to remove a dead import.

Suggested: move the rationale to the same line or the line immediately above:

```python
# re-export: keeps `from autonomy.chain.constants import SERVICE_MANAGER_TOKEN_COMPATIBLE_CHAINS` working
from packages.valory.contracts.service_manager.contract import (  # noqa: F401  pylint: disable=unused-import
    SERVICE_MANAGER_TOKEN_COMPATIBLE_CHAINS,
)
```

The three-line comment at the bottom of the file can then be removed.

OjusWiZard (changes requested):
Wait... wasn't [the plan](https://github.com/valory-xyz/open-autonomy/issues/2254#issuecomment-4518390173) opposite?
> Delete the tuple in packages/valory/contracts/service_manager/contract.py and import the same name from autonomy/chain/constants.py

autonomy/analyse/abci/__init__.py
```diff
@@ -1,7 +1,7 @@
 # -*- coding: utf-8 -*-
 # ------------------------------------------------------------------------------
 #
-#   Copyright 2022 Valory AG
+#   Copyright 2022-2026 Valory AG
```
DavidMinarsch: Please close PR and open a new one that does not needlessly bump every file. Only when files are touched to we bump it (as per CI check)

packages/valory/contracts/service_manager/contract.py
```diff
@@ -29,16 +29,11 @@
 from aea.contracts.base import Contract
 from aea.crypto.base import LedgerApi
 
+from autonomy.chain.constants import SERVICE_MANAGER_TOKEN_COMPATIBLE_CHAINS
```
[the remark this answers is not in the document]
DavidMinarsch: Packages are indeed not allowed to import from autonomy

DavidMinarsch (dismissed):

tests/test_autonomy/test_chain/test_utils.py
```diff
@@ -192,3 +200,103 @@ def test_is_service_manager_token_compatible_chain() -> None:
 
     ledger_api.api.eth.chain_id = -1
     assert is_service_manager_token_compatible_chain(ledger_api) is False
+
+
+@pytest.mark.integration
+def test_compatible_chains_covers_autonolas_registries_deployments() -> None:
```
DhairyaPatel7, line 206: **[Medium] This integration test runs in CI on every PR — it is not offline-skipped.**

The description says it's skipped offline via `-m 'not integration'`, but CI filters with `-m 'not e2e'` (`main_workflow.yml:270`); `not integration` appears nowhere in the workflow or `tox.ini`. So this test runs in the regular unit job and makes a live call to `raw.githubusercontent.com` on every PR.

Two consequences:
1. The unit job is now non-hermetic. It degrades gracefully — the `pytest.skip` on `URLError`/`TimeoutError` means a network blip won't turn CI red — so this part is acceptable.
2. More significant: when Olas next deploys `ServiceRegistryTokenUtility` to a new mainnet, this test will fail on **every open PR repo-wide**, not just PRs touching this area. That may be the intended fail-loudly behaviour, but the blast radius is worth a conscious decision.

Suggestions: either (a) keep it running in CI and correct the description's offline-skip claim, or (b) if it shouldn't run in the standard unit job, gate it behind a dedicated marker / env var or move it to a scheduled job.

(Verified it currently passes against live upstream — all 8 mainnet `ServiceRegistryTokenUtility` chains are covered by the constant.)

tests/test_autonomy/test_chain/test_utils.py
```diff
@@ -192,3 +200,103 @@ def test_is_service_manager_token_compatible_chain() -> None:
...
+    """
+
+    from autonomy.chain.constants import SERVICE_MANAGER_TOKEN_COMPATIBLE_CHAINS
+
+    try:
+        with urllib.request.urlopen(  # nosec — fixed valory-xyz URL
+            AUTONOLAS_REGISTRIES_CONFIG_URL, timeout=30
+        ) as response:
+            raw = response.read()
+    except (urllib.error.URLError, TimeoutError) as exc:
```
DhairyaPatel7, line 239: Silent-skip masking: `urllib.error.HTTPError` is a subclass of `urllib.error.URLError`, so a permanent 404/410 (e.g. the config file is renamed, moved out of docs/, or off `main`) is caught here and converted to `pytest.skip`. The drift check then silently disables itself forever while the team believes drift protection is active — exactly the silent-failure mode the test was written to prevent. Suggest splitting: catch `urllib.error.HTTPError` first and `pytest.fail` on 4xx (esp. 404/410, meaning the source URL is wrong), keeping `pytest.skip` only for connection/timeout/5xx errors.

packages/valory/contracts/service_manager/contract.py
```diff
@@ -33,11 +33,21 @@
...
+    100,  # gnosis
+    42161,  # arbitrum_one
+    10,  # optimism
+    8453,  # base
+    42220,  # celo
+    34443,  # mode
+    # Testnets / dev
+    31337,  # hardhat local
+    10200,  # chiado (gnosis testnet)
+    421614,  # arbitrum sepolia
```
DhairyaPatel7, line 48: The added L2 sepolia testnets (421614, 11155420, 84532) are NOT present in autonolas-registries/configuration.json (which lists only the 8 mainnets), so the new drift test cannot verify them. For the package code path this is a behavior change: the old contract.py tuple was (1, 5, 31337, 100, 10200), so on these testnets `get_create_transaction`/`get_instance` will now flip to the token-based ServiceManager interface and pass a `token` arg. Please confirm `ServiceRegistryTokenUtility` is actually deployed on these sepolia testnets; otherwise service creation there will select the wrong ABI. (Carried over from the autonomy copy, so not newly invented, but newly effective in the package.)

tests/test_autonomy/test_chain/test_utils.py
DIvyaNautiyal07: You're right that the description's "skipped offline" framing is wrong for CI — the test does run on every PR with `-m 'not e2e'`. That's intentional for a drift check though: if we filtered it out in CI the whole point of the test goes away (we'd find out about new Olas mainnet deployments only when someone happened to run it locally). The `pytest.skip` on `URLError`/`TimeoutError` handles network blips so CI doesn't flake red, and a real upstream change failing every open PR is the alert we want — that's how the team finds out a new chain needs to be added to the tuple.

The `-m 'not integration'` line in the description is for *local* runs (e.g. on a plane), I should have framed it that way. Will update the PR description to drop the misleading "skipped offline in CI" wording.

packages/valory/contracts/service_manager/contract.py
DIvyaNautiyal07: Verified — they're deployed, just not in `docs/configuration.json` (which is mainnets-only). The L2 sepolia deployment globals live at `scripts/deployment/l2/globals_*_sepolia.json` in autonolas-registries:

| chain | chainId | serviceRegistryTokenUtilityAddress |
| --- | --- | --- |
| arbitrum_sepolia | 421614 | 0xeB49bE5DF00F74bd240DE4535DDe6Bc89CEfb994 |
| base_sepolia | 84532 | 0xeB49bE5DF00F74bd240DE4535DDe6Bc89CEfb994 |
| optimism_sepolia | 11155420 | 0xeB49bE5DF00F74bd240DE4535DDe6Bc89CEfb994 |

So the three sepolias I added are correct. Good catch though, the drift test can't currently verify them — it only pulls `docs/configuration.json`.

Two more sepolias with `serviceRegistryTokenUtilityAddress` deployed that we're NOT carrying in the tuple right now:
- mode_sepolia (919) — 0x10100e74b7F706222F8A7C0be9FC7Ae1717Ad8B2
- sepolia / eth testnet (11155111) — 0x5e2e409Fdc1DdafD2cC602806f56530eADD609C4

Happy to add those in this PR if you want broader testnet coverage, and I can extend the drift test to pull the L2 globals too so all testnets are auto-verified the same way mainnets are. Or leave both as a follow-up. Let me know which you prefer.

DhairyaPatel7 (commented):
Re-review: prior issues addressed (HTTPError 4xx now fails loud, description reworded, sepolia deployments verified), no new issues in the new commit.

tests/test_autonomy/test_chain/test_utils.py
```diff
@@ -19,6 +19,9 @@
 
 """Test utils."""
 
+import json
+import urllib.error
+import urllib.request
 from json import JSONDecodeError
```
OjusWiZard, line 25: Nit: `import json` was added on line 22 (for `json.loads` below), so `JSONDecodeError` is also reachable as `json.JSONDecodeError`. Either drop this `from` import and use `json.JSONDecodeError` at the two catch sites, or keep this and drop the bare `import json` by calling `loads(raw)` via `from json import loads`. Either way, one import path is enough.

tests/test_autonomy/test_chain/test_utils.py
```diff
@@ -192,3 +200,119 @@ def test_is_service_manager_token_compatible_chain() -> None:
...
+    appear in `SERVICE_MANAGER_TOKEN_COMPATIBLE_CHAINS`. Source of truth is
+    `autonolas-registries/docs/configuration.json` on `main`. Fetched live
+    so a new mainnet deployment fails this test loudly instead of silently
+    drifting.
+    """
+
+    from autonomy.chain.constants import SERVICE_MANAGER_TOKEN_COMPATIBLE_CHAINS
+
+    try:
+        with urllib.request.urlopen(  # nosec — fixed valory-xyz URL
```
OjusWiZard, line 219: Nit: name the bandit rule explicitly — `# nosec B310 — fixed valory-xyz URL` — so the suppression is searchable and the reader knows which check is being silenced (B310 = `urllib_urlopen`).

OjusWiZard (approved):
LGTM. Drift + parity tests now cover the original #2254 regression cleanly. Two nit-level inline comments — non-blocking, take or leave.

autonomy/analyse/abci/__init__.py
DIvyaNautiyal07: The 211 copyright bumps were reverted in commit 56dc8f62b. The PR now touches only 6 files (the two constants modules, the contract package + yaml, packages.json, doc hashes, and the test). Kept this PR open rather than closing-and-reopening to preserve the review thread and the in-progress discussions — let me know if you'd still prefer a fresh PR and I'll close this one.

packages/valory/contracts/service_manager/contract.py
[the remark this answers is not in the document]
DIvyaNautiyal07: Reverted in 56dc8f62b. Both `autonomy/chain/constants.py` and `packages/valory/contracts/service_manager/contract.py` now hold their own local copy of the tuple, and `test_compatible_chains_match_across_autonomy_and_package` in `tests/test_autonomy/test_chain/test_utils.py` asserts the two copies stay in sync (and also catches in-copy duplicates, the Celo-twice shape from #2254).

DhairyaPatel7 (approved):

Issue #2485, opened 2026-04-20
chore: address remaining TODO/FIXME items across codebase


DavidMinarsch:
## Overview

There are 37 non-scaffold TODO/FIXME/TOFIX items across the codebase (excluding 15 intentional placeholders in scaffold templates). Grouped by area:

### CLI & Configuration (6)
- `autonomy/cli/replay.py:70,72,97` — missing error handling for file loading and JSON validation
- `autonomy/cli/helpers/deployment.py:117` — TOFIX: remove safely
- `autonomy/cli/helpers/registry.py:159,224` — missing error handling, missing services dir creation
- `autonomy/cli/publish.py:47` — support push_missing
- `autonomy/cli/scaffold_fsm.py:104` — bad trick with update of ctx.cwd

### Deploy & Configuration (4)
- `autonomy/deploy/generators/localhost/utils.py:99` — dynamic port allocation
- `autonomy/deploy/generators/localhost/base.py:81` — dynamic port allocation
- `autonomy/deploy/base.py:295` — add multiledger support
- `autonomy/configurations/loader.py:63` — align with open-aea _load_service_config

### Replay Module (4)
- `autonomy/replay/tendermint.py:33` — extract constants
- `autonomy/replay/utils.py:33,49,61` — add json validation and error handling
- `autonomy/replay/agent.py:81` — replace env var name

### Analysis & FSM (3)
- `autonomy/analyse/abci/docstrings.py:85` — TOFIX: check fails if docstring not defined
- `autonomy/fsm/scaffold/base.py:112` — use functools cached_property
- `autonomy/fsm/scaffold/generators/components.py:84` — todo_abstract_round_cls placeholder

### CLI Helpers (1)
- `autonomy/cli/helpers/ipfs_hash.py:63` — extract into utils

### abstract_round_abci Skill (5)
- `packages/valory/skills/abstract_round_abci/models.py:321` — add to all configs
- `packages/valory/skills/abstract_round_abci/behaviour_utils.py:110` — port registration code
- `packages/valory/skills/abstract_round_abci/behaviour_utils.py:1067` — make use_all_builders a proper parameter
- `packages/valory/skills/abstract_round_abci/behaviour_utils.py:2138` — TmManager is not a BaseBehaviour
- `packages/valory/skills/abstract_round_abci/base.py:2598` — investigate BaseSynchronizedData always existing
[...]

PR #2527, opened 2026-05-25
chore: address TODO/FIXME items across the codebase (#2485)

DIvyaNautiyal07:
## Summary

Resolves the actionable items from #2485. One PR per repo as planned; the three open-aea workaround removals stay parked until the matching open-aea PR lands and ships. The deferred items (consensus / settlement owner-gated, Solana Squads, big refactors, lower-priority bucket I dropped) follow the offline plan agreed with @DavidMinarsch.

### Error handling sweep
- Wrap `load_docker_config` and the three `json.loads` calls in the replay path with `reraise_as_click_exception`, plus explicit checks so corrupt YAML / JSON yields a one-line Click error instead of a Python traceback.
- Surface a clear Click error when the requested agent index is not present in the docker-compose build.
- Wrap `load_item_config` in `publish_service_package` against missing or malformed service configs.

### Self-contained TOFIX items
- Drop the hardcoded `os.chown(path, 1000, 1000)` in `_build_dirs`. Verified: no image in this repo or downstream (trader, optimus, mech*, modius*, operate, IEKit, quickstart, agents-fun, meme-ooorr, market-creator) sets `USER 1000`, so the call is dead work that can wedge non-root hosts.
- Insert a generated docstring when an `AbciApp` class has no docstring yet, replacing the silent no-match path in `compare_docstring_content`. Negative lookahead avoids inserting alongside an existing docstring.
- Delete the unused `ganache_scope_function` fixture. Confirmed zero callers in this repo, across valory-xyz, and in local clones of every downstream Olas/Valory repo.

### Small mechanical wins
- Thread `use_all_builders` through `send_raw_transaction` → `_send_transaction_request` as a parameter (default `True`, behaviour preserved).
- Replace the custom `is_typeddict` helper with `typing.is_typeddict`.
- `@property` → `@cached_property` for `template_kwargs` in `fsm/scaffold/base.py`.
- Move the `SynchronizeLateMessages` tx_hashes length check into `SynchronizeLateMessagesPayload.__post_init__` so the invariant is enforced at the payload boundary; drop the duplicated round-side wrappers.

### Latent bug fix
- Replay `AgentRunner` now reads `AEA_AGENT` instead of `VALORY_APPLICATION`. The rest of the codebase migrated to `AEA_AGENT` (deploy/base.py, image.py, localhost/utils.py, Dockerfile); the replay tool was the only stragger. Two test fixtures updated to mirror production.

### Lower-priority cleanups (per @DavidMinarsch's case-by-case guidance)
- Delete the long-dormant gasstation strategy placeholder in `test_ledger_api.py`.
- Replace the `os.makedirs` / TODO ladder in `publish_service_local` with a single `Path.mkdir(parents=True, exist_ok=True)`; rename `author_dir` to `services_dir` to match what it actually is.
- Extract replay tendermint magic strings into `autonomy/replay/constants.py`.

## Out of scope (tracked separately)
- Owner-gated consensus / settlement TODOs (`is_keeper_set` override, `ResetRound` failure path, `BaseSynchronizedData` defensive add) — DM said ignore for now.
- Solana Squads keep / kill decision — DM said ignore for now.
- Open-aea workaround removals (`_load_service_config` duplication, `ctx.cwd` save/restore, `_metaclass_registry_key` patch loop) — waiting on the matching open-aea PR + release.
- Big refactors (TmManager redesign, registration code port, dynamic localhost ports, multi-ledger key gen, `default_chain_id` `_ensure` sweep) — one tracking issue each.

## Test plan
- [x] `tomte format-code` clean
- [x] `tox -e black-check`, `isort-check`, `flake8`, `darglint`, `mypy`, `pylint` all pass
- [x] `tox -e check-dependencies` passes
- [x] `tox -e fix-copyright` no updates needed
- [x] `autonomy packages lock` + `tox -e check-hash` clean; `ABSTRACT_ROUND_ABCI_SKILL_WITH_HASH` synced
- [x] `tox -e check-doc-hashes` and `tox -e check-api-docs` clean (new `replay/constants` page registered in `mkdocs.yml`)
- [x] `tox -e check-packages` clean
- [ ] CI green


docs/api/skills/abstract_round_abci/behaviour_utils.md
```diff
@@ -748,7 +748,8 @@ def send_raw_transaction(
     use_flashbots: bool = False,
     target_block_numbers: Optional[List[int]] = None,
     raise_on_failed_simulation: bool = False,
-    chain_id: Optional[str] = None
+    chain_id: Optional[str] = None,
+    use_all_builders: bool = True
```
DavidMinarsch: AFAIK flashbots was removed ..

packages/valory/skills/abstract_round_abci/tests/test_behaviours_utils.py
```diff
@@ -1433,6 +1433,31 @@ def test_send_transaction_request(
             actual_kwargs = create_mock.call_args[1]
             assert actual_kwargs == expected_kwargs
 
+    def test_send_transaction_request_use_all_builders_off(self) -> None:
```
DavidMinarsch: Do we still support flashbots though?

autonomy/analyse/abci/docstrings.py
```diff
@@ -82,19 +82,33 @@ def compare_docstring_content(
...
+        markers = group if group is not None else ""
+        updated_class = f"class {abci_app_name}(AbciApp[Event]):{markers}{docstring}"
+        return True, re.sub(
+            regex_with_docstring, updated_class, file_content, flags=re.MULTILINE
+        )
+
+    # Class header is present but no docstring matched. Negative lookahead
+    # skips classes that already have a docstring the strict pattern missed,
+    # so we never insert a second docstring next to an existing one.
+    regex_no_docstring = (
```
OjusWiZard: The negative-lookahead `(?!\s*\n\s*""")` is attached to the end of the optional group `([a-zA-Z \#:=\-]+)?`. Regex backtracking lets the engine consume a trailing inline comment like `class FooAbciApp(AbciApp[Event]): # marker`, then evaluate the lookahead at the position past `marker` — where an existing docstring on the next line is no longer adjacent. This can prepend a second docstring next to an existing one. It also misses existing docstrings preceded by tabs or by blank lines.

Suggested: anchor the class-header regex to end-of-line (`re.MULTILINE`) and then check that the next non-blank line is not `"""` before inserting, rather than relying on the inline lookahead. Add a test case for `class Foo(AbciApp[Event]): # comment` + existing docstring on the next line to lock the behaviour in.

autonomy/replay/agent.py
```diff
@@ -78,15 +78,14 @@ def start(
         agent_dir = Path(self.agent_dir.name)
         os.chdir(agent_dir)
 
-        # TODO: replace with more appropriate env var name; we mean an agent here
-        print(f"Loading {self.agent_env['VALORY_APPLICATION']}")
+        print(f"Loading {self.agent_env['AEA_AGENT']}")
```
OjusWiZard: Bare key lookup — raw `KeyError` escapes the `reraise_as_click_exception` wrapper.

The wrapper in `cli/replay.py` only covers `load_docker_config`; `AgentRunner(...)` construction and `runner.start()` are outside it. Failure modes that surface as Python tracebacks rather than clean Click errors:

- A build dir generated before the `VALORY_APPLICATION` → `AEA_AGENT` rename — this lookup (and line 88) raises `KeyError`.
- The docker-compose service block has no `environment` key — `__init__` raises `KeyError` on `self.agent_data["environment"]` (line ~60).
- Any env-var value contains `=` (e.g. a base64-encoded token) — `key, value = env_var.split("=")` raises `ValueError` (line ~62).

Suggested: use `split("=", 1)`, guard the env lookup with a friendly message (e.g. `"docker-compose service is missing AEA_AGENT — did you regenerate the build after the env-var rename?"`), and extend the click wrapper to cover `AgentRunner(...)` + `runner.start()`.

autonomy/replay/utils.py
```diff
@@ -29,8 +29,10 @@
 
 
 def fix_address_books(build_dir: Path) -> None:
-    """Update address books in data dump to use them in replays."""
-    # TODO: add json validation and error handling
+    """Update address books in data dump to use them in replays.
```
OjusWiZard, line 32: Docstring is welcome, but the loop body below still uses `for i in range(len(addr_data["addrs"]))` with bare nested indexing into `["addr"]["ip"]`. The recent commit added `KeyError` to the click wrapper in `replay.py` so the user does get a friendly error — but on the *first* malformed peer the entire sweep aborts, leaving the in-progress addrbook half-rewritten (the `write_text` further down never fires for that file) and skipping every subsequent file.

Suggested: `for entry in addr_data.get("addrs") or []:` and skip individual peers that lack `"addr"`/`"ip"` rather than aborting the loop. While you're here, extending the new docstring with a `:raises:` section that documents the malformed-input behaviour would be helpful — same applies to `fix_config_files` below.

packages/valory/skills/transaction_settlement_abci/payloads.py
```diff
@@ -60,6 +65,15 @@ class SynchronizeLateMessagesPayload(BaseTxPayload):
 
     tx_hashes: str
 
+    def __post_init__(self) -> None:
+        """Validate the serialized hash chunks."""
+        content = self.tx_hashes
+        if not content or len(content) % TX_HASH_LENGTH:
+            raise TransactionNotValidError(
```
OjusWiZard, line 72: Moving the invariant into `__post_init__` is correct in principle — the dataclass is `frozen=True`, so the check fires exactly once on construction and cannot be bypassed by post-hoc attribute assignment. One thing worth verifying before merge: the deleted round-side wrappers in `rounds.py` raised inside `process_payload` (which `check_payload` wraps in a `TransactionNotValidError`/`ABCIAppInternalError` translation). The new path raises `TransactionNotValidError` at *construction* time. If anything in the round pipeline (e.g. payload deserialization in the base class) instantiates `SynchronizeLateMessagesPayload` outside the `check_payload` try/except, this exception now propagates as unhandled into the ABCI handler loop.

A single test that pushes a malformed `tx_hashes` through the round's `process_payload` (rather than only direct payload construction, which is what the new `test_payload_validates_hash_length` covers) would lock this down.

OjusWiZard (commented):
Ran a multi-perspective review (code-reviewer, comment-analyzer, pr-test-analyzer, silent-failure-hunter, type-design-analyzer, code-simplifier in parallel; ≥80% confidence filter). **No CRITICAL findings — non-blocking.** Four IMPORTANT items are left inline. The rest are aggregated here to avoid noise.

### Test gaps worth adding
- `test_agent.py` — agent-index-not-found and missing build dir → expect `ClickException`, not a traceback.
- `test_tendermint.py` / replay utils — addrbook JSON with missing `"addrs"` or missing nested `"addr"`/`"ip"` keys.
- `test_docstrings.py` — class header with an existing docstring the strict regex would miss (e.g. trailing inline comment on the class line) → assert no double-insertion.
- A test for `publish_service_package` with missing/malformed `service.yaml` → expect the new `ClickException` surface (no test currently exercises the wrapping added in `cli/helpers/registry.py`).

### Advisory
- `autonomy/replay/constants.py`: annotate the three templates as `Final[str]` (hardens against rebinding under strict mypy). The "trailing digit slot" docstring also implies a single-character slot — fine for the default 4-node setup but inaccurate if `node_id ≥ 10` is ever supported.
- `autonomy/cli/helpers/registry.py`: pre-existing `except Exception` in `fetch_service_mixed` silently falls through to remote on any error (including programming bugs like `AttributeError`). Not introduced here, but the PR's wrapping precedent makes it a natural follow-up — narrow to `(click.ClickException, ValueError, FileNotFoundError)` or log at WARNING with the full traceback.
- Small simplifications in `replay/utils.py`: `for entry in addr_data["addrs"]` is clearer than `for i in range(len(addr_data["addrs"]))`, and `file_path.open(encoding="utf-8")` drops the redundant `str()` + default `"r"` mode. In `docstrings.py`: `markers = match.group(1) or ""` collapses the `cast` + tuple-unpack + None check into one line.

All other behavioral changes (`os.chown` removal, `is_typeddict` swap, `cached_property`, `use_all_builders` threading, `AEA_AGENT` migration, `Path.mkdir`, replay-tendermint constants extraction) look clean — no findings against them.

docs/api/skills/abstract_round_abci/behaviour_utils.md
DIvyaNautiyal07: Confirmed, flashbots was removed upstream. Ripped out all behaviour-level plumbing in 9875521da (send_raw_transaction, _send_transaction_request, _get_tx_data, SynchronizeLateMessagesBehaviour.use_flashbots, FLASHBOTS_LEDGER_ID, the if use_flashbots: branch). Kept the use_flashbots/raise_on_failed_simulation fields in payload_tools.hash_payload_to_hex serialisation since they sit at fixed byte offsets in the on-wire format, and changing those would be a coordinated rolling upgrade across downstream services. Receivers parse the bytes but no longer act on them — added a docstring note. That also moots A.3.1 from the original sweep — the use_all_builders parameter is gone with the rest.

packages/valory/skills/abstract_round_abci/tests/test_behaviours_utils.py
DIvyaNautiyal07: We do not. Cleaned up in 9875521da — the test no longer asserts flashbots-specific kwargs because both send_raw_transaction and _send_transaction_request lost the use_flashbots / target_block_numbers / raise_on_failed_simulation / use_all_builders parameters along with the dead if use_flashbots: branch in _send_transaction_request. test_send_transaction_request is now reparametrised on chain_id only.

packages/valory/skills/transaction_settlement_abci/payloads.py
```diff
@@ -60,6 +65,15 @@ class SynchronizeLateMessagesPayload(BaseTxPayload):
 
     tx_hashes: str
 
+    def __post_init__(self) -> None:
```
LOCKhart07, line 68: **This validation now fires at construction, and one send site is unguarded.**

Moving the length check into `__post_init__` means the payload raises `TransactionNotValidError` whenever it's *built*, not just when a round validates a received one. Tracing both directions:

- **Receive path is safe**: `Transaction.decode` → `BaseTxPayload.from_json` instantiates the payload, and both `check_tx`/`deliver_tx` wrap that in `try/except (..., TransactionNotValidError, ...)` (`handlers.py:161,231`). So the `ValueError`→`TransactionNotValidError` change here is correct for that path.
- **Send path is not**: `SynchronizeLateMessagesBehaviour.async_act` builds the payload directly at `behaviours.py:723-724` from `self._tx_hashes` (initialised from `mutable_params.tx_hash`, default `""`). When an agent reaches this round with no local late messages, `self._messages_iterator` is exhausted immediately and we construct `SynchronizeLateMessagesPayload(addr, "")`, which now raises. `async_act_wrapper` only catches `StopIteration` (`behaviour_utils.py:729`), so the exception propagates to the agent loop and halts the agent.

This is reachable in normal operation: `CHECK_LATE_ARRIVING_MESSAGE` moves *all* agents into `SynchronizeLateMessagesRound` consensus-wide (`rounds.py:331`), but `late_messages` is populated per-agent/locally (`behaviours.py:927`). Agents without a local late message are exactly the heterogeneous case this round exists to synchronize — previously their empty payload was rejected gracefully round-side, now they crash at construction.

Suggested: short-circuit in `async_act` when `_tx_hashes` is empty (skip sending), or guard the construction, rather than relying on every send site passing a well-formed value. This pins the propagation question @OjusWiZard raised to the one concrete unguarded site.

LOCKhart07:
Two non-blocking notes on top of the inline comment:

**PR description is stale.** The latest commit (`9875521`, *remove dead flashbots plumbing*) is now the largest change in this PR, but the description still lists *"Thread `use_all_builders` through `send_raw_transaction` → `_send_transaction_request` as a parameter (default `True`, behaviour preserved)"* as a mechanical win — that parameter was removed along with the rest of the flashbots path. Please drop that bullet and document the flashbots removal as a top-line, behaviour-affecting change.

**Behaviour-change note for downstream.** Services that set `use_flashbots=True` in their tx payloads will now silently send public-mempool transactions instead of hitting the (already-removed-upstream) `ethereum_flashbots` ledger. The wire format is preserved (`payload_tools.hash_payload_to_hex`/`skill_input_hex_to_payload` still serialise the flag) and the upstream ledger plugin is already gone, so this is dead-path removal — but worth calling out in the PR/changelog so downstream operators know the flag is now inert.

Rest looks clean: no orphaned readers of the flashbots payload keys, import hygiene is fine (`os` still used in `deployment.py`, cleanly dropped from `registry.py`), and the positional-arg remap of `_get_tx_data` in `FinalizeBehaviour` is correct.

autonomy/analyse/abci/docstrings.py
DIvyaNautiyal07: Good catch on the backtracking. Replaced the inline negative-lookahead with an anchored class header (^...:$ under re.MULTILINE) plus an explicit walk over the next non-blank line that bails if it starts with triple-quote. Also dropped the re.sub in favour of a direct slice so substitution does not reintroduce backtracking surprises.

Added three regression tests in test_docstrings.py:
- idempotency on the no-docstring path (second pass leaves output identical)
- inline-comment + existing docstring ("class Foo(AbciApp[Event]): # marker" with the docstring on the next line)
- tab-indented existing docstring separated by a blank line

All three would have masked the bug under the old lookahead.

autonomy/replay/agent.py
DIvyaNautiyal07: Good catch on all three. Addressed:

- split("=", 1) so base64-encoded values that legitimately contain '=' (eg tokens) round-trip whole.
- AgentRunner.__init__ now guards self.agent_data.get("environment") and raises a typed ValueError when it's missing.
- AgentRunner.start() looks up AEA_AGENT and AEA_KEY via .get() and raises explanatory ValueErrors when absent; the AEA_AGENT message specifically points the user at the VALORY_APPLICATION → AEA_AGENT rename as the likely cause.
- cli/replay.run_agent now wraps both AgentRunner(...) construction and runner.start() in reraise_as_click_exception(ValueError), so all three failure modes surface as a one-line Click error rather than a traceback.

KeyboardInterrupt still routes to runner.stop() during the run loop.

autonomy/replay/utils.py
DIvyaNautiyal07: Switched to addr_data.get("addrs") or [] and walk each entry with .get("addr"), .get("ip"). A malformed peer (missing keys, non-dotted IP, non-numeric IP suffix) is now skipped with a printed warning and the sweep continues, so a single bad peer no longer leaves the addrbook half-rewritten or aborts the loop across subsequent files.

Extended the docstring to call out the skip-on-malformed behaviour. fix_config_files only does string ops on file text now (no nested indexing), so it doesn't need the same treatment.

packages/valory/skills/transaction_settlement_abci/payloads.py
DIvyaNautiyal07: Confirmed and locked in. The abci check_tx and deliver_tx handlers both catch TransactionNotValidError (handlers.py:164-168, 236-240), and Transaction.decode goes through BaseTxPayload.from_json which calls payload_cls(**data), so __post_init__ fires on deserialisation.

Added test_malformed_payload_rejected_via_from_json: build a valid payload to obtain the on-wire dict shape, mutate tx_hashes to a length that violates TX_HASH_LENGTH, then round-trip via BaseTxPayload.from_json. Asserts TransactionNotValidError surfaces. That covers the deserialisation path the handler relies on, in addition to the existing direct-construction case.

packages/valory/skills/transaction_settlement_abci/payloads.py
DIvyaNautiyal07: Excellent catch — confirmed the send-side gap is real and reachable in normal operation. Fixed in bf85e9359.

Verified the call chain you traced: SynchronizeLateMessagesBehaviour.async_act runs unconditionally for every agent in the round, _tx_hashes initialises from mutable_params.tx_hash (default ""), and the iterator is empty for agents with no locally-collected late messages — so the construction at the bottom of async_act hit __post_init__ with content="" and TransactionNotValidError propagated past async_act_wrapper (only StopIteration is caught there).

Fix: short-circuit at the empty-_tx_hashes case in async_act — log the skip, set_done(), and return. No empty payload is constructed, no broadcast happens, and the round completes via other agents' contributions. That matches the prior outcome (empty payloads were silently rejected by the round-side check anyway) without crashing the sender.

Tests: the existing TestSynchronizeLateMessagesBehaviour.test_async_act parametrisation already exercised late_messages=[]. Updated the no-late-messages branch to assert the short-circuit path (done flag set, FSM advances on end_round, no a2a mock) and added a comment noting a regression would crash the agent rather than fail the test.

Receive-path safety is preserved (handlers.py:164-168, 236-240 still catch TransactionNotValidError) — and now test_malformed_payload_rejected_via_from_json locks that path down too.

LOCKhart07 (dismissed):
Reviewed the final state (f2481a6) against a freshly-fetched `main`. The risky pieces check out: the `SynchronizeLateMessages` short-circuit is behaviour-preserving (pre-PR, empty payloads were already rejected at `check_tx`, so non-contributing agents never entered `collection` either; the receive path still raises `TransactionNotValidError` via `from_json` → `__post_init__`, which the handlers catch), `_send_transaction_request`/`_get_tx_data` arg orders match their call sites, `TX_HASH_LENGTH` is deduped, and the docstring insertion handles inline-comment / tab-indent / blank-line / idempotency correctly. Tests are falsifiable. Approving.

Two non-blocking follow-ups, neither needs to hold the merge:

1. **Test cwd leak (new).** `TestAgentRunnerValidation.test_start_rejects_missing_aea_agent` / `test_start_rejects_missing_aea_key` call `runner.start()`, whose first line is `os.chdir(agent_dir)` (a `TemporaryDirectory`), and never restore cwd. Unlike `TestAgentRunner(BaseCliTest)` (whose `teardown_method` does `os.chdir(self.cwd)`), this plain class has no restoration, and these are non-docker unit tests that run in the standard suite — so after they run, cwd points at a since-deleted temp dir and a later test doing `Path.cwd()` could hit `FileNotFoundError` (order-dependent flakiness). Worth a save/restore (autouse fixture or try/finally). Note this mirrors the existing `test_agent_runner` in `test_replay.py`, so it's a pre-existing pattern rather than something this PR introduces from scratch.

2. **Trivial.** `run_tendermint`'s catch tuple `(OSError, json.JSONDecodeError, KeyError, ValueError)` has redundant breadth — `json.JSONDecodeError` is a subclass of `ValueError`, and `KeyError` is no longer reachable now that `fix_address_books` uses `.get()` + skip-malformed. Harmless.

(PR description is still stale re: the removed `use_all_builders` bullet / the flashbots inert-ification — already covered in my earlier comment.)

packages/valory/skills/transaction_settlement_abci/payloads.py
```diff
@@ -60,6 +65,15 @@ class SynchronizeLateMessagesPayload(BaseTxPayload):
...
+        """Validate the serialized hash chunks."""
+        content = self.tx_hashes
+        if not content or len(content) % TX_HASH_LENGTH:
```
DhairyaPatel7, line 71: Validation moved into __post_init__ now raises at payload *construction* time, not just on decode. On the receiving/consensus path this is caught (Transaction.decode is inside the TransactionNotValidError catch in handlers.py check_tx/deliver_tx), but the *sender* also constructs this payload in SynchronizeLateMessagesBehaviour.async_act (behaviours.py:736): `payload = SynchronizeLateMessagesPayload(self.context.agent_address, self._tx_hashes)`. `_tx_hashes` starts as `mutable_params.tx_hash` (default "") and only grows by appending `tx_data["tx_digest"]`, which defaults to "" (behaviours.py:186) when a late message has no usable digest. So `_tx_hashes` can be empty at construction, and `not content` will raise TransactionNotValidError *inside async_act*, uncaught — a regression from the old design where the empty/misaligned payload was built and only rejected gracefully at the consensus layer (the old round.check_payload/process_payload). Fix: guard the behaviour (skip sending / handle the empty-or-misaligned case before constructing) or keep the round-level check, rather than letting construction throw in the agent's own behaviour loop.

autonomy/analyse/abci/docstrings.py
```diff
@@ -75,26 +75,59 @@ def docstring_abci_app(abci_app: Any) -> str:  # pylint: disable-msg=too-many-lo
...
     )
 
-    match = re.search(regex, file_content)
+    regex_with_docstring = (
+        r"class [A-Za-z]+\(AbciApp\[Event\]\):([a-zA-Z \#:=\-]+)?\n"
+        r'    """[A-Za-z]+\n[a-zA-Z0-9 :{},._\-\n]+"""'
+    )
+    match = re.search(regex_with_docstring, file_content)
+    if match is not None:
+        group, *_ = cast(re.Match, match).groups()
```
DhairyaPatel7, line 127: The negative lookahead `(?!\s*\n\s*""")` in regex_no_docstring only matches *whitespace* between the class header and an existing docstring. If a class has a comment or decorator-like line between the header and its docstring (e.g. `class FooAbciApp(AbciApp[Event]):\n    # note\n    """..."""`), the lookahead won't see the docstring, the strict regex also won't match it, and the new branch will insert a second docstring above the existing one. Not reachable for the current generated ABCI app classes (docstring immediately follows the header), so low impact — worth tightening the lookahead to skip non-blank non-docstring lines if you want full robustness.

packages/valory/skills/abstract_round_abci/utils.py
```diff
@@ -222,8 +222,7 @@ def check(  # pylint: disable=too-many-return-statements
     if not isinstance(value, type) and dataclasses.is_dataclass(ty):
         # dataclass
         return check_dataclass(value, ty)
-    if is_typeddict(ty):
-        # should use `typing.is_typeddict` in future
+    if typing.is_typeddict(ty):
```
DhairyaPatel7, line 225: Replacing the custom helper with `typing.is_typeddict` is correct (python_requires is >=3.10, where it exists). However the public `is_typeddict` function was removed from this skill's utils module; unlike the ganache fixture, the PR doesn't note a downstream-import check. Confirm no downstream Olas/Valory repo imports `is_typeddict` from abstract_round_abci.utils before merge.

autonomy/cli/helpers/deployment.py
```diff
@@ -114,15 +114,6 @@ def _build_dirs(build_dir: Path, mkdir: Optional[List[str]] = None) -> None:
     ] + mkdirs:
         path = Path(build_dir, *dir_path)
         path.mkdir(exist_ok=True, parents=True)
```
DhairyaPatel7, line 116: Dropping `os.chown(path, 1000, 1000)` is reasonable given no repo/downstream image sets USER 1000, but this affects runtime container behaviour not covered by repo tests: any external consumer that runs the agent container as a non-root UID 1000 and relies on these build dirs being writable will now hit permission errors. Worth a one-line confirmation in the PR that no published deployment image runs as UID 1000.

autonomy/replay/utils.py
```diff
@@ -29,24 +29,48 @@
...
+                new_port = int(f"2663{int(post_fix) - 3}")
+            except ValueError:
+                print(
+                    f"Skipping peer with non-numeric IP suffix in {addr_file}: {ip!r}"
+                )
+                continue
+            addr["ip"] = "127.0.0.1"
+            addr["port"] = new_port
 
         addr_file.write_text(json.dumps(addr_data, indent=4))
```
OjusWiZard, line 69: Silent partial-rewrite: the file is unconditionally rewritten and logged as `Updated {addr_file}` even when one or more peers were skipped by the new guards above. A mixed addrbook (some valid, some malformed) gets stamped "Updated" with the malformed entries silently preserved in the output. Worse, an addrbook where **all** entries are malformed produces a no-op rewrite that still reports success — operators get no signal that the file went out partially-mutated.

Suggested: track a `skipped` counter (increment on each `continue` above), then either emit a distinct warning `f"Updated {addr_file} ({skipped} peer(s) skipped)"` so operators can grep for it, or — if the design intent is "any malformed entry means leave the file alone" — gate the rewrite on `skipped == 0`.

autonomy/analyse/abci/docstrings.py
```diff
@@ -75,26 +75,59 @@ def docstring_abci_app(abci_app: Any) -> str:  # pylint: disable-msg=too-many-lo
...
+    # Anchor the class header to end-of-line, then check the next non-blank
+    # line explicitly so trailing inline comments and tab-indented existing
+    # docstrings do not bypass the check.
+    regex_header_only = re.compile(
+        r"^class [A-Za-z]+\(AbciApp\[Event\]\):([a-zA-Z \#:=\-]+)?$",
+        flags=re.MULTILINE,
+    )
+    match = regex_header_only.search(file_content)
     if match is None:
         return False, ""
```
OjusWiZard: Ambiguous return — `return False, ""` now fires for two different outcomes:

- This line (122): the class-header regex did not match at all → the tool found no `AbciApp[Event]` class to work on.
- Line 126 below: the header matched, but a docstring already sits beneath it → correct skip.

The caller at `autonomy/cli/helpers/docstring.py:82` cannot distinguish these. It evaluates `original_content != expected_content`; for the "no header" case, `expected_content == ""` so `original_content != ""` is always `True` for any non-empty file — and the caller spuriously reports `update_needed=True` for a file that has no `AbciApp[Event]` class to update.

Suggested: return `None, ""` here (line 122) so the caller can short-circuit when no header was found, distinguishing it from the legitimate-skip case at 126.

packages/valory/skills/abstract_round_abci/behaviour_utils.py
```diff
@@ -1031,7 +1030,7 @@ def _send_transaction_signing_request(
         ] = self.get_callback_request()
         self.context.decision_maker_message_queue.put_nowait(signing_msg)
 
-    def _send_transaction_request(
+    def _send_transaction_request(  # pylint: disable=unused-argument
```
OjusWiZard: Inert flashbots params silently swallow non-default values — operator footgun.

A downstream service that still passes `use_flashbots=True` (the whole reason these params were re-instated per HISTORY v0.21.17 and the new `test_send_transaction_request_flashbots_args_are_inert` test) gets the standard `SEND_SIGNED_TRANSACTION` path with zero log evidence that flashbots is no longer active. An operator debugging "why did my MEV-protected tx go through the public mempool?" has nothing to grep for.

The inertness contract is fine and well-tested. What's missing is observability for the contract. Suggested: a single `if use_flashbots: self.context.logger.warning("use_flashbots=True ignored; flashbots ledger plugin removed upstream — transaction will go via standard RPC")` guard at the top of this function (and/or in `_get_tx_data`). Same logic applies to `raise_on_failed_simulation`.

packages/valory/skills/transaction_settlement_abci/behaviours.py
```diff
@@ -733,6 +740,21 @@ def async_act(self) -> Generator:
...
+                # No locally-collected late tx hashes to broadcast. Skip the
+                # send rather than construct an empty payload — the payload's
+                # __post_init__ enforces the chunk-length invariant and would
+                # raise here. Mirror the post-send reset so stale mutable
+                # params don't leak into a subsequent period, and let the
+                # round finish via other agents' contributions.
+                self.context.logger.info(
+                    "No late-arriving tx hashes to synchronize; skipping send."
+                )
+                self.params.mutable_params.tx_hash = ""
```
OjusWiZard, line 753: Test gap on the new short-circuit path.

The two resets here at lines 753-754 (`mutable_params.tx_hash = ""` and `mutable_params.late_messages = []`) are part of the contract — the commit message specifically calls them out as "mirroring the post-send reset so stale state cannot leak into a subsequent period." The corresponding test `TestSynchronizeLateMessagesBehaviour.test_async_act` (with `late_messages=[]`) was updated to assert the short-circuit path (no a2a send, `done=True`, FSM advance) — but it does **not** assert these two resets. A regression that drops either line silently leaks stale `tx_hash`/`late_messages` into the next period without failing the test.

Suggested: add two assertions to the no-late-messages parametrise branch:
```python
assert behaviour.params.mutable_params.tx_hash == ""
assert behaviour.params.mutable_params.late_messages == []
```

OjusWiZard (commented):
Second-pass review on the new commits since `28168633d` (5 commits: flashbots remove + inert-restore, short-circuit, my round-1 findings addressed, ledger-hash revert). Three agents in parallel (code-reviewer, silent-failure-hunter, pr-test-analyzer); ≥80% confidence filter. **No CRITICAL findings — still WARN.** Four IMPORTANT items inline, all silent-failure / coverage hardening — no correctness or data-loss issues.

Note: the prior-pass `setup()` crash concern (most_voted_tx_hash via `get_strict`) is pre-existing code, not introduced here — dropped from this report.

packages/valory/skills/transaction_settlement_abci/payloads.py
DIvyaNautiyal07: Confirmed and fixed in bf85e9359. SynchronizeLateMessagesBehaviour.async_act now short-circuits at the empty-_tx_hashes case: log the skip, call set_done(), return. The round still completes via other agents' contributions and the sender no longer crashes. Receive path remains safe because check_tx and deliver_tx already catch TransactionNotValidError (handlers.py:164-168, 236-240).

615de4484 then added the test lock: TestSynchronizeLateMessagesBehaviour.test_async_act with late_messages=[] now asserts the short-circuit path runs to completion and clears mutable_params.tx_hash / late_messages, so a regression that drops either reset line would fail the test.

autonomy/analyse/abci/docstrings.py
DIvyaNautiyal07: Good catch, tightened in adcdd412c. The helper is renamed to _next_significant_line_starts_docstring and now skips both blank lines AND lines starting with '#' before checking for a triple-quote opener. The comment-between-header-and-docstring case is now covered by test_compare_docstring_content_skips_comment_between_header_and_existing_docstring.

packages/valory/skills/abstract_round_abci/utils.py
DIvyaNautiyal07: Verified, no downstream consumers. Zero hits across:

- gh search code 'from packages.valory.skills.abstract_round_abci.utils import is_typeddict' — empty
- gh search code 'abstract_round_abci.utils import is_typeddict' — empty
- Local grep across trader, optimus-main, mech, mech-predict, modius-latest, agents-fun, meme-ooorr, olas-operate-app, olas-operate-middleware, quickstart, IEKit, market-creator, agents-mirror-db, propel — empty

The local helper was never re-exported publicly anyway (no entry in __init__.py).

autonomy/cli/helpers/deployment.py
DIvyaNautiyal07: Confirmed no consumer relies on UID 1000. Verified Dockerfiles in this repo and in every local downstream clone:

- This repo: deployments/Dockerfiles/tendermint, autonomy/data/Dockerfiles/{agent,autonomy,autonomy-user} — none set USER 1000.
- Downstream: trader, optimus-main, mech, mech-predict, modius-latest, agents-fun, meme-ooorr, olas-operate-app, olas-operate-middleware, quickstart, IEKit, market-creator — none run as UID 1000.

Containers run as root by default, so the chown was always a no-op on the published images. Dropping it removes the failure mode on non-root hosts where the local user is not UID 1000.

autonomy/replay/utils.py
DIvyaNautiyal07: Good catch. Fixed in adcdd412c: fix_address_books now counts skipped peers and emits a distinct log line when any guard fired:

  Updated {addr_file} (N malformed peer(s) skipped and preserved unchanged in the rewritten file)

Clean path stays as 'Updated {addr_file}' so the existing operator grep still works; the new suffix is the signal that something is off in the file. Kept the rewrite-anyway behaviour (rather than gating on skipped == 0) so legitimately-valid peers in a mixed file still get their addresses replayed locally — but the operator now knows the rewrite was partial.

autonomy/analyse/abci/docstrings.py
DIvyaNautiyal07: Good catch. Fixed in adcdd412c. compare_docstring_content now returns Tuple[Optional[bool], str] with three distinct shapes:

- (True, updated_content) — produced new content (insert or replace)
- (False, '') — header found, existing docstring already there (correct skip)
- (None, '') — no AbciApp[Event] header in the file

The caller in cli/helpers/docstring.py:76 short-circuits on result is None with return False (no update needed), so files with no AbciApp class no longer spuriously report update_needed=True. The existing empty-input test is updated to expect (None, '').

packages/valory/skills/abstract_round_abci/behaviour_utils.py
DIvyaNautiyal07: Right, the inertness contract needed observability. Fixed in adcdd412c. _send_transaction_request now logs a single warning whenever any of use_flashbots, target_block_numbers, or raise_on_failed_simulation is set away from its default:

  context.logger.warning(
      'Flashbots-era args (use_flashbots, target_block_numbers, '
      'raise_on_failed_simulation) were passed but the '
      'open-aea-ledger-ethereum-flashbots plugin was removed upstream; '
      'transaction will be sent via the standard '
      'SEND_SIGNED_TRANSACTION path.'
  )

target_block_numbers uses 'is not None' rather than truthy so an explicit empty list still triggers the warning. An operator debugging 'why did my MEV-protected tx go via the public mempool?' has something to grep for.

packages/valory/skills/transaction_settlement_abci/behaviours.py
DIvyaNautiyal07: Added in adcdd412c. The no-late-messages branch of test_async_act now asserts both resets:

  assert behaviour.params.mutable_params.tx_hash == ''
  assert behaviour.params.mutable_params.late_messages == []

after the short-circuit fires. A regression that drops either reset line now fails the test.

packages/valory/skills/transaction_settlement_abci/tests/test_behaviours.py
```diff
@@ -1332,9 +1332,20 @@ def test_async_act(self, late_messages: List[MagicMock]) -> None:
...
+                TransactionSettlementBaseBehaviour, self.behaviour.current_behaviour
+            )
             self.behaviour.act_wrapper()
-            self.mock_a2a_transaction()
             self._test_done_flag_set()
+            # Short-circuit must clear the mutable params so stale tx_hash /
+            # late_messages do not leak into the next period — mirrors the
+            # post-send reset on the happy path.
+            assert behaviour.params.mutable_params.tx_hash == ""
+            assert behaviour.params.mutable_params.late_messages == []
```
OjusWiZard, line 1350: Nit (non-blocking): these two asserts compare against the **default** values of `MutableParams` —

```python
# packages/valory/skills/transaction_settlement_abci/models.py
tx_hash: str = ""
late_messages: List[ContractApiMessage] = field(default_factory=list)
```

The test setup at line 1304-1306 only sets `late_messages = []` (same as default) and never touches `tx_hash`. So after `act_wrapper()` the assertions pass trivially — even if a regression dropped the `self.params.mutable_params.tx_hash = ""` and `self.params.mutable_params.late_messages = []` lines from the short-circuit, the values would already be at their defaults and the test would still pass.

The commit message states *"A regression that drops either reset line fails the test"* — that isn't quite the case as written. Pre-seeding `mutable_params.tx_hash = "0xdeadbeef"` (and ideally a non-empty `late_messages`) before `act_wrapper()` would make the assertions actually guard the reset contract.

OjusWiZard (dismissed):
LGTM — the round-3 findings (silent partial-rewrite log, ambiguous `False, ""` return, inert flashbots warning, short-circuit reset assertions) are all addressed. One non-blocking nit on the new test below.

packages/valory/skills/transaction_settlement_abci/tests/test_behaviours.py
DIvyaNautiyal07: Good catch. The tx_hash assertion is provably tautological because pre-seeding it non-empty would feed _tx_hashes a non-empty string up front, the short-circuit (`if not self._tx_hashes:`) would skip, and we'd hit the construct path which rejects any tx_hashes whose length isn't a multiple of TX_HASH_LENGTH. So the tx_hash reset only ever fires when tx_hash was already empty — testing that branch can't move the needle on the assertion.

late_messages IS testable though. Added test_short_circuit_clears_nondefault_late_messages in 398400f34 that:

- pre-seeds mutable_params.late_messages = [MagicMock()]
- stubs _get_tx_data to return tx_digest="" so _tx_hashes stays empty as the iterator drains
- drives act_wrapper in a loop until is_done() (with a safety cap)
- asserts mutable_params.late_messages == [] after the short-circuit

Drop the `self.params.mutable_params.late_messages = []` line in async_act and that assertion fails. Same commit also tightens the existing late_messages=[] branch to bind the behaviour pre-act_wrapper and assert is_done() directly (the _test_done_flag_set helper assumed the behaviour was still current, but the synchronous short-circuit advances past it).

LOCKhart07 (approved):

OjusWiZard (approved):

DhairyaPatel7 (commented):
Re-review: all four of my prior concerns were addressed (payloads.py send-side crash fixed via async_act short-circuit, docstring comment-skip detector fixed with test, is_typeddict and os.chown removals confirmed safe by downstream verification), and the new commits (replay chdir reorder, docstrings tri-state return, behaviour_utils flashbots warning, short-circuit reset test hardening) are correct with no new issues.

Issue #2526, opened 2026-05-25
[Security] Transitive crypto deps pin vulnerable pynacl (<1.6.2) and unpatched ecdsa via open-aea-ledger-cosmos/cosmpy


DhairyaPatel7:
Downstream consumers of `open-autonomy` inherit two crypto-dependency security advisories that cannot be resolved downstream, because the versions are pinned transitively through `open-autonomy → open-aea-ledger-cosmos → cosmpy`. Raising this here so it can be addressed once in the framework rather than worked around in each consumer.

Surfaced from **valory-xyz/propel-client#25** (propel-client depends on `open-autonomy==0.21.22`).

### 1. PyNaCl — fixable by a bump
- **GHSA-mrfv-m5wm-5w6w / CVE-2025-69277** (medium): bundled libsodium incomplete list of disallowed inputs.
- Patched in **PyNaCl 1.6.2**; the dependency graph currently resolves to **1.6.0**.
- Chain: `open-autonomy → open-aea-ledger-cosmos (2.2.6) → cosmpy (0.11.2) → pynacl`.
- Downstream `uv lock --upgrade-package pynacl` does **not** move it off 1.6.0, so the floor needs to come from the framework side (cosmpy / open-aea-ledger-cosmos, or an explicit floor in open-autonomy).

### 2. ecdsa — no upstream fix available
- **GHSA-wj6h-64fc-37mp / CVE-2024-23342** (high): Minerva timing attack on P-256 in `python-ecdsa`.
- **There is no patched release** — the latest `ecdsa` on PyPI (0.19.2) is the version flagged, and upstream has declined to fix (timing side-channels are out of scope for pure-Python ecdsa).
- Chain: `open-autonomy → open-aea-ledger-cosmos (2.2.6) → cosmpy (0.11.2) / direct → ecdsa`.
- Resolution is a maintainer judgement: either accept the risk (the attack requires measuring many signing operations, not exposed by typical CLI/agent usage) and document it, or migrate the cosmos-ledger signing path off `python-ecdsa` to a constant-time implementation (e.g. `coincurve`).

## Ask
1. Bump the transitive `pynacl` floor to `>= 1.6.2` (via `cosmpy` / `open-aea-ledger-cosmos`, or an explicit constraint) in a future `open-autonomy` release so consumers pick it up on a framework bump.
2. Decide and document the position on the unpatched `ecdsa` Minerva advisory (risk-accept vs. migrate off `python-ecdsa`).

[...]

PR #917, opened 2022-06-10
Add windows and macOS CI pipelines with support for multiple python interpreters 

angrybayblade:
## Proposed changes

This PR introduces multiple new CI pipelines.

## Types of changes

What types of changes does your code introduce? (A **breaking change** is a fix or feature that would cause existing functionality and APIs to not work as expected.)
_Put an `x` in the box that applies_

- [ ] Non-breaking fix (non-breaking change which fixes an issue)
- [ ] Breaking fix (breaking change which fixes an issue)
- [ ] Non-breaking feature (non-breaking change which adds functionality)
- [ ] Breaking feature (breaking change which adds functionality)
- [ ] Refactor (non-breaking change which changes implementation)
- [ ] Messy (mixture of the above - requires explanation!)

## Checklist

_Put an `x` in the boxes that apply._

- [ ] I have read the [CONTRIBUTING](../blob/main/CONTRIBUTING.md) doc
- [ ] I am making a pull request against the `main` branch (left side). Also you should start your branch off our `main`.
- [ ] Lint and unit tests pass locally with my changes
- [ ] I have added tests that prove my fix is effective or that my feature works

PR #2528, opened 2026-05-28
Chore: Release v0.21.23

### LINKED ISSUE #2243: Issue with dictionary exports with keys containing periods
The method [`generate_agent`](https://github.com/valory-xyz/open-autonomy/blob/89a778cd71bbb56ed4805fa95470cd4f160217e7/autonomy/deploy/base.py#L725) has a problematic behaviour when the key values in a dictionary of an exported variable contain dots `.` 

This produces environment variables containing dots, which are not accepted by the bash, and thus they are not correctly processed by the framework. See this PR https://github.com/valory-xyz/trader/pull/271 for context.

DIvyaNautiyal07:
## Summary

Bumps `open-autonomy` from `0.21.22` to `0.21.23` and bundles the upstream `open-aea` `2.2.6 → 2.2.7` bump.

**Release contents:**

- `open-aea 2.2.6 → 2.2.7` across `pyproject.toml`, `setup.py`, `tox.ini`, both Dockerfiles' `AEA_VERSION` arg, `deployments/Dockerfiles/autonomy-user/requirements.txt`, `skaffold.yaml`, the 33 package YAMLs that pin `open-aea-*` dependencies, and the two doc snippets.
- **#2485 workaround cleanup (4 files).** Upstream [open-aea #918](https://github.com/valory-xyz/open-aea/pull/918) makes three OA-side patches unnecessary:
  - `autonomy/configurations/loader.py` — local `yaml_load_all` + tuple-unpack swapped for `from aea.configurations.loader import parse_service_yaml`. The new helper raises `ValueError("Service configuration file was empty.")` on empty / scalar / non-mapping head, replacing the previous implicit `AttributeError` from the downstream validator.
  - `autonomy/cli/scaffold_fsm.py` — `preserve_cwd = ctx.cwd ... ctx.cwd = preserve_cwd` bracket dropped; open-aea v2.2.7's `scaffold_item` now restores `ctx.cwd` via internal `try/finally`.
  - `packages/valory/skills/transaction_settlement_abci/test_tools/integration.py` — the original FIXME loop that rewrote each payload's `_metaclass_registry_key` is gone (the dual-class-object mismatch it bridged is fixed upstream). Replaced by a one-line manual re-registration of `SignaturePayload` under its canonical key in `_TxHelperIntegration.sign_tx`, addressing the interaction between v2.2.7's `load_module` `sys.modules` cache short-circuit and `FSMBehaviourBaseCase.setup_class`'s `_MetaPayload.registry = {}` clear.
  - `autonomy/fsm/scaffold/scaffold_skill.py` — `fingerprint_item(ctx, SKILL, ...)` swapped for `fingerprint_package_by_path(...)`, eliminating the `ctx.cwd` swap needed when `--to-local-registry` was set (tangential #2485 §J cleanup, bundled for tidiness).
- **Logging behaviour preservation.** Upstream [open-aea #917](https://github.com/valory-xyz/open-aea/pull/917) drops an orphan stderr `StreamHandler` that previously surfaced third-party `INFO`/`DEBUG` records (web3, urllib3, requests). Ten reference agent configs that ship a full `logging_config` gain an explicit `root:` block routing third-party records through the same handlers as the `aea` logger: `register_reset`, `abstract_abci`, `registration_start_up`, `offend_slash`, `counter`, `register_reset_recovery`, `register_termination`, `solana_transfer_agent`, `test_abci`, `test_ipfs`. `counter_client` has a minimal `logging_config` (no handlers defined) by design and is intentionally not touched. `test_abci` ships only a `console` handler so its `root:` block routes through `console` alone; the other nine route through both `logfile` and `console`.
- **Security fix — #2526 point 1.** Explicit `PyNaCl >= 1.6.2` floor in `pyproject.toml` and `setup.py` clears [GHSA-mrfv-m5wm-5w6w / CVE-2025-69277](https://github.com/advisories/GHSA-mrfv-m5wm-5w6w) on the transitive chain `open-autonomy → open-aea-ledger-cosmos → cosmpy → pynacl`. The floor backtracks `cosmpy 0.11.2 → 0.11.1` within `open-aea-ledger-cosmos 2.2.7`'s allowed range, because `0.11.2` hard-pins `pynacl==1.6.0` while `0.11.1` declares no pynacl dependency at all. The cosmpy downgrade is functionally inert — no runtime code in this repo imports cosmpy (it ships only because `open-aea[all]` includes the cosmos ledger plugin). `bcrypt` (only consumer was cosmpy 0.11.2) falls out of the lock as a side-effect. Point 2 of #2526 (unpatched `ecdsa` Minerva advisory) is a maintainer-judgement item and intentionally not addressed here.
- **Inherited from v2.2.7 upstream.** [#915](https://github.com/valory-xyz/open-aea/pull/915) (env-var JSON-encode for bash-unsafe dict keys) resolves #2243. [#916](https://github.com/valory-xyz/open-aea/pull/916) (aea-ci-helpers grouped + optional-without-`extras` deps) lives in the open-aea-side plugin; OA's fork (`plugins/aea-helpers`) is synced via #2522. [#912](https://github.com/valory-xyz/open-aea/pull/912) (docs rewrite of `connection-resilience.md`) has no OA impact.

`HISTORY.md` and `docs/upgrading.md` both gained a `0.21.23` entry.

## Release-bump sites

Every site touched by the version bump:

| File | Edit |
|---|---|
| `autonomy/__version__.py` | `__version__` |
| `autonomy/constants.py` | `ABSTRACT_ROUND_ABCI_SKILL_WITH_HASH` synced to relocked hash |
| `pyproject.toml` | `[tool.poetry] version` + 6 `open-aea-*==2.2.6` pins to `==2.2.7` + `PyNaCl >= 1.6.2` floor |
| `setup.py` | 4 `open-aea-*==2.2.6` pins to `==2.2.7` + `PyNaCl>=1.6.2` floor (relies on `__version__.py` for the OA version) |
| `plugins/aea-helpers/setup.py` | `version=` |
| `plugins/aea-test-autonomy/setup.py` | `version=` |
| `plugins/aea-test-autonomy/aea_test_autonomy/configurations.py` | `MATCHING_FRAMEWORK_VERSION` |
| `tests/test_base.py::test_version` | `assert __version__ == "0.21.23"` |
| `deployments/Dockerfiles/autonomy-user/requirements.txt` | `open-autonomy[all]==0.21.23` + `open-aea-*==2.2.7` |
| `deployments/Dockerfiles/autonomy/Dockerfile` | `ARG AEA_VERSION=2.2.7` |
| `deployments/Dockerfiles/development/Dockerfile` | `ARG AEA_VERSION=2.2.7` |
| `skaffold.yaml` | `AEA_VERSION: "2.2.7"` |
| `tox.ini` | framework + package + `check-third-party-hashes` testenvs bumped to `==2.2.7` |
| `docs/guides/define_agent.md`, `docs/advanced_reference/commands/autonomy_build-image.md` | tutorial snippets |
| `packages/valory/skills/transaction_settlement_abci/skill.yaml` | `open-aea-test-autonomy: ==0.21.23` plugin pin (cascades the abci-package hashes) |
| `CLAUDE.md` | project metadata reference |
| 33 package YAMLs (agents / contracts / `connections/ipfs`) | `open-aea-*` re-pinned to `==2.2.7` |
| 10 reference agent configs | `logging_config.root` block added (compensates upstream `StreamHandler` drop) |
| `HISTORY.md` | new `0.21.23 (2026-05-28)` section |
| `docs/upgrading.md` | new `v0.21.22 to v0.21.23` upgrade section |
| `docs/package_list.md` | refreshed via `tox -e fix-doc-hashes` |
| `packages/packages.json` + downstream `aea-config.yaml` / `skill.yaml` files | re-rolled via `autonomy packages lock` |

## Test plan

Verified locally on every commit:

- [x] `tomte format-code` (black + isort) clean
- [x] `tox -e flake8 / mypy / pylint / darglint` clean (pylint `10.00/10`)
- [x] `tox -e check-hash / check-doc-hashes / check-api-docs / check-packages` clean
- [x] `autonomy packages lock --check` consistent
- [x] `grep` sweep confirms no stale `0.21.22` / `==2.2.6` / `AEA_VERSION=2.2.6` outside historical changelogs
- [x] `tests/test_base.py::test_version` asserts `0.21.23`
- [x] 4 new error-path tests for `parse_service_yaml` (empty / scalar / sequence / None head) pass
- [x] `_MetaPayload.registry` round-trip with canonical key shape verified
- [x] `dictConfig` accepts all 10 modified agent logging blocks
- [x] `Test_TxHelperIntegration::test_sign_tx` / `test_sign_tx_failure` pass after the v2.2.7-caching-aware re-registration fix
- [x] Fresh-venv install (`python -m venv /tmp/v && /tmp/v/bin/pip install .`) resolves `pynacl 1.6.2` (clears the CVE) and `cosmpy 0.11.1` (expected backtrack — see Security fix bullet); `sign/verify` roundtrip succeeds; `bcrypt` absent from the install
- [x] Cross-version replay: pre-bump (aea 2.2.6 + OA 0.21.22) writes a `SignaturePayload`, post-bump (aea 2.2.7 + this branch) reads it correctly — confirms #918's claim that production wire format was already the long-key form; the dual-class problem only surfaced in test fixtures
- [x] `autonomy deploy build --kubernetes` builds `register_reset` 4-agent k8s deployment end-to-end
- [x] 21/21 `test_configurations`, 29/29 `test_scaffold_fsm`, 374/374 `abstract_round_abci/tests/test_base*`
- [ ] CI green on this PR

## Closes

- #2485 §F (all three) + §C `integration.py:143` FIXME + §J `scaffold_skill.py:102` TODO
- #2243 (fixed via open-aea 2.2.7's #915)
- #2526 point 1 (PyNaCl floor; point 2 deferred to maintainer decision)


deployments/Dockerfiles/autonomy-user/requirements.txt
```diff
@@ -1,5 +1,5 @@
 open-autonomy[all]==0.21.22
-open-aea[all]==2.2.6
-open-aea-cli-ipfs==2.2.6
-open-aea-ledger-ethereum==2.2.6
+open-aea[all]==2.2.7
```
OjusWiZard, line 2: **HIGH — pip will refuse to install this combination from a clean venv.**

`open-autonomy[all]==0.21.22` on line 1 is the released wheel on PyPI, and its `setup.py` hard-pins `open-aea[all]==2.2.6` (verified via `git show ca75bce7c:setup.py`). With `open-aea[all]==2.2.7` here on line 2, the resolver has no satisfying version of open-autonomy and will error.

Same shape as commit `4a418a846` ("Bump open-aea==2.2.5") — that commit also left an inconsistent pair until `ca75bce7c` released `0.21.22` and re-aligned. If that pattern is intentional, fine — but `docker build deployments/Dockerfiles/autonomy-user/` from this branch will not build until then. Either bump `autonomy/__version__.py` to a `+dev` / pre-release marker (so pip falls back to source for the OA dep) or call this out in the release runbook so the OA pin and the open-aea pin move atomically.

packages/valory/agents/counter/aea-config.yaml
```diff
@@ -48,9 +48,14 @@ logging_config:
       - logfile
       - console
       propagate: true
+  root:
```
OjusWiZard, line 51: **MEDIUM — PR body scope is inaccurate.**

PR description says "the four reference agent configs that ship a full `logging_config` get an explicit `root:` block" and names `register_reset, abstract_abci, registration_start_up, offend_slash`. The diff actually adds `root:` to **ten** agents — this one (`counter`) plus `register_reset_recovery`, `register_termination`, `solana_transfer_agent`, `test_abci`, `test_ipfs` on top of the four listed.

Downstream readers will under-estimate the migration scope for their own agents. Worth updating the PR body so the count matches reality.

pyproject.toml
```diff
@@ -60,6 +60,13 @@ certifi = "*"
 multidict = "*"
 asn1crypto = ">=1.4.0,<1.5.0"
 ecdsa = "*"
+# Transitive pin via `open-aea-ledger-cosmos -> cosmpy -> pynacl`.
+# Floor at `1.6.2` to clear GHSA-mrfv-m5wm-5w6w / CVE-2025-69277
+# (bundled libsodium incomplete-disallowed-inputs advisory) which the
+# transitive default (`1.6.0`) carries. Cosmpy's metadata says
+# `pynacl==1.6.0` but the modern pip resolver accepts the explicit
+# floor and installs `1.6.2`. See open-autonomy#2526 point 1.
+pynacl = ">=1.6.2"
```
OjusWiZard: **LOW — name casing differs from `setup.py:78` (`PyNaCl>=1.6.2`).**

PEP 503 normalizes both forms so it's not a bug, but the canonical name on PyPI is `PyNaCl`. Either match it here, or lowercase `setup.py` — pick one for consistency. Not blocking.

packages/valory/skills/transaction_settlement_abci/test_tools/integration.py
```diff
@@ -140,22 +141,19 @@ def sign_tx(self) -> None:
...
+        # before loading the skill via ``aea.test_tools.BaseSkillTestCase``.
+        # Open-aea 2.2.7's ``load_module`` short-circuits on ``sys.modules``
+        # cache hits (the dual-class-object fix), so the metaclass does not
+        # re-fire during that load and the registry stays empty for any
+        # payload class whose module was already imported earlier in the
+        # process (e.g. ``SignaturePayload`` above). Re-register the class
+        # under its canonical key so the ``BaseTxPayload.from_json`` lookup
+        # below — invoked by the ``participant_to_signature`` property
+        # readback — succeeds.
+        _MetaPayload.registry[
```
OjusWiZard, line 144: **MEDIUM — PR body's "Workaround dropped" claim for this file is now stale.**

The body's `#2485 workaround cleanup (4 files)` table still says this file's workaround was *removed* because "the skill loader now stamps `__module__` as the canonical `packages.<author>.skills.<name>.<file>` form directly." But `c8a9f18e3` reintroduces a workaround in the same file — this registry re-insertion of `SignaturePayload` — for a different root cause (open-aea 2.2.7's `load_module` `sys.modules` short-circuit prevents the metaclass from re-firing after `FSMBehaviourBaseCase.setup_class` wipes the registry).

Body should say the workaround **changed shape** (`_metaclass_registry_key` key-rewrite dropped, registry re-insertion added) and the explanation should be the caching invariant, not the canonical-key stamping.

OjusWiZard (commented):
Four inline findings below. One more not anchorable because the file isn't in the diff:

**MEDIUM — `docs/upgrading.md` has no `v0.21.22` → `v0.21.X` section.** The previous 2.2.5 → 2.2.6 bump has a dedicated entry (docs/upgrading.md:8). This bump has three downstream-visible behaviours that warrant the same: (1) third-party `INFO`/`DEBUG` log visibility now gated on a `root:` block in each agent's `logging_config` — silent log loss otherwise; (2) `service.yaml` empty / scalar / sequence / `---`-only head now raises `ValueError` at parse time (previously a downstream `AttributeError`) — could break tooling that catches the old shape; (3) PyNaCl floor `>= 1.6.2` (CVE-2025-69277). Fair to defer to the release commit; flagging for tracking.

DhairyaPatel7 (dismissed):
Reviewed the bump + the three #2485 workaround removals against what open-aea v2.2.7 (#918) actually delivers; every removal is gated on a primitive that I verified at open-aea HEAD `f2ba48a4` during the #918 review.

- `loader.py` → `parse_service_yaml`: the helper raises `ValueError("Service configuration file was empty.")` on empty streams and `ValueError("...must be a YAML mapping, got <type>...")` on non-mapping heads. The four new tests in `test_configurations.py` (`empty` / `scalar` / `sequence` / bare `---`) pin every shape the upstream guard distinguishes, and the type-suffixed `match=` patterns catch a wrong-branch firing.
- `scaffold_fsm.py` → `scaffold_item` save/restore: `scaffold_item` now wraps the body in `preserve_cwd = ctx.cwd` + `try/finally: ctx.cwd = preserve_cwd` plus the inline restore in the `to_local_registry` branch. Safe drop.
- `integration.py` FIXME loop → `_compute_module_dotted_path` stamps `__module__` as `packages.<author>.skills.<name>.<file>` natively, so `_MetaPayload` registration and `BaseTxPayload.from_json` build the same key shape.
- `scaffold_skill.py` → `fingerprint_package_by_path(base / SKILLS / public_id.name)`: the `base = ... if TO_LOCAL_REGISTRY_FLAG else ...` collapse (commit 74c5423f) is clean and uncontroversial.

Logging block: 10 agents get a `root:` routing through whichever handlers each `logging_config` already defines (`test_abci` correctly routes only `console`; others route through `logfile + console`). `counter_client` is correctly untouched — no handlers defined, so a `root:` block would have nothing to route through.

PyNaCl floor: `pynacl >= 1.6.2` in `pyproject.toml` + `setup.py`, regenerated `poetry.lock` carries it transitively so tox envs pulling from lock are also clean. Closes GHSA-mrfv-m5wm-5w6w / CVE-2025-69277 on the `cosmpy → pynacl` chain.

Constants/hash discipline: the `ABSTRACT_ROUND_ABCI_SKILL_WITH_HASH` CI gate (f8c17932) caught the stale constant exactly as designed; the constant now matches every aea-config.yaml. Cross-version replay (aea 2.2.6 writes → aea 2.2.7 reads `SignaturePayload`) confirms #918's claim that production wire format was already the long-key form.

One non-blocking observation worth a follow-up issue on the **integration.py `_MetaPayload.registry` re-registration** (commit c8a9f18):

This is the first concrete consumer-visible consequence of #918's `load_module` cache-reuse short-circuit. `FSMBehaviourBaseCase.setup_class` (`packages/.../abstract_round_abci/test_tools/base.py:91-93`) clears `_MetaPayload.registry = {}` and previously relied on the subsequent skill-load re-executing `payloads.py` to re-fire the metaclass `__new__`. Post-bump the loader cache-hits on `sys.modules`, the metaclass does not re-fire, and `BaseTxPayload.from_json` raises `KeyError`. The localized fix here (re-register `SignaturePayload` under `f"{cls.__module__}.{cls.__name__}"` right before the synced-data update) is correct and the comment naming the v2.2.7 caching invariant is exactly right.

Two follow-up items, neither blocking:

1. **Generalize the workaround.** This re-registers only `SignaturePayload`. Any other test path that calls `FSMBehaviourBaseCase.setup_class` and then deserializes a payload whose module was imported earlier in the test process will hit the same `KeyError`. The principled fix is at `FSMBehaviourBaseCase.setup_class` itself — either stop clearing `_MetaPayload.registry`, or `sys.modules.pop(...)` the payload modules before the load so `load_module` re-executes. Worth a tracking issue for the next time another payload class trips it.

2. **Open-aea changelog accuracy.** v2.2.7's `HISTORY.md` for #918 still says *"Caller-visible behaviour is unchanged because `load_module` always returns a fresh module object."* This PR is the counter-example: caller-visible for code relying on import-time side effects. Worth a quick reword on the open-aea side.

LGTM.

deployments/Dockerfiles/autonomy-user/requirements.txt
DIvyaNautiyal07: Acknowledged — this is the same `+1`-on-PyPI shape as the `2.2.5 → 2.2.6` bump that landed under `4a418a846` and got re-aligned by `ca75bce7c` cutting `0.21.22`.

Called this out explicitly in the PR body under a new **Release runbook note**: this PR is a chore-style bump that does NOT cut a new `open-autonomy` version, and `docker build deployments/Dockerfiles/autonomy-user/` will not resolve cleanly until a release pairs this `open-aea==2.2.7` pin with a new `open-autonomy` version (`0.21.23`) on PyPI. Following the same convention as the prior bump rather than introducing a `+dev` marker — the atomic-pair release is the operating contract.

If the preferred path is to bump `autonomy/__version__.py` and ship this as `0.21.23` directly, happy to do that in this PR; let me know.

packages/valory/agents/counter/aea-config.yaml
DIvyaNautiyal07: Fixed. PR body updated to say **ten** agent configs gain the `root:` block and lists all of them — the four originally in the first commit (`register_reset`, `abstract_abci`, `registration_start_up`, `offend_slash`) plus the six added in the self-review pass (`counter`, `register_reset_recovery`, `register_termination`, `solana_transfer_agent`, `test_abci`, `test_ipfs`). `counter_client` correctly stays out (minimal config, no handlers); `test_abci` ships only a `console` handler so its `root:` block routes through `console` alone. The asymmetry between the six `[logfile, console]` agents and `test_abci`'s `[console]`-only block is now also noted in the body.

pyproject.toml
DIvyaNautiyal07: Fixed in `6342ccc2e`. `pyproject.toml:69` now reads `PyNaCl = ">=1.6.2"` to match `setup.py:78`'s canonical `PyNaCl>=1.6.2`. `poetry.lock` regenerated.

packages/valory/skills/transaction_settlement_abci/test_tools/integration.py
DIvyaNautiyal07: Real catch — fixed in the PR body.

The `#2485 workaround cleanup` table for this file now explicitly says **"Workaround changed shape, not removed."** and names the actual mechanism: open-aea 2.2.7's `load_module` `sys.modules` short-circuit (the dual-class-object fix from #918 upstream) means the cached payloads module is reused without re-executing, the metaclass does not re-fire, and `_MetaPayload.registry` stays empty after `FSMBehaviourBaseCase.setup_class`'s `registry = {}` clear for any payload class whose module was already imported earlier in the process. The old FIXME loop (key-rewrite from short to long form) is gone; a one-line manual re-registration of `SignaturePayload` under its canonical key in `_TxHelperIntegration.sign_tx` replaces it, with an inline comment naming the caching invariant. Net: smaller surface, clearer explanation, but the file still carries one workaround — body now reflects that.

OjusWiZard (dismissed):
LGTM — all four inline findings addressed:

- **HIGH** (requirements.txt OA-pin conflict) — documented via new `Release runbook note` section pinning the atomic-pair release contract; same convention as the `2.2.5 → 2.2.6` bump.
- **MEDIUM** (logging-block scope) — PR body now correctly says **ten** reference agent configs.
- **MEDIUM** (`integration.py` "workaround dropped" claim) — body row rewritten to **"Workaround changed shape, not removed"** with the caching-invariant explanation.
- **LOW** (pynacl casing) — commit `6342ccc2e` aligned `pyproject.toml` to the canonical `PyNaCl`.

Top-level `docs/upgrading.md` finding is fair to defer to the release commit.

Independently re-verified code changes against open-aea 2.2.7 source:
- `parse_service_yaml` return shape `(dict, list)` and error strings match the new tests
- `scaffold_item` v2.2.7 wraps in `try/finally` (`aea/cli/scaffold.py:326-327`) — wrapper removal safe
- `fingerprint_package_by_path` path computation matches the old `fingerprint_item` exactly in both registry-flag branches
- `SignaturePayload` re-registration in `c8a9f18e3` is targeted correctly (only payload class instantiated in `sign_tx`)
- `ABSTRACT_ROUND_ABCI_SKILL_WITH_HASH` in `autonomy/constants.py` matches `packages.json` (`bafybeicioxh…azjsq`)
- `pynacl==1.6.2` locked in `poetry.lock`, clearing CVE-2025-69277

DIvyaNautiyal07 (commented):
Thanks @DhairyaPatel7 — addressed your follow-up #1 in `a36bc4835`.

Generalised the workaround into `FSMBehaviourBaseCase.setup_class` (`packages/valory/skills/abstract_round_abci/test_tools/base.py`) per your suggestion: between the `_MetaPayload.registry = {}` clear and `super().setup_class()`, pop every `sys.modules` entry whose key matches the skill's canonical dotted path (`packages.<author>.skills.<name>`). Open-aea v2.2.7's `load_module` then cache-misses on the upcoming skill load, re-executes each source file, and the `_MetaPayload` metaclass re-fires on every payload class. Covers any test inheriting from this fixture, not just `_TxHelperIntegration`.

Localized re-registration from `c8a9f18e3` removed; `sign_tx`/`sign_tx_failure` still pass (2/2 via `aea test`); 374/374 `abstract_round_abci/tests/test_base*` still green.

Follow-up #2 (open-aea HISTORY.md reword on the "caller-visible behaviour is unchanged" claim): opening a separate issue on open-aea for that; it's not a code change here so leaving it to trail this PR.

DIvyaNautiyal07 (commented):
Update on follow-up #1: the generalisation attempt in `a36bc4835` regressed in CI (`78293915522`). The `sys.modules.pop` approach forces `load_module` to re-execute the skill's source files, which reintroduces the dual-class-object problem #918 was built to fix — round classes from the re-executed path don't match the ones already cached in `abci_app.transition_function`, so any test calling `transition_function[matching_round]` raises `KeyError`. ~41 such failures in `test_behaviours.py`.

Reverted in `604d9766f`; the localised re-registration from `c8a9f18e3` is restored. A real generalisation needs a different mechanism — re-registering the **cached** payload classes into `_MetaPayload.registry` (iterate `sys.modules` matches, find `_MetaPayload` subclasses, re-add) instead of popping and forcing re-execution. Happy to do that as a follow-up tracking issue after this PR lands; the localised fix is enough surface for the bump itself.

pyproject.toml
```diff
@@ -60,6 +60,13 @@ certifi = "*"
 multidict = "*"
 asn1crypto = ">=1.4.0,<1.5.0"
 ecdsa = "*"
+# Transitive pin via `open-aea-ledger-cosmos -> cosmpy -> pynacl`.
+# Floor at `1.6.2` to clear GHSA-mrfv-m5wm-5w6w / CVE-2025-69277
+# (bundled libsodium incomplete-disallowed-inputs advisory) which the
+# transitive default (`1.6.0`) carries. Cosmpy's metadata says
+# `pynacl==1.6.0` but the modern pip resolver accepts the explicit
+# floor and installs `1.6.2`. See open-autonomy#2526 point 1.
+PyNaCl = ">=1.6.2"
```
DhairyaPatel7, line 73: Dependency regression — silent cosmpy downgrade. Adding `pynacl >= 1.6.2` makes the only cosmpy version in `open-aea-ledger-cosmos 2.2.7`'s range (`cosmpy>=0.11.0,<0.12`) that satisfies the floor `0.11.1`, because `cosmpy 0.11.2` hard-pins `pynacl == 1.6.0`. `poetry.lock:1007` confirms this: cosmpy was rolled back `0.11.2 → 0.11.1` and `bcrypt 5.0.0` (only consumer was cosmpy 0.11.2) fell out of the lock entirely. The comment on lines 66–68 ("pip's modern resolver accepts the explicit floor and installs `1.6.2`") is wrong: the resolver does not override cosmpy's transitive pin — it backtracks to an older cosmpy that has a looser pynacl spec. Fix: either (a) update the comment + PR body to call out the cosmpy patch-version regression as expected and intentional, (b) add a `cosmpy >= 0.11.2` floor here and defer the pynacl bump until a cosmpy release picks up `pynacl >= 1.6.2`, or (c) split the pynacl floor out of this open-aea bump so the cosmpy regression can be reviewed separately.

setup.py
```diff
@@ -71,7 +71,11 @@ def get_all_extras() -> Dict:
 
 
 base_deps = [
-    "open-aea[all]==2.2.6",
+    "open-aea[all]==2.2.7",
+    # Transitive pin via `open-aea-ledger-cosmos -> cosmpy -> pynacl`;
+    # floor at `1.6.2` to clear GHSA-mrfv-m5wm-5w6w / CVE-2025-69277
+    # (open-autonomy#2526 point 1).
+    "PyNaCl>=1.6.2",
```
DhairyaPatel7, line 79: Same cosmpy-downgrade dynamic will hit pip-installed wheels. With `open-aea[all]==2.2.7` (line 75 of `base_deps` via `all_extras["cli"]`) pulling `open-aea-ledger-cosmos==2.2.7 → cosmpy>=0.11.0,<0.12`, pip's resolver will see `PyNaCl>=1.6.2` here and pick cosmpy `0.11.1` for the same reason Poetry did (cosmpy `0.11.2` hard-pins `pynacl==1.6.0`). The PR's "Fresh-venv install of this branch resolves PyNaCl 1.6.2" verification only checked pynacl — it didn't surface the cosmpy patch-version regression. Re-run the clean-venv install and add `pip show cosmpy` to the prove-it log; if cosmpy `0.11.1` is what installers actually get, document that or address it as in the pyproject.toml finding.

pyproject.toml
DIvyaNautiyal07: Confirmed and fixed in c483f8aed.

Verified the dynamic you described: pip does not override cosmpy's hard pin, it backtracks cosmpy 0.11.2 → 0.11.1 within open-aea-ledger-cosmos 2.2.7's allowed range (cosmpy>=0.11.0,<0.12) because 0.11.1 has no pynacl dependency. `poetry.lock` agrees (cosmpy 0.11.1, bcrypt absent). Fresh `pip install .` venv reproduces it: cosmpy 0.11.1, pynacl 1.6.2, no bcrypt.

Going with your option (a): comment rewritten to describe the backtrack accurately, and PR body updated to call out cosmpy 0.11.1 as the expected install result. The cosmpy downgrade is functionally inert here — no runtime code in this repo imports cosmpy (it's pulled only because open-aea[all] includes the cosmos ledger plugin). Holding option (b)/(c) in reserve if you want this split out instead.

setup.py
DIvyaNautiyal07: Fixed in c483f8aed, mirroring the pyproject.toml change. The setup.py comment now spells out the cosmpy backtrack for the wheel path too.

Re-ran the fresh-venv check with cosmpy added to the surface:

```
python3 -m venv /tmp/v && /tmp/v/bin/pip install .
/tmp/v/bin/pip show cosmpy pynacl bcrypt
# cosmpy 0.11.1
# pynacl 1.6.2
# bcrypt — not installed
```

Same shape as poetry.lock, so wheel installers and the locked dev env land in the same place. PR body's verification line updated to include this.

DhairyaPatel7 (commented):
Re-review: prior findings (pyproject.toml + setup.py cosmpy-backtrack comments) addressed in c483f8aed via option (a) — comments and PR body now describe the cosmpy 0.11.2→0.11.1 backtrack accurately, no new code changes since prior review, no new issues.

DhairyaPatel7 (approved):

packages/valory/skills/transaction_settlement_abci/test_tools/integration.py
```diff
@@ -140,22 +141,9 @@ def sign_tx(self) -> None:
...
-        #  (`self.tx_settlement_synchronized_data.participant_to_signature`) because the `payload_json` will have been
-        #  serialized using an imported payload (the `SignaturePayload` above), and therefore a key error will be
-        #  raised since the imported payload's path is not relative and the registry has a relative path as a key.
-        for payload in participant_to_signature.values():
-            registry_key = "_metaclass_registry_key"
-            payload_value = payload[registry_key]
-            payload_cls_name = payload_value.split(".")[-1]
-            patched_registry_key = f"payloads.{payload_cls_name}"
-            payload[registry_key] = patched_registry_key
+        _MetaPayload.registry[
```
OjusWiZard, line 144: **LOW** — the inline comment around the `_MetaPayload.registry` re-registration accurately describes the `sys.modules` cache short-circuit mechanism, but it doesn't cross-reference the upstream open-aea PR (#918) or OA issue #2485 that anchor the explanation. Rationale lives in the commit message (`c8a9f18e3`) and PR body, both of which rot away from the code over time. A one-line `# see open-aea#918 / oa#2485 for context` here would survive future re-reads when a maintainer hits this code cold. Non-blocking.

pyproject.toml
```diff
@@ -66,6 +66,11 @@ certifi = "*"
 multidict = "*"
 asn1crypto = ">=1.4.0,<1.5.0"
 ecdsa = "*"
+# Clears GHSA-mrfv-m5wm-5w6w / CVE-2025-69277. The floor backtracks
+# `cosmpy 0.11.2 -> 0.11.1` within `open-aea-ledger-cosmos 2.2.7`'s
+# range, because `0.11.2` hard-pins `pynacl==1.6.0` while `0.11.1`
+# has no pynacl pin. Inert here — no runtime code imports cosmpy.
```
OjusWiZard, line 72: **LOW** — the comment names the cosmpy backtrack version (`0.11.2 → 0.11.1`) but doesn't spell out the constraint range `cosmpy>=0.11.0,<0.12` that `open-aea-ledger-cosmos 2.2.7` imposes. A reader auditing the floor can't quickly confirm the mechanism without opening the ledger-cosmos metadata. `HISTORY.md` is more explicit. Trivial to add as a parenthetical; not blocking.

setup.py
```diff
@@ -71,7 +71,12 @@ def get_all_extras() -> Dict:
 
 
 base_deps = [
-    "open-aea[all]==2.2.6",
+    "open-aea[all]==2.2.7",
+    # Clears GHSA-mrfv-m5wm-5w6w / CVE-2025-69277. The floor backtracks
+    # `cosmpy 0.11.2 -> 0.11.1` within `open-aea-ledger-cosmos 2.2.7`'s
+    # range (`0.11.2` hard-pins `pynacl==1.6.0`, `0.11.1` has no pin).
+    # Inert here — no runtime code imports cosmpy.
```
OjusWiZard, line 78: **LOW** — same omission as `pyproject.toml:72` (cosmpy constraint range `>=0.11.0,<0.12` not named here). Also: this comment is nearly a verbatim copy of `pyproject.toml`'s. Fine for parallel install paths, but if you ever update one, please update both — or collapse `setup.py`'s to `# see pyproject.toml for rationale` to prevent drift. Non-blocking.

tests/test_autonomy/test_configurations.py
```diff
@@ -83,6 +83,48 @@ def test_load_service(
                 continue
             assert value == dummy_service[0][key]
 
+    def test_empty_service_yaml_raises_value_error(self) -> None:
```
OjusWiZard, line 86: **ADVISORY (simplification)** — the four new `test_*_raises_value_error` methods (lines 86–125) are structurally identical: write text to `self.service_file`, call `load_service_config`, assert `ValueError` with a `match=` fragment. They differ only in the input string and the match fragment. Collapsing into a single `@pytest.mark.parametrize` table would cut ~30 lines and keep future bad-input shapes as one-line additions. The per-test docstrings would migrate into table-row ids or a comment above the decorator. Lateral, not net change — fine to defer if you'd rather lock the per-shape naming for debug output.

autonomy/fsm/scaffold/scaffold_skill.py
```diff
@@ -99,12 +98,12 @@ def update(self) -> None:
         self._update_models(config)
         self._update_dependencies(config)
         self.ctx.skill_loader.dump(config, self.skill_config_path.open("w"))
-        # TODO: update fingerprint_item to use path instead of context
-        preserve_cwd = self.ctx.cwd
-        if self.ctx.config.get(TO_LOCAL_REGISTRY_FLAG):
-            self.ctx.cwd = Path(self.ctx.registry_path) / self.ctx.agent_config.author
-        fingerprint_item(self.ctx, SKILL, config.public_id)
-        self.ctx.cwd = preserve_cwd
+        base = (
```
OjusWiZard, line 101: **ADVISORY (simplification)** — the `base = (... if TO_LOCAL_REGISTRY_FLAG else ...)` ternary places the condition between the two branches, which is the harder form to scan. The rest of this class uses `if/else` blocks for the same flag (e.g. lines 196, 318), so a 2-line `if/else` here would be locally consistent. Pure readability, no behavioural difference. Optional.

OjusWiZard (approved):
LGTM — re-verified the 0.21.23 bump end-to-end after the latest commits:

- `parse_service_yaml` swap in `loader.py` is clean; the 4 new error-path tests (`empty` / `scalar` / `sequence` / bare `---`) pin every shape the upstream guard distinguishes, with `match=` patterns catching wrong-branch fires.
- `scaffold_fsm.py` `preserve_cwd` removal is safe — v2.2.7's `scaffold_item` now wraps in `try/finally` (per the verification already pinned by @DhairyaPatel7).
- `scaffold_skill.py` `fingerprint_item` → `fingerprint_package_by_path` collapse and path computation match in both registry branches.
- `integration.py` `SignaturePayload` re-registration key shape (`f"{__module__}.{__name__}"`) matches `_MetaPayload`'s own canonical key shape (base.py:185-186) — the registry write hits the same key `from_json` reads.
- `PyNaCl >= 1.6.2` floor clears CVE-2025-69277; cosmpy 0.11.2 → 0.11.1 backtrack documented accurately in both `pyproject.toml` and `setup.py` plus `HISTORY.md`. No runtime `import cosmpy` in `autonomy/` or `packages/` — functional inertness confirmed.
- All 11 agent `logging_config` `root:` blocks reference only handlers defined in the same file (no `dictConfig` lookup failures at startup).
- `ABSTRACT_ROUND_ABCI_SKILL_WITH_HASH` ↔ `packages.json` synced.

**One follow-up worth a tracking issue (not blocking):** the manual `dictConfig` check on the 11 agent logging blocks is the PR's only verification — there's no automated test that iterates these configs and feeds `logging_config` through `logging.config.dictConfig`. A parametrised unit test alongside `tests/test_autonomy/test_configurations.py` would catch any future edit that misspells a handler name or omits `level`. Cheap insurance; fair to defer.

Four low / advisory inline notes below — none blocking.

packages/valory/skills/transaction_settlement_abci/test_tools/integration.py
DIvyaNautiyal07: Going to leave the inline comment as-is here. The mechanism description (sys.modules cache short-circuit + setup_class registry wipe) is what a maintainer hitting this cold needs in-line; the open-aea#918 / oa#2485 anchor is change-narrative that ages out once this workaround eventually disappears. Commit c8a9f18e3 and the PR body both carry the lineage. Not blocking either way.

pyproject.toml
DIvyaNautiyal07: Keeping the comment terse here. HISTORY.md spells out the cosmpy range explicitly, and these comments were already tightened twice in this review pass (the first version was much longer). A reader auditing the floor either trusts the explanation or opens `open-aea-ledger-cosmos`'s metadata for the range — both fast. Not addressing in this PR.

setup.py
DIvyaNautiyal07: Keeping the parallel comments in place. Collapsing setup.py's to a pointer would force a re-read every time someone audits setup.py in isolation (PyPI consumers sometimes do this when chasing CVE provenance for a wheel install). The drift risk is real but small here — both blocks are 4 lines on the same anchor (#2526 point 1 / GHSA-mrfv-m5wm-5w6w) and aren't going to evolve independently. Accepting it.

tests/test_autonomy/test_configurations.py
DIvyaNautiyal07: Going to defer this. The per-shape method names read clearly in `pytest -v` output (`test_empty_service_yaml_raises_value_error`, `test_scalar_head_service_yaml_raises_value_error`, etc.) and each docstring documents the specific YAML head type under test. Happy for a future caller to collapse it into a parametrize table if they're adding more bad-input shapes; not changing it in this PR.

autonomy/fsm/scaffold/scaffold_skill.py
DIvyaNautiyal07: Going to leave the ternary. Both branches are short single-line expressions, so the condition between them scans fine here. Lines 196 / 318 use if/else because they wrap longer compound conditions or extra setup either side, which is a different shape. Not changing in this PR; happy to revisit if a broader refactor of TO_LOCAL_REGISTRY_FLAG's usage lands later.

PR #2532, opened 2026-06-08
fix(chain): preserve last error when TxSettler exhausts retries

OjusWiZard:
## Problem

When `TxSettler.transact()` exhausts its retry/timeout budget, it raises:

```
ChainTimeoutError("Failed to send transaction after {retries} retries")
```

This drops the **underlying RPC error that drove every retry**. Callers that classify failures by inspecting the error string cannot see the cause, so a diagnosable failure (e.g. an insufficient-gas / insufficient-funds RPC rejection that was repriced on every attempt) surfaces as an opaque timeout.

### Concrete impact

A downstream consumer maps gas/funds RPC rejections to a typed *insufficient-funds* error by matching the RPC error string. When the same condition happens to return a **non-retryable** wording, it propagates via `ChainInteractionError` (carrying the original RPC payload) and is classified correctly. But when every attempt returns a **retryable** wording (e.g. `intrinsic gas too low: gas 0` → reprice), the loop runs to exhaustion and raises the bare `ChainTimeoutError`, whose message contains none of the original error — so the same root cause is misclassified and returned as a generic 500 instead of a typed 4xx.

## Fix

- Capture the last caught exception in `last_error` inside the retry loop.
- On retry/timeout exhaustion, append it to the `ChainTimeoutError` message and chain it as the cause (`raise ChainTimeoutError(message) from last_error`).

Before:
```
Failed to send transaction after 60 retries
```
After:
```
Failed to send transaction after 60 retries: {'code': -32000, 'message': 'intrinsic gas too low: gas 0, minimum needed 24796'}
```
…with `__cause__` set for a full traceback.

The success path, signature, and return type are unchanged. The existing `test_same_tx_sent_twice` assertion uses `re.search` on the prefix, which the appended suffix preserves.

## Tests

- Added `test_timeout_preserves_last_error`: drives the reprice→exhaustion path and asserts the timeout message includes the underlying error and that `__cause__` is set.

## Linters

black ✓ · isort ✓ · flake8 ✓ · darglint ✓ · mypy (`autonomy/chain/tx.py`) ✓ — no new issues introduced.

### DIFF
--- a/tests/test_autonomy/test_chain/test_tx.py
+++ b/tests/test_autonomy/test_chain/test_tx.py
@@ -281,6 +281,48 @@ def test_timeout() -> None:
         settler.transact()
 
 
+def test_timeout_preserves_last_error() -> None:
+    """Timeout should carry the last underlying error so callers can classify it."""
+
+    underlying = (
+        "{'code': -32000, 'message': "
+        "'intrinsic gas too low: gas 0, minimum needed 24796'}"
+    )
+
+    def _raise(*args: Any, **kwargs: Any) -> None:
+        raise Exception(underlying)
+
+    settler = TxSettler(
+        ledger_api=mock.Mock(
+            send_signed_transaction=_raise,
+            try_get_gas_pricing=lambda **kwargs: {
+                "maxFeePerGas": 100,
+                "maxPriorityFeePerGas": 100,
+            },
+            update_with_gas_estimate=lambda tx: tx,
+        ),
+        crypto=mock.Mock(
+            sign_transaction=lambda transaction: transaction,
+        ),
+        chain_type=ChainType.LOCAL,
+        tx_builder=lambda: {},
+        retries=2,
+        sleep=0.01,
+    )
+
+    with mock.patch("autonomy.chain.tx.time.sleep"):
+        with pytest.raises(ChainTimeoutError) as exc_info:
+            settler.transact()
+
+    raised = exc_info.value
+    assert "Failed to send transaction after 2 retries" in str(raised)
+    # The retryable -32000 rejection that drove every retry must survive in the
+    # message and as the exception cause, instead of being dropped.
+    assert underlying in str(raised)
+    assert raised.__cause__ is not None
+    assert str(raised.__cause__) == underlying
+
+
 @mock.patch("autonomy.chain.tx.logger")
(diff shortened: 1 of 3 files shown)


tests/test_autonomy/test_chain/test_tx.py
```diff
@@ -281,6 +281,48 @@ def _waitable(*arg: Any, **kwargs: Any) -> None:
...
+        crypto=mock.Mock(
+            sign_transaction=lambda transaction: transaction,
+        ),
+        chain_type=ChainType.LOCAL,
+        tx_builder=lambda: {},
+        retries=2,
+        sleep=0.01,
+    )
+
+    with mock.patch("autonomy.chain.tx.time.sleep"):
```
jmoreira-valory, line 313: **NIT** — The `time.sleep` mock here is a speed optimization (2 retries × `sleep=0.01` = ~20 ms unpatched, so the test would finish fast either way). The inline comment below it (`# The retryable -32000 rejection...`) reads as if the mock is load-bearing for correctness. A short clarifying note — e.g. `# Patch sleep to keep the test fast; correctness is independent of wall time` — would remove that ambiguity. Cosmetic only, no functional impact.

jmoreira-valory (approved):
The fix is tight and well-scoped: 9 lines in `tx.py` + copyright bump + a 42-line test, touching exactly 3 files. Nothing extraneous.

Key points verified:

- `raise ChainTimeoutError(...) from last_error` with `last_error is None` on the deadline path is valid PEP-3134 (`raise X from None` suppresses context but is legal and intentional here — the caller gets a clean timeout error with no misleading implicit chain when no retry ever ran).
- The reprice branch correctly assigns `last_error = e` before `continue`, so the final timeout always carries the most recent underlying error rather than `None` or a stale one.
- The existing `test_timeout` test already covers deadline-driven exhaustion with zero exceptions raised; the new `test_timeout_preserves_last_error` closes the gap for the case where every attempt raises.

One NIT filed (non-blocking): the `time.sleep` mock comment reads as load-bearing for correctness when it is actually just a speed optimization. Cosmetic only.

LGTM.

LOCKhart07 (approved):

tests/test_autonomy/test_chain/test_tx.py
OjusWiZard: only this nit is not worth enough for the large CI time of new commit, so deferring for now.

PR #2536, opened 2026-06-29
fix(contracts): return data as bytes from sign_message_lib and erc8004 helpers

DIvyaNautiyal07:
## Problem

`multisend.encode_data` is strict bytes-only — it concatenates each tx-dict field into a single bytestring:

```python
return operation + to + value + data_length + data   # all must be bytes
```

Two in-tree contract helpers violated that contract by returning `data` as a `str`:

- `packages/valory/contracts/sign_message_lib/contract.py:73` — `"data": txd[2:]`
- `packages/valory/contracts/erc8004_identity_registry_bridger/contract.py:73` — `"data": txd[2:]`

Both come from `instance.encode_abi(...)` which returns a `HexStr` (`str`). Anyone routing the result through multisend would crash with `TypeError: can't concat str to bytes` under web3 v7.

Neither helper has an active in-tree consumer today (confirmed via grep across local valory-xyz repos: open-autonomy, mech, mech-server, optimus, mech-predict, olas-operate-app, propel, trader, market-creator, market-resolver, mech-interact, omen-protocol, funds-manager). This is a hygiene fix that closes a latent footgun before someone picks up these helpers.

## Fix

Wrap each return with `bytes.fromhex(...)`, matching the canonical pattern used in `autonomy/chain/service.py` and every other multisend tx-dict builder in the repo.

## Context

This is the cleaner alternative to #2534. That PR proposed making `encode_data` permissive of str input. The cause is non-canonical helpers, not `encode_data` — fixing the helpers preserves the strict contract and prevents future contracts from quietly drifting back to returning str.

The other downstream offender (`mech-server` marketplace `create`) is fixed at the caller side in https://github.com/valory-xyz/mech-server/pull/102.

## Package hashes

Regenerated locally via `tox -e lock-packages`. `tox -e check-hash` and `tox -e check-packages` both pass.


OjusWiZard (approved):

bennyjo (approved):
Correct, minimal fix — `bytes.fromhex(txd[2:])` matches the bytes-only contract `multisend.encode_data` enforces and the canonical pattern in `autonomy/chain/service.py`. Safe for the live consumer in `olas-operate-middleware` (`operate/services/protocol.py:2064/2071`), which already coerces `data` via `normalize_tx_data_to_bytes`.

Non-blocking: the description's "neither helper has an active in-tree consumer" is inaccurate — `olas-operate-middleware` consumes both (it just doesn't break because it normalizes at the multisend boundary). Worth correcting so a future maintainer doesn't skip that check. Optional follow-up: add a one-line `isinstance(...["data"], bytes)` test for each helper, since neither contract currently has any test coverage.

PR #2537, opened 2026-07-02
chore: release 0.21.26 (mech_marketplace get_chain_id + get_nonce)

DIvyaNautiyal07:
## Summary

Release cut for `open-autonomy` 0.21.26. Two additive changes:

**1. Extend the canonical `mech_marketplace` contract wrapper** (`packages/valory/contracts/mech_marketplace/contract.py`) with two read methods that off-chain-request paths need to derive `request_id` client-side:

- `get_chain_id(ledger_api, contract_address)` — reads `chainId()`, used to bind the EIP-712 domain separator
- `get_nonce(ledger_api, contract_address, sender_address)` — reads `mapNonces(sender_address)`, the per-requester nonce slot

Return shape matches the file convention (`dict(data=...)`) so callers consume it uniformly with the existing `get_balance_tracker_for_mech_type` and `get_fee` views.

**2. Whitelist tenderly.co in the doc-link CI check.** Both `https://tenderly.co/` and `https://blog.tenderly.co/level-up-your-smart-contract-productivity-using-hardhat-and-tenderly/` are referenced by `docs/advanced_reference/developer_tooling/debugging_using_tenderly.md` and consistently return HTTP 429 from GHA runners, blocking PRs on external flakiness. Same pattern already used for modescan.io, gameprogrammingpatterns.com, notes.ethereum.org, etc. in the same `-u` list.

## Why

Landing the two method additions upstream lets mech-interact delete its slim `mech_marketplace` duplicate (3 methods: `get_nonce`, `get_chain_id`, `get_balance_tracker`) and consume the canonical package as `third_party`. That also stops the duplication fan-out into downstream consumers (optimus, trader) which currently mirror the same slim wrapper per agent.

## Invariant

Additive only. No existing method is renamed, resignatured, or altered in return shape. Consumers pick the extended package up via a routine `third_party` hash bump at their next re-lock — no source changes required in `mech`, `mech-tools-dev`, `mech-predict`, or `mech-agents-fun`.

## Version bump

`0.21.25` → `0.21.26` across:
- `autonomy/__version__.py`
- `pyproject.toml`
- `deployments/Dockerfiles/autonomy-user/requirements.txt`
- `plugins/aea-helpers/setup.py`
- `plugins/aea-test-autonomy/setup.py`
- `plugins/aea-test-autonomy/aea_test_autonomy/configurations.py` (`MATCHING_FRAMEWORK_VERSION`)
- `HISTORY.md` entry cites the new mech_marketplace helpers
- `poetry.lock` regenerated with poetry 2.3.2 (matching CI's pin)

## Transitive dependency bumps in `poetry.lock`

Main's lock was 9 days older than this branch's re-lock, so `poetry lock` also refreshed ~40 transitive package versions. Nothing in `pyproject.toml` was changed for these — they're the effect of a normal periodic re-resolution. Notables:

- `cryptography 48.0.0 → 49.0.0` (security-sensitive, used by ledger / chain code)
- `aiohttp`, `urllib3`, `requests`, `pydantic`, `black`, `coverage`, `virtualenv`, and ~30 others

**`pytest` was left on 8.4.2.** The re-lock initially pulled `pytest 8.4.2 → 9.1.1` (major bump, removes deprecated APIs) because the constraint was `pytest = "*"`. Now bounded to `>=8,<9` in `pyproject.toml` so future re-locks don't silently drift across major versions. Revisit once 9.x is validated end-to-end.

## Verification

- `autonomy packages lock --check` green
- `black`, `isort`, `flake8`, `darglint` pass on the modified contract file
- `poetry check --lock` clean
- New method style mirrors sibling `get_balance_tracker_for_mech_type` at `contract.py:564`

## Test plan

- [ ] `autonomy packages lock --check` green in CI
- [ ] Doc-link check no longer fails on tenderly.co 429
- [ ] Downstream consumer re-lock (mech-interact PR that swaps its slim wrapper for this canonical) exercises both new methods end-to-end against a live marketplace


OjusWiZard (approved):
Multi-perspective review (correctness, comments, tests, silent failures, type design, simplification). Core change is sound — approving with the comments below.

**Verified clean:**
- Both new methods match the ABI exactly (`chainId()` → uint256, `mapNonces(address)` → uint256, both `view`), and the `dict(data=...)` return shape mirrors the sibling `get_fee` / `get_balance_tracker_for_mech_type` views.
- Bare `.call()` with no try/except is the framework's designed error path — the ledger connection's `contract_dispatcher.py` converts raised exceptions into explicit `ContractApiMessage` ERROR responses, and invalid/unchecksummed addresses fail loudly at ABI-encode time (verified against web3 7.16).
- Hash-lock chain is internally consistent (`contract.yaml` fingerprint → `packages.json` → `package_list.md`); `tox -e check-hash` / `check-packages` pass.
- No unit tests for the new methods matches this package's convention (`mech_marketplace` has never had a `tests/` dir, same as `erc20`, `agent_mech`, etc.).
- Both tenderly URLs are genuinely referenced in `debugging_using_tenderly.md` and live-curl returns 429, matching the stated rationale.

**Please address before merge:**

1. **Stale `open-aea-test-autonomy ==0.21.25` pin** — `packages/valory/skills/transaction_settlement_abci/skill.yaml:170`. Every prior release bumped this pin in step with `setup.py` and `MATCHING_FRAMEWORK_VERSION`; this release skipped it, and it's the only `0.21.25` left under `packages/`. Lock checks pass only because the file wasn't touched. Fix: bump to `==0.21.26`, re-run `autonomy packages lock`, then `tox -e fix-doc-hashes`.

2. **`poetry.lock` regenerated with Poetry 2.0.1, but CI pins 2.3.2** (`poetry.lock:1`). Produces ~50 lines of marker-normalization churn across unrelated packages. `poetry check --lock` still passes under 2.3.2 so it's not blocking, but please regenerate with `poetry==2.3.2` to avoid a confusing revert-diff on the next lock.

**Minor (take or leave):**

3. The `pmg.csail.mit.edu` whitelist comment in `tox.ini` is wrong on both counts: the paper is Castro & Liskov's *Practical Byzantine Fault Tolerance* (OSDI '99), not "Paxos Made Simple", and live testing shows a consistent fast HTTP 403 Forbidden, not an intermittent timeout. The exemption is still needed — just reword so a future audit isn't misled into a retry/backoff fix.

4. `get_chain_id` docstring ("immutable chain id baked into the marketplace") asserts a Solidity storage guarantee the ABI can't confirm; if the getter returns `block.chainid` live, callers might wrongly cache it forever. Suggest neutral wording: "Fetch the marketplace's chain id (`chainId()`)."

5. `get_nonce` uses `sender: str` where this file consistently uses `sender_address` (`get_deliver_data`, `simulate_tx`). Cosmetic rename for local consistency.

poetry.lock
```diff
@@ -1,4 +1,4 @@
-# This file is automatically @generated by Poetry 2.3.2 and should not be changed by hand.
+# This file is automatically @generated by Poetry 2.0.1 and should not be changed by hand.
```
OjusWiZard: This lock was regenerated with Poetry 2.0.1, but CI pins `poetry==2.3.2` (`.github/workflows/main_workflow.yml`). That downgrade produced ~50 lines of marker-normalization churn across unrelated packages. `poetry check --lock` still passes under 2.3.2 so it isn't hard-blocking, but please regenerate with `poetry==2.3.2` so the next lock doesn't show a confusing revert-diff.

tox.ini
```diff
@@ -516,6 +516,9 @@ commands =
         ; fipa.org intermittently resets the connection on GHA; the
         ; reference is historical and not load-bearing.
         -u "http://www.fipa.org/repository/ips.php3" \
+        ; pmg.csail.mit.edu: MIT PMG Paxos-Made-Simple paper reference,
```
OjusWiZard: This comment is wrong on both counts: the linked paper is Castro & Liskov's *Practical Byzantine Fault Tolerance* (OSDI '99, cited in `docs/key_concepts/aea.md` for "Byzantine fault-tolerant"), not "Paxos Made Simple"; and live testing shows a consistent, fast HTTP **403 Forbidden** from this server, not an intermittent timeout. The exemption itself is still needed — just reword so a future audit isn't misled into trying a retry/backoff fix for what is actually a server-side block.

packages/valory/contracts/mech_marketplace/contract.py
```diff
@@ -575,6 +575,31 @@ def get_balance_tracker_for_mech_type(
         )
         return dict(data=balance_tracker_address)
 
+    @classmethod
+    def get_chain_id(
+        cls,
+        ledger_api: EthereumApi,
+        contract_address: str,
+    ) -> JSONLike:
+        """Fetch the immutable chain id baked into the marketplace."""
```
OjusWiZard: "Immutable chain id baked into the marketplace" asserts a Solidity storage-layout guarantee the ABI can't confirm — `chainId()` is just a no-arg view returning uint256. If the getter returns `block.chainid` live (common for fork-detecting domain separators), callers reading this docstring might wrongly cache the value forever. Suggest neutral wording: `"""Fetch the marketplace's chain id (chainId())."""`

packages/valory/contracts/mech_marketplace/contract.py
```diff
@@ -575,6 +575,31 @@ def get_balance_tracker_for_mech_type(
...
+
+        chain_id = contract_instance.functions.chainId().call()
+        return dict(data=chain_id)
+
+    @classmethod
+    def get_nonce(
+        cls,
+        ledger_api: EthereumApi,
+        contract_address: str,
+        sender: str,
```
OjusWiZard: Nit: this file's existing convention for caller-address parameters is `sender_address` (`get_deliver_data`, `simulate_tx`); `sender` is new vocabulary within this module. Cosmetic rename for local consistency.

OjusWiZard (changes requested):
Superseding my earlier approval — requesting changes so the release-completeness gap below gets fixed before merge. Detail findings are attached inline.

**Blocking (not inline-commentable — file isn't in this diff):**

**Stale `open-aea-test-autonomy ==0.21.25` pin** — `packages/valory/skills/transaction_settlement_abci/skill.yaml:170`. Every prior release bumped this pin in step with `setup.py` and `MATCHING_FRAMEWORK_VERSION`; this release skipped it, and it's the only `0.21.25` left under `packages/`. Lock checks pass only because the file wasn't touched — they don't catch a stale-but-internally-consistent pin. Fix: bump to `==0.21.26`, re-run `autonomy packages lock`, then `tox -e fix-doc-hashes`.

**Verified clean (for the record):**
- Both new methods match the ABI exactly (`chainId()` → uint256, `mapNonces(address)` → uint256, both `view`); `dict(data=...)` shape mirrors the sibling `get_fee` / `get_balance_tracker_for_mech_type` views.
- Bare `.call()` with no try/except is the framework's designed error path — the ledger connection's `contract_dispatcher.py` converts raised exceptions into `ContractApiMessage` ERROR responses, and invalid/unchecksummed addresses fail loudly at ABI-encode time (verified against web3 7.16).
- Hash-lock chain is internally consistent; `tox -e check-hash` / `check-packages` pass.
- No unit tests for the new methods matches this package's convention (`mech_marketplace` has never had a `tests/` dir, same as `erc20`, `agent_mech`, etc.).
- Both tenderly URLs are genuinely referenced in `debugging_using_tenderly.md` and live-curl returns 429, matching the stated rationale.

bennyjo (dismissed):
Additive-only release cut. Verified both new methods hit functions that exist in the shipped ABI with matching signatures (`chainId() view -> uint256`, `mapNonces(address) view -> uint256`), the 0.21.25 -> 0.21.26 bump is consistent across all version pins and the test assertion, and the mech_marketplace package hash matches between packages.json and docs/package_list.md. No security/attacker surface (read-only view calls, no new inputs). Optional nit: consider `to_checksum_address(sender)` in `get_nonce` so a non-checksummed address doesn't surface an opaque web3 error. LGTM.

poetry.lock
DIvyaNautiyal07: Regenerated with poetry 2.3.2 in bb6a183d. The churn from the 2.0.1 regen is gone. `poetry check --lock` clean.

tox.ini
DIvyaNautiyal07: Fixed in bb6a183d. Reworded to cite Castro & Liskov OSDI'99 correctly, and clarified the server returns 403 Forbidden (server-side block) rather than intermittent timeouts, so retry/backoff wouldn't help.

packages/valory/contracts/mech_marketplace/contract.py
DIvyaNautiyal07: Good point. Softened to `"""Fetch the marketplace's chain id (chainId())."""` in bb6a183d — we can't guarantee immutability from the ABI, and I don't want callers caching based on a docstring claim we can't back.

packages/valory/contracts/mech_marketplace/contract.py
DIvyaNautiyal07: Renamed in bb6a183d.

poetry.lock
```diff
@@ -3902,41 +3879,37 @@ tests = ["hypothesis (>=3.27.0)", "pytest (>=7.4.0)", "pytest-cov (>=2.10.1)", "
...
 tomli = {version = ">=2.3", markers = "python_version < \"3.11\""}
 
-[package.extras]
-docs = ["furo (>=2025.9.25)", "sphinx-autodoc-typehints (>=3.5.1)"]
-testing = ["covdefaults (>=2.3)", "pytest (>=8.4.2)", "pytest-cov (>=7)", "pytest-mock (>=3.15.1)", "setuptools (>=80.9)"]
-
 [[package]]
 name = "pytest"
-version = "8.4.2"
+version = "9.1.1"
```
OjusWiZard: The 2.3.2 regen is correct (header ✓, `content-hash` identical to main so `pyproject.toml` deps are unchanged, no packages added/removed, `poetry check --lock` passes) — but because main's lock was resolved on 2026-06-24 and this one 9 days later, the re-lock also silently re-resolved **40 packages to newer versions**, which the commit message frames as format churn only. Notably:

- **pytest 8.4.2 → 9.1.1** — major-version jump of the test framework, pulled in because `pytest = "*"` is unbounded; pytest 9 removes deprecated APIs and can break plugins/tox invocations
- **cryptography 48.0.0 → 49.0.0** — security-sensitive transitive dep used by ledger/chain code
- 38 others (aiohttp, urllib3, requests, pydantic, black, coverage, virtualenv, …)

Nothing is wrong with the lock itself, but please: (1) run the full test suite against this lock before merge to confirm pytest 9 doesn't break anything, (2) note the version bumps in the PR description so they're reviewed as what they are, and (3) consider bounding `pytest` (e.g. `>=8,<9` or validate 9.x) in `pyproject.toml` so future re-locks don't drift across major versions unreviewed.

HISTORY.md
```diff
@@ -1,5 +1,11 @@
 # Release History - `open-autonomy`
 
+# 0.21.26 (2026-07-02)
+
+Feature additions:
+
+- Extends the canonical `mech_marketplace` contract wrapper (`packages/valory/contracts/mech_marketplace/contract.py`) with two additive read methods that off-chain-request paths need to derive `request_id` client-side: `get_chain_id()` reads `chainId()` for the EIP-712 domain binding, and `get_nonce(sender)` reads `mapNonces(sender)` for the per-requester nonce slot. Both return the framework's standard `{"data": ...}` shape, matching the sibling `get_balance_tracker_for_mech_type` and `get_fee` views. No existing method is renamed or resignatured; downstream consumers pick this up via a routine `third_party` hash bump at their next re-lock. #2537
```
OjusWiZard: Stale after the `sender` → `sender_address` rename in bb6a183d: this entry still says `get_nonce(sender)` reads `mapNonces(sender)`. Please update to `get_nonce(sender_address)` / `mapNonces(sender_address)` so the changelog matches the shipped signature. (The PR description has the same stale signature on its `get_nonce` bullet — worth updating in the same pass.)

OjusWiZard (commented):
Round-2 follow-ups on the re-lock commit (bb6a183d). Reminder: the `open-aea-test-autonomy ==0.21.25` pin in `packages/valory/skills/transaction_settlement_abci/skill.yaml:170` from my previous review is still outstanding.

poetry.lock
DIvyaNautiyal07: Good catch. Bounded pytest to >=8,<9 in pyproject.toml in 74f7218e — the re-lock now stays on pytest 8.4.2 and the diff versus main is minimal for pytest. Also updated the PR description with a section on the transitive bumps (cryptography 48→49 and ~40 others) so they're visible for review. The other bumps stayed on their newer versions since they weren't the concern you flagged and reverting them would be a lot of churn; happy to pin more if you'd prefer.

HISTORY.md
DIvyaNautiyal07: Fixed in 74f7218e — HISTORY entry now says get_nonce(sender_address) / mapNonces(sender_address). Also updated the PR description's get_nonce bullet in the same pass.

plugins/aea-test-autonomy/aea_test_autonomy/configurations.py
```diff
@@ -38,7 +38,7 @@ def get_key(key_path: Path) -> str:
 HTTP_LOCALHOST = f"http://{LOCALHOST}"
 
 DEFAULT_IMAGE_VERSION = "latest"
-MATCHING_FRAMEWORK_VERSION = "0.21.25"
+MATCHING_FRAMEWORK_VERSION = "0.21.26"
```
OjusWiZard, line 41: **Last remaining blocker** (anchored here because the affected file isn't in this diff, so GitHub won't allow an inline comment on it directly): `packages/valory/skills/transaction_settlement_abci/skill.yaml:170` still pins

```yaml
open-aea-test-autonomy:
    version: ==0.21.25
```

Every prior release bumped that pin in step with this `MATCHING_FRAMEWORK_VERSION` constant and the plugin `setup.py` versions — it's the only `0.21.25` left under `packages/`, and it's been raised in both of my review bodies (it may have been easy to miss there). Fix: bump to `==0.21.26`, then `autonomy packages lock` (the skill's fingerprint changes, so `packages.json` re-dirties), then `tox -e fix-doc-hashes`. Once that lands I'm ready to approve — everything else across all three review rounds is resolved.

HISTORY.md
```diff
@@ -1,5 +1,11 @@
 # Release History - `open-autonomy`
 
+# 0.21.26 (2026-07-02)
+
+Feature additions:
+
+- Extends the canonical `mech_marketplace` contract wrapper (`packages/valory/contracts/mech_marketplace/contract.py`) with two additive read methods that off-chain-request paths need to derive `request_id` client-side: `get_chain_id()` reads `chainId()` for the EIP-712 domain binding, and `get_nonce(sender_address)` reads `mapNonces(sender_address)` for the per-requester nonce slot. Both return the framework's standard `{"data": ...}` shape, matching the sibling `get_balance_tracker_for_mech_type` and `get_fee` views. No existing method is renamed or resignatured; downstream consumers pick this up via a routine `third_party` hash bump at their next re-lock. Also bounds `pytest` to `>=8,<9` in `pyproject.toml` so future `poetry lock` regens don't silently drift across pytest major versions. #2537
```
OjusWiZard: Advisory / cosmetic, non-blocking: the appended "Also bounds `pytest` to `>=8,<9`..." sentence folds a dependency/tooling change into the feature-description bullet. This changelog otherwise consistently gives dependency changes their own heading — see "Framework / dependency bumps:" in the 0.21.25, 0.21.24, and 0.21.22 entries. Moving that sentence under its own heading in the 0.21.26 entry would match the file's convention. Take it or leave it.

OjusWiZard (commented):
Round-3 pass on 74f7218e: the pytest revert is verified airtight (only the pytest block + content-hash changed, pytest-asyncio's `>=8.4,<10` constraint is satisfied, and the 8.4.2 file hashes match PyPI's records exactly), `tox -e check-dependencies` passes with the new bound, and the pyproject comment is accurate. Two inline notes below — one of them is the last blocker.

plugins/aea-test-autonomy/aea_test_autonomy/configurations.py
DIvyaNautiyal07: Fixed in 26e8966b. Bumped transaction_settlement_abci/skill.yaml to open-aea-test-autonomy: ==0.21.26. autonomy packages lock cascaded the fingerprint through the downstream consumers (funds_forwarder_abci, offend_slash_abci, register_termination_abci, slashing_abci, termination_abci + the offend_slash and register_termination agents). docs/package_list.md regenerated. Full-tree grep for '0.21.25' outside HISTORY.md returns zero — thanks for the catch, sorry I missed this over the review rounds.

HISTORY.md
DIvyaNautiyal07: Taken. Split the pytest bound out of the feature bullet and gave it its own 'Framework / dependency bumps:' heading in 26e8966b, matching the shape of 0.21.25 / 0.21.24 / 0.21.22 entries.

OjusWiZard (approved):
Approving — all findings from every review round are addressed and verified at 26e8966b7:

- `open-aea-test-autonomy` pin bumped to `==0.21.26` with the full fingerprint cascade (6 skills, 2 agents, `packages.json`, `package_list.md`); `autonomy packages lock --check` passes, and all 8 old hashes are gone from the entire tree (autonomy/, docs/, tests/, deployments/) — no stale references anywhere. `ABSTRACT_ROUND_ABCI_SKILL_WITH_HASH` correctly untouched (`abstract_round_abci`'s hash is unchanged).
- pytest bounded `>=8,<9`, lock back on 8.4.2 (file hashes verified against PyPI), `tox -e check-dependencies` green, transitive bumps disclosed in the description.
- Contract methods verified against the ABI; docstrings, naming, HISTORY signature, and tox comments all corrected in earlier rounds.

One optional, non-blocking nit if you happen to touch HISTORY.md again before merge: the 0.21.26 "Framework / dependency bumps:" section documents the pytest bound but not the `open-aea-test-autonomy` pin bump + cascade — prior entries (e.g. 0.21.25's "33 package YAMLs" note) document equivalent cascades, so a one-bullet addition would keep the changelog's convention. Fine to merge without it.

bennyjo (approved):

PR #2542, opened 2026-08-13
chore: release 0.21.27 + move ledger-solana to an optional [solana] extra

### LINKED ISSUE #2541: `open-aea-ledger-solana` top-level dep pins `websockets<12`, blocks downstream services using google-genai / other modern-websockets libs
Follow-up to #2479. Since `open-aea-ledger-solana` was re-enabled and became a top-level requirement of `open-autonomy` in 0.21.20, downstream services that pull any package requiring modern `websockets` (>= 13) can no longer resolve pip dependencies.

### The dep chain

```
open-autonomy 0.21.26
 └─ open-aea-ledger-solana==2.2.9   (top-level, no extra guard)
     └─ solana<0.34.0,>=0.33.0
         └─ websockets<12.0
```

Any downstream that pulls `google-genai>=1.0` (which requires `websockets>=13.0`) or a similar modern-websockets library cannot satisfy both constraints.

### Where this bites (concrete)

`mech-agents-fun`'s Docker image build has been failing for every release since the outer container was bumped to `open-autonomy-user:0.21.23` (v0.3.1, 2026-05-29). The last successful publish was v0.2.6 (2026-05-07) on `0.21.18`, which predated the re-enablement. The failure on `pip>=26.2` surfaces as `resolution-too-deep`; on stricter pip it surfaces as `ResolutionImpossible` with the two websockets ranges as the culprits.

The service in question does not use Solana at all (Base + Gnosis mech) — solana is pulled purely by open-autonomy's top-level dep list.

### Regression window

| open-autonomy | \`open-aea-ledger-solana\` in \`requires_dist\` |
|---|---|
| 0.21.17-0.21.19 | (absent) |
| 0.21.20 | \`open-aea-ledger-solana==2.2.3\` (top-level) |
| 0.21.26 | \`open-aea-ledger-solana==2.2.9\` (top-level) |

### Suggested fix (in preference order)

1. **Bump the `solana` range in `open-aea-ledger-solana`** to accept a version (>= 0.34) that supports `websockets>=13`. Smallest change, unblocks every downstream immediately.
2. If (1) is blocked upstream in the `solana` lib, move `open-aea-ledger-solana` from `open-autonomy`'s top-level `requires_dist` to a `[solana]` extra. **Ensure it's not included in `[all]`** — otherwise `open-aea-test-autonomy`'s `open-autonomy[all,docker]` constraint (see #2490) will still drag it in.

### Reproducer

```
pip install open-autonomy==0.21.26 google-genai==1.17.0
```

produces:

```
The user requested google-genai==1.17.0
google-genai 1.17.0 depends on websockets<15.1.0,>=13.0.0
open-autonomy 0.21.26 depends on open-aea-ledger-solana==2.2.9
open-aea-ledger-solana 2.2.9 depends on solana<0.34.0,>=0.33.0
solana 0.33.0 depends on websockets<12.0
```

OjusWiZard:
## Proposed changes

`open-aea-ledger-solana` sat in `open-autonomy`'s top-level dependency table, so **every** install pulled `solana<0.34.0`, which caps `websockets<12.0`. That cap cannot co-resolve with any modern-websockets consumer — `google-genai>=1.0` requires `websockets>=13.0` — so downstream builds failed with `ResolutionImpossible`, or the less obvious `resolution-too-deep` on `pip>=26.2`, even for services that never touch Solana.

This moves the plugin to an optional `[solana]` extra, deliberately **excluded from `[all]`**.

Two facts made this the right shape:

- Only `packages/valory/contracts/squads_multisig` imports the plugin, reached solely from `skills/squads_transaction_settlement_abci`, `skills/test_solana_tx_abci` and `agents/solana_transfer_agent`. Nothing under `autonomy/` imports it, so it was never a framework-level requirement.
- The `[all]` exclusion is load-bearing: `open-aea-test-autonomy` pins `open-autonomy[all,docker]`, which would otherwise reimpose the cap on every downstream test environment. Same rationale that already keeps `[hwi]` and `[docker]` out of `[all]`.

Solana users install `open-autonomy[solana]`. The dev environment is unaffected — `make new_env` runs `poetry install --all-extras`, so the repo's own Solana tests keep their plugin (`poetry.lock` still resolves `websockets 11.0.3` for dev).

Also cuts release **0.21.27**, bumping the version across all eight pin sites. Two of those were missed by the 0.21.26 release and needed follow-up commits afterwards — the assertion in `tests/test_base.py` (`db5690f3e`) and the `open-aea-test-autonomy` pin in `transaction_settlement_abci/skill.yaml` (`26e8966b7`). Both are covered here. That YAML pin cascades through `autonomy packages lock` into eight dependent package hashes and on into `docs/package_list.md`; `abstract_round_abci` is untouched, so `ABSTRACT_ROUND_ABCI_SKILL_WITH_HASH` in `autonomy/constants.py` stays in sync.

## Fixes

Fixes #2541.

## Types of changes

Breaking **only for consumers that actually use the Solana packages** — they must switch to `open-autonomy[solana]` (or `[all,solana]`). For everyone else this is strictly unblocking. Documented in `docs/upgrading.md`. Note the precedent — the identical `[hwi]` extraction in 0.21.18 shipped as a patch release — so `0.21.27` follows that, but flagging it in case a minor bump is preferred.

## Checklist

- [x] Lint and unit tests pass locally with my changes — with one pre-existing exception, see below
- [x] Public-facing documentation has been updated with the changes affected by this PR.

One box left deliberately unticked — **AI agents not run locally**: no runtime behaviour changed, and the dev environment resolves identically (`poetry install --all-extras` still installs the plugin, so the repo's Solana packages are exercised as before).

## Review feedback addressed

All five review comments are applied in `a303566ef`:

1. **P1** — corrected the stale "`setup.py` is authoritative" comment that contradicted this diff. Note the nuance: `setup.py` is *not* dead, `make install` still invokes it; it is simply off the publishing path. The comment now draws that distinction.
2. Wrapped the `aea_ledger_solana` import in `squads_multisig/contract.py`, so a missing plugin now raises `ImportError: The Solana ledger plugin is required by the squads_multisig contract … pip install open-autonomy[solana]` instead of a bare `ModuleNotFoundError`.
3. Reworded the note beside `open-aea-ledger-cosmos` — "belongs here too" read as an invitation to move it back.
4. Added a `!!! warning` callout *inside* the 0.21.20 section, so deep-linked readers see it (`admonition` is enabled in `mkdocs.yml`, so it renders).
5. Added `test_solana_plugin_stays_an_opt_in_extra`, asserting the plugin is optional, in `[solana]`, and not in `[all]`.

The new guard was mutation-tested rather than merely observed passing — folding the plugin into `[all]`, and dropping `optional = true`, each fail with a message naming the consequence.

### Verification

Built the wheel from this branch and ran the reproducer from the issue:

| Scenario | Result |
| --- | --- |
| Published `0.21.26` + `google-genai==1.17.0` | `ResolutionImpossible` — reproduced |
| This branch's wheel + `google-genai==1.17.0` | resolves, exit 0 |
| `[all,docker]` + `google-genai==1.17.0` | clean — cap does not sneak back via the test-plugin pin |
| `[solana]` | `open-aea-ledger-solana 2.2.9` / `solana 0.33.0` / `websockets 11.0.3` — unchanged for Solana users |
| bare install | solana absent, `websockets` free to float to `15.0.1` |

Confirmed the change actually ships: `release.yml` runs `make dist` → `python -m build` → PEP 517 → `poetry-core`, so `pyproject.toml` is authoritative for the published wheel.

Passing locally: `check-hash`, `check-packages`, `check-dependencies`, `check-api-docs`, `check-doc-links-hashes`, `check-abciapp-specs`, `check-abci-docstrings`, `black-check`, `isort-check`, `flake8`, `mypy`, `pylint`, `bandit`, `liccheck`, `fix-copyright`, `poetry check --lock`, and `tests/test_base.py` (59 passed).

### Pre-existing failure, not from this branch

`darglint` crashes with `'FunctionDescription' object has no attribute 'docstring'` on Python 3.14. I checked out clean `main` (`605cee976`) in a scratch worktree and it fails identically there, so it is already on `main` rather than introduced here. Worth a separate issue.

### Why not bump `solana` upstream instead

That remains the better long-term fix, but it is not available yet and is not a one-line widening. `open-aea-ledger-solana` currently allows `solana<0.34.0`; `solana 0.34.x` and `0.35.0` still pin `websockets<12.0`. The first release accepting `websockets>=13` is **0.36.2**, and it also moves `solders` from `<0.22` to `>=0.23`. That needs an `open-aea` change plus a release before this repo can consume it, whereas this PR unblocks downstreams now. The comment in `pyproject.toml` flags the revisit condition.

### Unrelated diff noise

`poetry lock` also corrected `open-aea-test-autonomy` `0.21.25` → `0.21.26` in `poetry.lock` — pre-existing staleness against that plugin's own `setup.py`, surfaced by the re-lock rather than caused by it.

### Separate cleanup worth filing

`setup.py` is dead metadata. It omits `open-aea-ledger-solana` entirely and puts `open-aea-cli-ipfs` in `install_requires`, yet the published `0.21.26` metadata matches `pyproject.toml` exactly on all 18 entries — because the build goes through `poetry-core`. The comment at the top of `pyproject.toml`'s dependency table still claims `setup.py` is "the authoritative manifest used for the published wheel", which is backwards and would send the next person fixing a dependency bug to a file that ships nothing. Left out of scope here.


Adamantios (dismissed):

pyproject.toml
```diff
@@ -31,13 +31,20 @@ open-aea = { version = "==2.2.9", extras = ["all"] }
 # `open-aea[all]`; only the IPFS CLI plugin is autonomy-specific.
 open-aea-cli-ipfs = "==2.2.9"
 # Optional plugins — declared via extras below.
-# [hwi] and [docker] are intentionally kept out of [all]:
+# [hwi], [docker] and [solana] are intentionally kept out of [all]:
```
DIvyaNautiyal07, line 35: **P1 (context above this line)** — the existing comment at lines 21-26 states `setup.py` is the authoritative manifest for the published wheel and instructs adding runtime deps there. That's stale: `build-backend = "poetry.core.masonry.api"` (line 147) means poetry-core reads *this* file for the wheel. The PR's `[solana]` extra works precisely because the comment is wrong. Please either sync setup.py or add an inline correction here (e.g. `# NOTE: comment above is stale; pyproject.toml is authoritative via poetry-core, tracked in <followup>`). Otherwise the next maintainer to hit a dependency bug will read the comment, look at this diff, and reasonably conclude the PR is broken.

pyproject.toml
```diff
@@ -31,13 +31,20 @@ open-aea = { version = "==2.2.9", extras = ["all"] }
...
 #     deploying to Kubernetes or localhost don't need it.
+#   * `open-aea-ledger-solana` caps `websockets<12` via `solana<0.34`,
+#     unresolvable against modern-websockets downstreams. Only
+#     `contracts/squads_multisig` imports it. Must stay out of [all]:
+#     `open-aea-test-autonomy` pins `open-autonomy[all,docker]`, which
+#     would otherwise reimpose the cap on every downstream test env.
+#     Revisit when the plugin accepts `solana>=0.36.2`.
 open-aea-ledger-ethereum = {version = "==2.2.9", optional = true}
 open-aea-ledger-ethereum-hwi = {version = "==2.2.9", optional = true}
+open-aea-ledger-solana = {version = "==2.2.9", optional = true}
```
DIvyaNautiyal07, line 48: **Nit (activated by this PR, pre-existing code)** — `packages/valory/contracts/squads_multisig/contract.py:31` still has a bare `from aea_ledger_solana import SolanaApi` at module scope. With the plugin now optional, a user who accidentally includes a Solana-dependent skill without `[solana]` gets a raw `ImportError: No module named 'aea_ledger_solana'` at load time with no hint. A wrapped import (`except ImportError as e: raise ImportError("Install open-autonomy[solana] to use squads_multisig") from e`) would turn a cryptic crash into an actionable message. Not strictly required for this PR, but this PR is the moment where it starts to bite.

pyproject.toml
```diff
@@ -57,7 +64,8 @@ docker = {version = "==7.1.0", optional = true}
 #     corresponding package YAML.
 # ---------------------------------------------------------------------------
 open-aea-ledger-cosmos = "==2.2.9"
-open-aea-ledger-solana = "==2.2.9"
+# `open-aea-ledger-solana` belongs here too, but is declared optional
+# above and exposed via the [solana] extra.
```
DIvyaNautiyal07: **Nit** — "belongs here too" reads as "should be moved back here." A future maintainer doing a cleanup pass could act on that reading and re-introduce the `websockets<12` cap. Suggest: `# open-aea-ledger-solana is kept optional above to avoid the websockets<12 cap; exposed via [solana].`

docs/upgrading.md
```diff
@@ -5,6 +5,24 @@ Below, we describe the additional manual steps required to upgrade between diffe
...
+
+```bash
+pip install open-autonomy[solana]
+```
+
+`[all]` does not imply `[solana]`, so combine them if you need both: `open-autonomy[all,solana]`. The same applies to `deployments/Dockerfiles/autonomy-user/requirements.txt`-style pins in downstream images.
+
+**If your service does not use Solana**, no action is required — and this release unblocks a dependency conflict you may have been hitting. The plugin pulled in `solana<0.34.0`, which caps `websockets<12.0`, so any downstream needing modern `websockets` (e.g. `google-genai>=1.0`, which requires `websockets>=13.0`) could not resolve against `open-autonomy` `0.21.20`–`0.21.26`. It failed with `ResolutionImpossible`, or on `pip>=26.2` the less obvious `resolution-too-deep`. If you pinned `open-autonomy<=0.21.19`, forced a `websockets` override, or vendored a constraints file to work around it, drop the workaround.
+
+This supersedes the "`open-aea-ledger-solana` re-enabled" note under `v0.21.19` to `v0.21.20` below: letting `open-autonomy[all]` pull the plugin in is no longer the recommended path.
```
DIvyaNautiyal07, line 24: **P1** — the "supersedes" note here is good, but the old 0.21.20 section (around line 189) still stands untouched with the affirmative advice *"you can let `open-autonomy[all]` pull it back in."* A reader landing directly on the 0.21.20 note (deep-link, or scrolling from an older upgrade) won't see this superseding note. Suggest adding an inline callout inside the 0.21.20 section itself, e.g. **"Superseded by v0.21.26 → v0.21.27 above — do not follow this step against ≥0.21.27."**

tests/test_base.py
```diff
@@ -62,7 +62,7 @@ def get_test_files(package_type: PackageType) -> List[Path]:
 
 def test_version() -> None:
     """Test the version."""
-    assert autonomy.__version__ == "0.21.26"
+    assert autonomy.__version__ == "0.21.27"
```
DIvyaNautiyal07, line 65: **Nit** — `tests/test_autonomy/test_packaging.py` already parses `pyproject.toml` structurally, so two extra assertions there would cheaply lock in this PR's invariant:

```python
assert "open-aea-ledger-solana" in extras["solana"]
assert "open-aea-ledger-solana" not in extras["all"]
```

Catches the regression where someone folds the plugin back into `[all]` (which would silently reimpose the `websockets<12` cap on every downstream test env).

DIvyaNautiyal07 (dismissed):

DIvyaNautiyal07 (approved):

Adamantios (approved):

PR #2543, opened 2026-08-27
fix(analyse): select the chained ABCI skill deterministically + release 0.21.28

OjusWiZard:
## Proposed changes

`autonomy analyse service` picked the chained ABCI skill non-deterministically.

`_get_chained_abci_skill` iterated `AgentConfig.skills` — a `Set[PublicId]` — and returned the **first** non-abstract skill declaring `abstract_round_abci` as a dependency. With more than one qualifying skill the winner was decided by set iteration order, so the command's answer changed with `PYTHONHASHSEED`.

On [valory-xyz/trader](https://github.com/valory-xyz/trader) both `trader_abci` and `funds_manager` qualify (each `is_abstract: false`, each declaring `abstract_round_abci`):

| `PYTHONHASHSEED` | selected skill |
| --- | --- |
| 0, 1 | `funds_manager` |
| 2, 3 | `trader_abci` |

That coin flip is why trader carries `analyse-service` commented out of its `common_checks` workflow with the note that it *"is checking funds manager skill which is not an ABCI skill"* — a description of one face of the coin rather than a stable property.

### The fix

Collect every qualifying candidate, then pick the one **nothing else in the agent declares as a dependency** — the root of the dependency DAG, which is what "chained app" means. `trader_abci` lists `funds_manager` among its `skills`, so it is the unique root. Where there is no single root, the command raises a named `ClickException` listing the candidates instead of silently picking one.

The code this replaces documented the assumption it was breaking:

> This statement makes an assumption skills other than the chained/main abci are defined as abstract

Three details are load-bearing, each covered by a test:

- **Edges come from every skill the agent overrides, not just the candidates.** Abstract skills are filtered out before candidacy, so reading edges off candidates alone misses composition through an abstract intermediate: for `A → B → C` with `B` abstract, the `B → C` edge is invisible, `A` and `C` both look like roots, and a chain with one obvious root is rejected as ambiguous.
- **Candidates are keyed on the versioned identifier**, so two versions of one skill stay two candidates and get reported rather than one silently overwriting the other.
- **The edge comparison is version-insensitive**, so a dependency pinned to a different version than the agent's skill list still registers as the same skill. A version-sensitive comparison would see no edge and abort as ambiguous.

The pattern in all three: a hole in the edge model turns into a spurious hard error on a valid config, which is a worse failure than the coin flip being fixed.

### Why not just `sorted()`

Sorting alone is deterministic but *wrong*: in trader it would consistently select `funds_manager`, which is worse than the coin flip because it looks stable. The regression test's fixtures are therefore named `aaa_dependency` / `zzz_chained`, so the dependency sorts before the root exactly as `funds_manager` does before `trader_abci`. Stripping the dependency logic fails the test with `assert 'aaa_dependency' == 'zzz_chained'`; **please keep that naming.**

### Note for reviewers: the single-candidate fast path is load-bearing

`if len(candidates) > 1:` guards the DAG logic, and the lone-candidate case returns directly. That looks redundant — a single candidate cannot depend on itself — but it is not, because the dependency set now includes edges from **non-candidate** skills. Collapsing the two paths would mean a single candidate that any abstract skill happens to declare resolves to zero roots and raises: a spurious hard failure on the most common configuration there is. Please don't merge the branches.

### Release

Also cuts **0.21.28**, bumping the version across all eight pin sites. The `open-aea-test-autonomy` pin in `transaction_settlement_abci/skill.yaml` cascades through `autonomy packages lock` into eight dependent package hashes and on into `docs/package_list.md`. `abstract_round_abci` does not depend on the bumped package, so `ABSTRACT_ROUND_ABCI_SKILL_WITH_HASH` in `autonomy/constants.py` stays in sync without edit.

The two `poetry.lock` version fields are updated in place rather than regenerated: `poetry lock --regenerate` rewrites ~2200 lines of unrelated dependency metadata *and still* records the stale plugin version, because Poetry reads installed dist-info for `develop` path dependencies. `poetry check --lock` passes.

## Fixes

Unblocks `analyse-service` in `trader` and `optimus` (separate downstream PRs). `optimus`'s `basius` service, previously blocked by this bug, now reports its real findings.

## Types of changes

One operator-visible behaviour change, documented in `docs/upgrading.md`: an agent with several non-abstract skills depending on `abstract_round_abci` and no dependency relating them now fails with a named error instead of returning an arbitrary skill. That is the intended trade, but it can surface as a new error on a service that previously "passed" by luck.

## Checklist

- [x] Lint and unit tests pass locally with my changes — with two pre-existing exceptions, see below
- [x] Public-facing documentation has been updated with the changes affected by this PR.

## Further comments

**Verification.** Deterministic across six `PYTHONHASHSEED` values against trader's real packages, and across all four services in `trader` and `optimus`. Six new tests; all five behaviours above are mutation-verified — each dies without at least one test. `mypy`, `pylint` (10.00/10), `flake8`, `black`, `isort`, `vulture`, `check-hash`, `check-packages`, `check-dependencies`, `check-doc-links-hashes`, `check-third-party-hashes`, `check-abci-docstrings`, `check-abciapp-specs`, `check-handlers`, `check-dialogues`, `liccheck`, `bandit` and `poetry check --lock` all pass. Full framework unit suite: 633 passed.

**Two pre-existing local failures, neither caused by this PR:**

- `tox -e darglint` exits 129 on six files in `autonomy/`, `plugins/` and `packages/valory/` that this PR does not touch. `darglint 1.8.1` breaks on Python 3.14, which is why `main_workflow.yml:109` pins that job to 3.13 with the comment *"use 3.14 after darglint2 issue #44 is fixed"*. The two changed files pass darglint cleanly.
- The docker-image tests (`test_build_image.py`, `test_images/`) fail locally without the images built; they are unrelated to this change.

**Unrelated loose end spotted while working, not addressed here:** `tox -e check-api-docs` regenerates an untracked `docs/api/data/`, which `f21680f99` ("skip data packages from generate api documentation script") deliberately deleted. The generator still emits it and the check does not catch it, so the fix looks incomplete.


tests/test_autonomy/test_cli/test_analyse/test_verify_service.py
```diff
@@ -924,3 +927,142 @@ def test_run(self, caplog: Any) -> None:
             "valory/termination_abci:any is not defined as a dependency" in caplog.text
         )
         assert "valory/slashing_abci:any is not defined as a dependency" in caplog.text
+
+
+class TestChainedAbciSkillSelection:
```
DIvyaNautiyal07, line 932: **P1 — coverage gap.** No test in `TestChainedAbciSkillSelection` would catch dropping the `_has_abstract_round_abci_skill_as_dependency` check at `autonomy/cli/helpers/analyse.py:361`.

Every fixture skill is built from `get_dummy_skill_config()` which already includes `valory/abstract_round_abci:0.1.0` in its `skills` list, so the ABCI-dep gate never filters anything in these tests. If a future refactor removed the check (letting non-ABCI skills into `candidates`), every test still passes silently.

Add a targeted test: an agent with a non-abstract skill whose `.skills` list does NOT contain `abstract_round_abci`, alongside a proper chained ABCI skill, and assert the ABCI skill is returned (not ambiguity, not the non-ABCI one).

autonomy/cli/helpers/analyse.py
```diff
@@ -317,9 +317,12 @@ def _get_chained_abci_skill(
     :param is_on_chain_check: A boolean flag to specify whether this is an on-chain
                             check or a local check
     :return: Skill configuration object if found
+    :raises ClickException: If the chained ABCI skill cannot be identified.
```
DIvyaNautiyal07: **P2 — inaccurate `:raises:` clause.** The docstring says the function raises `ClickException` when the chained ABCI skill "cannot be identified", but the function only raises on the ambiguous multi-candidate case (lines 378-385). Zero candidates still returns `None` (line 367); the caller at `check_service_readiness` converts that separately into its own error message.

A reader guarding solely on the exception would miss the None-return path.

```suggestion
    :raises ClickException: If multiple non-abstract skills declare
        ``abstract_round_abci`` as a dependency and none is the sole root of the
        others.
```

autonomy/cli/helpers/analyse.py
```diff
@@ -354,11 +361,32 @@ def _get_chained_abci_skill(
...
+        ]
+        if len(roots) != 1:
+            raise click.ClickException(
+                "Could not determine the chained ABCI skill; "
+                f"{len(candidates)} non-abstract skills depend on "
+                "`abstract_round_abci` and none is the sole root of the "
+                "others: "
+                + ", ".join(sorted(map(str, candidates)))
+                + "\nMark the non-chained ones as abstract, or declare the "
+                "dependency between them."
```
DIvyaNautiyal07, line 395: **P2 — error message conflates two structurally different cases.** `if len(roots) != 1:` catches both `len(roots) == 0` (every candidate is declared as a dep of another — a cycle, or all candidates pulled in by a non-candidate) AND `len(roots) > 1` (two or more roots — the actual ambiguous case).

The phrase "none is the sole root of the others" reads accurately for the zero-roots case but wrong for the two-roots case. The hint "Mark the non-chained ones as abstract, or declare the dependency between them" is the right fix for two-roots and misdirects operators when the real problem is a cycle.

Split the branches so each gives an actionable message. Rough shape:

```python
if not roots:
    raise click.ClickException(
        "Could not determine the chained ABCI skill; every non-abstract "
        "candidate is declared as a dependency by another skill in the agent "
        "(possible cycle): " + ", ".join(sorted(map(str, candidates)))
    )
raise click.ClickException(
    f"Could not determine the chained ABCI skill; {len(roots)} candidate roots "
    "with no dependency relating them: " + ", ".join(sorted(map(str, roots))) +
    "\nMark the non-chained ones as abstract, or declare the dependency between them."
)
```

autonomy/cli/helpers/analyse.py
```diff
@@ -354,11 +361,32 @@ def _get_chained_abci_skill(
...
-        # abci are defined as abstract
-        return skill_config
+        candidates[skill_id.without_hash()] = skill_config
+
+    if not candidates:
+        return None
+
+    if len(candidates) > 1:
+        # The chained app composes the others, so it is the one nothing
+        # else in the agent declares as a dependency.
```
DIvyaNautiyal07: **Nit — comment slightly understates the edge set.** `declared_dependencies` at lines 348-350 is populated from every overridden skill's `.skills` list — candidates plus non-candidates (including abstract intermediates). That broader scope is load-bearing for the abstract-intermediate composition case: `A → B → C` with `B` abstract resolves correctly only because `B`'s `B → C` edge is captured even though `B` isn't a candidate.

The current comment reads as "nothing else in the agent" but silently means "nothing else *among what we loaded*". Small reword makes the invariant explicit:

```suggestion
        # The chained app composes the others, so it is the one no other
        # skill in the agent (candidate or abstract intermediate) declares
        # as a dependency.
```

tests/test_autonomy/test_cli/test_analyse/test_verify_service.py
```diff
@@ -924,3 +927,142 @@ def test_run(self, caplog: Any) -> None:
...
+    """Selecting the chained ABCI skill when several candidates qualify."""
+
+    def _skill(self, name: str, extra_skills: List[str]) -> Dict[str, Any]:
+        config = get_dummy_skill_config()
+        config["name"] = name
+        config["skills"] += extra_skills
+        return config
+
+    def _run(
```
DIvyaNautiyal07, line 943: **Nit — no direct test for the `return None` branch of `_get_chained_abci_skill`.**

`test_abci_skill_not_found` (elsewhere in this file) exercises the CLI wrapper's "chained ABCI skill package not found" message via a different mechanism (empty `.skills` on the chained skill config), but the helper's own zero-candidate return branch (`if not candidates: return None` at analyse.py:367) has no direct assertion.

5-line test: agent with one skill that lacks `abstract_round_abci`, assert `_get_chained_abci_skill(...) is None`. Pins the None-return contract that the caller relies on.

tests/test_autonomy/test_cli/test_analyse/test_verify_service.py
OjusWiZard: Done in `ec2926ef2` — with one correction to the diagnosis worth recording.

Removing the `_has_abstract_round_abci_skill_as_dependency` check does not leave the suite green: it fails `TestVerifySkillConfig::test_abci_skill_not_found`, which reaches the gate through the CLI. I mutated it away to check before acting.

Your point holds where it counts though — all of `TestChainedAbciSkillSelection` passed with the gate gone, exactly because every fixture inherits `abstract_round_abci` from `get_dummy_skill_config()`. The class that owns this logic could not see its own precondition, and the only thing catching it did so incidentally.

Added `test_a_skill_without_the_abci_dependency_is_not_a_candidate`: `_skill()` takes an `abci` flag, and the helper skill is built without the dependency. It also pins the consequence you identified — the non-ABCI skill must not become a second candidate and turn a fine agent into an ambiguity error. Removing the gate now fails 2 tests in the class.

autonomy/cli/helpers/analyse.py
OjusWiZard: Fixed in `ec2926ef2`, wording adapted slightly from your suggestion:

```
:raises ClickException: If several non-abstract skills declare
    `abstract_round_abci` as a dependency and none is the sole root of
    the others. A total absence of candidates returns `None` instead.
```

I kept your first sentence and added the explicit `None` note, since the failure mode you describe is a reader who guards only on the exception — saying what the other path does seemed more use than leaving it implied.

autonomy/cli/helpers/analyse.py
OjusWiZard: Agreed and done in `ec2926ef2`. This was the sharpest of the five.

The two faults are genuinely different and the advice inverts between them: for several roots the fix is to relate the skills by a dependency, and for no roots a dependency relating them is precisely what already exists. Telling an operator in the second case to declare one sends them the wrong way.

Split as you suggested, with one change — each message now names only the skills relevant to its own fault. The no-root message lists the candidates (any of them could be the culprit); the several-roots message lists `roots` rather than all candidates, so the operator sees only the skills actually competing.

Also added `test_no_root_at_all_is_reported_separately` — an abstract skill declaring the chained app, which leaves no root. That branch had no coverage when I first wrote it, so the split would have been unverified. It matters for a second reason: it is the exact case that makes the single-candidate fast path load-bearing, since merging the paths would raise here for a lone candidate.

autonomy/cli/helpers/analyse.py
OjusWiZard: Fixed in `ec2926ef2`. You are right that "in the agent" papered over the invariant that actually does the work.

```
# The chained app composes the others, so it is the one no other
# loaded skill -- candidate or abstract intermediate -- declares as a
# dependency.
```

Worth noting the wider scope is not incidental: reading edges from the candidates alone was a real bug caught in self-review. Abstract skills are filtered before candidacy, so for `A -> B -> C` with `B` abstract the `B -> C` edge vanished, `A` and `C` both looked like roots, and a chain with one obvious root was rejected as ambiguous. `test_an_abstract_intermediate_still_yields_one_root` covers it.

tests/test_autonomy/test_cli/test_analyse/test_verify_service.py
OjusWiZard: Done in `ec2926ef2` — `test_no_qualifying_skill_returns_none`, asserting `_get_chained_abci_skill(...) is None` for an agent whose only skill lacks `abstract_round_abci`.

Worth pinning for the reason the docstring comment above surfaced: the function signals "none found" and "ambiguous" through two different channels, a `None` return and an exception. The `None` side had no direct assertion, so nothing stopped a later change from raising there instead and breaking the caller's branch at `check_service_readiness`.

DIvyaNautiyal07 (approved):
All five previous review findings addressed cleanly in `ec2926ef`. Bonus mutation test pins the split-branches invariant. LGTM.

Adamantios (approved):

PR #2544, opened 2026-09-11
feat(chain): add Robinhood Chain support + release 0.21.29

OjusWiZard:
## Proposed changes

Adds **Robinhood Chain** (chain ID `4663`, an Arbitrum Orbit L2) as a supported chain. Olas registries are already deployed there.

- `ChainType.ROBINHOOD`, which also generates `--use-robinhood` on `mint`, `service`, `fetch`, `analyse` and `deploy`.
- A `robinhood` entry in `CHAIN_PROFILES`, `CHAIN_NAME_TO_CHAIN_ID` and `CHAIN_ID_TO_DEFAULT_PUBLIC_RPC` (`https://rpc.mainnet.chain.robinhood.com`, overridable with `ROBINHOOD_CHAIN_RPC`).
- `4663` in `SERVICE_MANAGER_TOKEN_COMPATIBLE_CHAINS`, in both `autonomy/chain/constants.py` and the `service_manager` contract package. Without it the existing live drift test fails, because upstream already lists Robinhood. This re-locks the `service_manager` hash; `abstract_round_abci` is unaffected.

### Where the addresses come from

- **Registry contracts:** the `robinhood` entry of autonolas-registries [`docs/configuration.json`](https://github.com/valory-xyz/autonolas-registries/blob/main/docs/configuration.json) on `main`, mapped to profile keys the same way Celo's are (`GnosisSafeMultisig` → `gnosis_safe_proxy_factory`, `IdentityRegistryBridgerProxy` → `erc8004_identity_registry_bridger`).
- **`sign_message_lib` and `erc8004_identity_registry`:** not in that file. They reuse the addresses every other chain uses and were checked only for deployed code on chain 4663; the bytecode was not compared with other chains.
- **Checked on-chain:**
  - the RPC returns chain ID `0x1237`
  - `ServiceRegistry.manager()` returns the proxy `0x63e66d7a…`, which answers `serviceRegistryTokenUtility()` and `identityRegistryBridger()` with the profile's addresses
  - blocks carry 32 bytes of `extraData`, so Robinhood stays out of `POA_CHAINS` (as `arbitrum_one` does); it is also correctly outside `L1_CHAINS`

### Docs

- `--use-robinhood` added to the five command reference pages.
- A Robinhood section appended **by hand** to `on_chain_addresses.md`. Regenerating the file would also rewrite every other chain's Service Manager row, and the generator emits the implementation address rather than the proxy that `manager()` returns.
- `generate_contract_list.py` had been failing with `KeyError: 'arbitrumOne'` since upstream renamed the chain to `arbitrum`. The key is fixed and a Robinhood Blockscout URL added.

### Tests

- **`test_addresses_match` is re-pinned** from registries tag `v1.3.0`, which predates Robinhood (as does every tag up to `v1.3.3`), to tag [`v1.3.4`](https://github.com/valory-xyz/autonolas-registries/releases/tag/v1.3.4). It now checks Robinhood with no skip. The newer file renamed `arbitrumOne` to `arbitrum` and added contracts that our profiles key differently or don't track, so the test maps names explicitly (`UPSTREAM_CHAIN_NAMES`, `UPSTREAM_TO_PROFILE_KEY`, `UNTRACKED_CONTRACTS`). Polygon stays skipped under its existing RPC TODO.
- **New `test_evm_chain_resolves_without_env`** runs for every EVM `ChainType` with its RPC env var unset and checks that the chain ID round-trips, a default RPC resolves and a `service_registry` address exists. It replaces a one-line `ROBINHOOD.id == 4663` assertion that couldn't catch anything. Deleting Robinhood's RPC entry makes it fail for Robinhood only.

### Constants the re-pin corrected

The `v1.3.0` pin had been hiding drift in existing chain profiles:

- **Celo `staking_verifier`** was stale: `0xc40C79…` is now `0x4aba1Cf7…`, which is what Celo's `StakingFactory.verifier()` returns on-chain.
- **`complementary_service_metadata`** is added for Arbitrum One, Optimism, Celo and Mode, from upstream.
- **Arbitrum One and Base `erc8004_identity_registry_bridger`** are checksummed; the addresses are unchanged.

### Release

Also cuts **0.21.29**, bumping the same eight pin sites as 0.21.28. The `open-aea-test-autonomy` pin cascades into the same eight dependent package hashes and `docs/package_list.md`. `ABSTRACT_ROUND_ABCI_SKILL_WITH_HASH` is unchanged. The two `poetry.lock` version fields are edited in place; `poetry check --lock` passes.

## Fixes

Part of OPE-1925 (bringing Robinhood Chain to Pearl).

## Types of changes

## Checklist

## Further comments

**Local verification.** Passing: black, isort, flake8, mypy, pylint, bandit, gitleaks, `check-copyright`, `check-hash`, `check-packages`, `check-api-docs`, `check-dependencies`, `poetry check --lock`, and the chain tests (`test_chain/test_config.py`, `test_utils.py`, `test_addresses.py`, `test_click_utils.py`, the `service_manager` contract tests, `test_version`). The lint/unit box is unchecked because:

- the full unit suite was not run locally
- whole-repo `darglint` crashes inside the tool (exit 129); it is clean on every changed file
- `safety` fails because tomte still calls the deprecated `safety check` command

**Follow-ups, not in this PR:**

- **`generate_contract_list.py`** maps `ServiceManager` to the implementation rather than `ServiceManagerProxy`, and regenerating `on_chain_addresses.md` would refresh stale Service Manager rows for the other chains.
- **`--use-mode`** is missing from the command reference docs (pre-existing).

The `HISTORY.md` entry for 0.21.29 still needs this PR's number.


Adamantios (approved):

DIvyaNautiyal07 (approved):

DIvyaNautiyal07 (commented):
This is an approve-adjacent review — no blocking issues. Verified the addresses match upstream, all round-2 fixes cleared, and the Robinhood-staking "shift" concern I hunted turned out to be coincidence (fetched `autonolas-registries` `v1.3.4` `docs/configuration.json` and the four staking / bridger addresses match verbatim). Findings below are all non-blocking; two P2s and two Nits.

---

**P2 — `SERVICE_MANAGER_TOKEN_COMPATIBLE_CHAINS` is duplicated with no drift-catching test**

Lives in both `autonomy/chain/constants.py:325-342` and `packages/valory/contracts/service_manager/contract.py:45`. This PR correctly updates both copies, but the PR description itself notes the dual listing is required for the live drift test to pass — which means only the contract-package copy is CI-checked against upstream. A future chain addition on one side and not the other will pass CI silently.

*Fix:* single source of truth (import one from the other), or add a small unit test asserting `set(<constants copy>) == set(<contract-package copy>)`.

---

**P2 — `test_addresses_match`'s `UNTRACKED_CONTRACTS` skips the non-proxy `service_manager` and three others**

`tests/test_autonomy/test_chain/test_addresses.py` excludes `service_manager`, `service_manager_token`, `identity_registry_bridger`, and `hash_checkpoint` from the upstream comparison. `service_manager_proxy` is checked via a live on-chain call that self-skips on RPC failure, but the non-proxy `service_manager` address and `service_manager_token` never get diffed against `CHAIN_PROFILES` at all. If upstream ever changes either address, this test will pass while the local profile silently points at stale contracts — exactly the class of drift the Celo `staking_verifier` staleness this PR fixes belongs to.

*Fix:* narrow `UNTRACKED_CONTRACTS` to only entries that genuinely have no equivalent in `CHAIN_PROFILES` and add `UPSTREAM_TO_PROFILE_KEY` entries for `service_manager` and `service_manager_token`; or leave an inline comment on each `UNTRACKED_CONTRACTS` entry explaining why the local profile intentionally doesn't track it.

---

**Nit — `generate_contract_list.py` still uses non-defensive `BLOCKSCAN_URLS[chain_name]`**

`plugins/aea-helpers/aea_helpers/generate_contract_list.py:77` is a direct dict index. This PR fixes the previous `KeyError: 'arbitrumOne'` by renaming the key and adding a `robinhood` entry, but two lines below, the sibling `CONTRACT_TO_SLUG` lookup already uses the defensive `if name not in CONTRACT_TO_SLUG: continue` pattern. The `arbitrumOne` → `arbitrum` rename in `autonolas-registries` broke this script silently for months precisely because there was no defensive fallback here — the next upstream rename will do it again.

*Fix:* mirror the `CONTRACT_TO_SLUG` pattern — `if chain_name not in BLOCKSCAN_URLS: log.warning("no blockscan URL for %s", chain_name); continue`.

---

**Nit — annotation gaps around the two dual-list additions and the mapping tables**

Three small doc/comment gaps clustered around the changes: (a) every existing entry in `SERVICE_MANAGER_TOKEN_COMPATIBLE_CHAINS` has a trailing `# chain_name` comment (`# polygon`, `# mode`, etc.); the new `4663,` line in both `constants.py:335` and `contract.py:45` has none. (b) Neither copy carries a comment noting the two lists must stay in sync — the drift risk in the first P2 above becomes invisible to a future reader. (c) The mapping tables in `tests/test_autonomy/test_chain/test_addresses.py` (`UPSTREAM_CHAIN_NAMES`, `UPSTREAM_TO_PROFILE_KEY`, `UNTRACKED_CONTRACTS`) have zero inline comments; non-obvious cases (why `identity_registry_bridger` is untracked while `identity_registry_bridger_proxy` maps to `erc8004_identity_registry_bridger`; why `arbitrum_one` maps to upstream `"arbitrum"`) go unexplained here — that context lives only in HISTORY.md.

*Fix:* add `4663,  # robinhood` in both files with a `# NOTE: keep in sync with the copy in <other file>` above; add short inline comments to the three test-file mapping tables explaining the upstream-vs-local naming choices.

---

**Cleared during verification**

- Robinhood profile addresses (`staking_token`, `staking_verifier`, `staking_factory`, `erc8004_identity_registry_bridger`, all others) match upstream `autonolas-registries` tag `v1.3.4` exactly — verified by fetching `docs/configuration.json` at that tag.
- The apparent one-row shift between Mode's and Robinhood's staking addresses is coincidence, not a mislabelling. Upstream confirms Robinhood's `staking_token` really is `0x87c511c8…` (which happens to equal Mode's `staking_verifier`), and so on.
- All Robinhood addresses pass EIP-55 checksum validation.
- `--use-robinhood` is auto-generated via `autonomy/cli/utils/click_utils.py:106`'s `f"--use-{chain_name}"` — adding the enum variant is enough. Docs across the 5 command reference pages are consistent.
- No hardcoded `ChainType.X` special-cases anywhere outside tests; `ChainType.ROBINHOOD` threading is fully generic through `CHAIN_PROFILES` / `CHAIN_NAME_TO_CHAIN_ID`.
- `POA_CHAINS` correctly excludes Robinhood, matching Arbitrum One.
- All 8 release-0.21.29 pin sites bumped consistently; `service_manager/contract.yaml` fingerprint and `packages/packages.json` re-locked correctly; dependent skill hashes rippled through the `open-aea-test-autonomy` bump.
- `test_evm_chain_resolves_without_env` iterates the live `ChainType` enum (minus `CUSTOM` and `SOLANA`), so a future EVM chain added without an RPC entry would fail the test.
- `test_addresses_match` is version-locked to `v1.3.4` (tag ref, not branch).

